PR #1, opened 2019-09-27
Playback volumes

juimonen:


ucm/sof-skl_hda_card/HiFi.conf
```diff
@@ -118,6 +122,7 @@ SectionDevice."HDMI1" {
 
 	Value {
 		PlaybackPCM "hw:0,3"
+		PlaybackVolume "PGA7.0 7 Master"
```
plbossart: Do we have a good reason to enable volume control for HDMI?
It can be done in PulseAudio or in the receiver.

ucm/sof-skl_hda_card/HiFi.conf
```diff
@@ -97,6 +100,7 @@ SectionDevice."Dmic" {
 
 	Value {
 		CapturePCM "hw:0,6"
+		CaptureVolume "PGA10.0 10 Master"
```
plbossart: That's super fragile. Any change to topology would result in this being broken.
We should really have an alias in the topology so that there is no dependency on PGA indices.
Also for DMIC the notion of hardware volume is a bit misleading. The real control with the 20dB boost is already part of the IIR filter added by @singalsuo

ucm/sof-skl_hda_card/HiFi.conf
juimonen: fine with me, can/will remove it.

ucm/sof-skl_hda_card/HiFi.conf
juimonen: Agree, the pga namings should be clear, I'll investigate this. So in this context with "hw volume" I mean something behind alsa, so not in user space (weather it is really how or not). So I need some volume control (and switch) for dmic, so that acpi led quirks can work. Also I think @singalsuo has (or has plans to) implemented over 0db gains to volume. Also this volume control seems to work so you can add least attenuate the dmic level and it seems to react nicely to UI changes.

juimonen: @plbossart can you elaborate little bit, so is it ok if I just rename/hardcode all dmic related pga/switches in our topology (like "dmic volume" / "dmic switch")? Or do you want some rename mechanism to sof driver (generic hda has some rename code). Both are fragile and dependent on topology names also, perhaps not so fragile as having the indices...

plbossart: I don't even know how we end-up with a name like PGA 10.0 10...
This numbering is fine for internal indexing but it's meaningless for users.
I don't understand M4/topology enough to recommend either a rename or the addition of a 'friendly-name'.

ucm/sof-skl_hda_card/HiFi.conf
juimonen: removed for now.

ucm/sof-skl_hda_card/HiFi.conf
juimonen: changed now to "dmic0" (and possibly "dmic1") which are now in upstream sof fw topology

juimonen:
@perexg added PlaybackSwitches although not used yet, but should not hurt either.

perexg:
Merged in c8808f6d60b5f38d1400bf3bd4cdc059fc080bb7 . Thanks.

perexg:
Note: The syntax (value) was wrong. The amixer style is supported only. Fixed in 244a73a5619308c739bedbc82a4da690a0dbb5d7 .

PR #2, opened 2019-09-27
sof-skl_hda: remove volume setting in enable/disable sequences

juimonen:
If the volume setting is defined in enable sequence you will
always get that exact volume when switching outputs in some
user space software like Pulseaudio. Pulseaudio will itself
keep track of the volume previously set.


juimonen:
This is little bit more invasive, I'm assuming that pulseaudio uses hardware volume in which case these removed settings will conflict when changing output.

PR #3, opened 2019-10-22
sof-skl_hda: add device definition to speaker disable sequence

juimonen:
For example Pulseaudio is not able to run the disable sequence
if the device is not defined. So add the definition.


perexg:
Merged in 8797ee0bd9210f91f2335b74840e49dd3b52b80f . Thanks.

PR #4, opened 2019-10-23
[RFC] sof-skl_hda: change to use common HDMI codec driver

kv2019i:
Here's a companion patch to the hda-hdmi kernel patchset on alsa-devel:
https://mailman.alsa-project.org/pipermail/alsa-devel/2019-October/157201.html

Modify UCM sequences to use user-space interface of the common
HDMI codec driver. Use of hdac-hdmi driver is no longer supported
by this UCM file.

To use this UCM file, kernel must be compiled with
SND_SOC_SOF_HDA_COMMON_HDMI_CODEC

### DIFF
--- a/ucm/sof-skl_hda_card/HiFi.conf
+++ b/ucm/sof-skl_hda_card/HiFi.conf
@@ -112,20 +112,18 @@ SectionDevice."HDMI1" {
 
 	EnableSequence [
 		cdev "hw:sofsklhdacard"
-		cset "name='hif5-0 Jack Switch' on"
-		cset "name='Pin5-Port0 Mux' 1"
+		cset "name='IEC958 Playback Switch' on"
 	]
 
 	DisableSequence [
 		cdev "hw:sofsklhdacard"
-		cset "name='Pin5-Port0 Mux' 0"
-		cset "name='hif5-0 Jack Switch' off"
+		cset "name='IEC958 Playback Switch' off"
 	]
 
 	Value {
-		PlaybackPCM "hw:0,3"
+		PlaybackPCM "hw:sofsklhdacard,3"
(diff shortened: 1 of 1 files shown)


ucm/sof-skl_hda_card/HiFi.conf
```diff
@@ -103,20 +103,17 @@ SectionDevice."HDMI1" {
...
 		cdev "hw:sofsklhdacard"
-		cset "name='hif5-0 Jack Switch' on"
-		cset "name='Pin5-Port0 Mux' 1"
+		cset "name='IEC958 Playback Switch' on"
 	]
 
 	DisableSequence [
 		cdev "hw:sofsklhdacard"
-		cset "name='Pin5-Port0 Mux' 0"
-		cset "name='hif5-0 Jack Switch' off"
```
perexg: What's the reason to remove the enable counterpart?

kv2019i: @perexg It seems the IEC958 mute usage is not very systematic and there seems to be no harm in leaving it enabled. It does have to be unmuted as e.g. some systems mute it in their default alsa-ctl init state. It took quite a bit of time e.g. to dig this up in Ubuntu (added in 2005):
“On-board SIS sound card needs muted "IEC958"”
https://bugs.launchpad.net/ubuntu/+source/alsa-utils/+bug/19648

With the older hdac-hdmi driver, the disable sequence was mandatory to avoid problems in routing.

perexg:
Merged to 32736b891913bf610bdacb259ff17607a4caf8ec . Tested with the latest SOF driver. I pushed many other SOF related cleanups, if you like to check.

PR #5, opened 2019-10-25
sof-skl_hda: fix headphone jack name

juimonen:
Change "Headphone Jack" to "Headphone Mic Jack".

### DIFF
--- a/ucm/sof-skl_hda_card/HiFi.conf
+++ b/ucm/sof-skl_hda_card/HiFi.conf
@@ -34,7 +34,7 @@ SectionDevice."Headphone" {
 		JackName "sof-skl_hda_card Headphone"
 		JackType "gpio"
 		JackSwitch "12"
-		JackControl "Headphone Jack"
+		JackControl "Headphone Mic Jack"
 	}
 }
 


ucm/sof-skl_hda_card/HiFi.conf
```diff
@@ -34,7 +34,7 @@ SectionDevice."Headphone" {
 		JackName "sof-skl_hda_card Headphone"
 		JackType "gpio"
 		JackSwitch "12"
-		JackControl "Headphone Jack"
+		JackControl "Headphone Mic Jack"
```
lgirdwood: Is this fixing a bug ? If so can you describe in the commit message.

perexg:
There is no "Headphone Mic Jack" in the current SOF driver:

<pre>
# amixer -c 0 controls | grep Jack
numid=17,iface=CARD,name='HDMI/DP,pcm=3 Jack'
numid=23,iface=CARD,name='HDMI/DP,pcm=4 Jack'
numid=29,iface=CARD,name='HDMI/DP,pcm=5 Jack'
numid=16,iface=CARD,name='Headphone Jack'
numid=13,iface=CARD,name='Mic Jack'
numid=14,iface=CARD,name='Speaker Front Phantom Jack'
numid=15,iface=CARD,name='Speaker Surround Phantom Jack'
</pre>

juimonen:
@perexg for me there is... :) Is this coming from somehow from the HDA side from codec quirks etc?... I see some code adding words like "Mic" and "Phantom" in to the Jack element.

juimonen:
from my dell device with sof:
amixer controls | grep -i Jack
numid=19,iface=CARD,name='HDMI/DP,pcm=3 Jack'
numid=25,iface=CARD,name='HDMI/DP,pcm=4 Jack'
numid=31,iface=CARD,name='HDMI/DP,pcm=5 Jack'
numid=16,iface=CARD,name='Headphone Mic Jack'
numid=17,iface=CARD,name='Headset Mic Phantom Jack'
numid=18,iface=CARD,name='Speaker Phantom Jack'

perexg:
Could you give me a link to the alsa-info.sh output on your machine? You need to define the DMI based configuration in this case (see the LENOVO-20QE000VMC-ThinkPadX1Carbon7th-20QE000VMC/ directory for an example).

juimonen:
Hi, ok so the "mic" thing is platform specific? Do you know the heuristics why it is parsed to the element name? Anyway I can do the DMI based configuration. Attached the alsainfo but not sure if the name DMI name is the final for this device, might not be.

So for example canonical is using 5.0 based kernel and pulseaudio 11.1. Only way I could actually get most use cases working was to separate all outputs as separate use cases (will become separate pulseaudio profiles) see: https://github.com/juimonen/alsa-ucm-conf-1/tree/dell_test 

I was testing this all week, sorry for not notifying you before, I though the PR's we're waiting for my input, actually I was suggesting @kv2019i that we close the existing PR's and I submit new.

[alsainfo.txt](https://github.com/alsa-project/alsa-ucm-conf/files/3795313/alsainfo.txt)

perexg:
It might be that your problems exist because the ConflictingDevice array is using wrong strings (comments instead the real device names).

The 'Headset Mic Jack' and 'Headset Jack' kcontrol name looks like a bug in the HDA driver. @tiwai ?

perexg:
https://mailman.alsa-project.org/pipermail/alsa-devel/2019-November/157794.html

perexg:
@juimonen : could you test the above kernel patch *without* UCM ?

perexg:
Sorry, forget my last two comments, I substituted Headphone with Headset somehow.

perexg:
This issue is addressed with my alsa-lib updates:

https://github.com/alsa-project/alsa-lib/commits/ucm2

The SOF branch for the new alsa-lib UCM code:

https://github.com/alsa-project/alsa-ucm-conf/commits/ucm2

lgirdwood:
@juimonen did this all test fine. I'm assuming so since it'as all applied.

juimonen:
@lgirdwood little bit unsure still... so I guess @perexg idea was to use conditional for this in ucm2? I still don't totally understand why the names are different? And was the idea to fix this in HDA or use conditional in ucm2?

perexg:
My vague understanding is that 'Headphone Jack' is that something is connected to the Headphone jack, but 'Headphone Mic Jack' check if the specifically microphone is in the jack. It depends on the HDA codec. I might be wrong. @tiwai , could you explain this?

I used conditions in ucm2 (merged to master right now because I'd like to release 1.2.10 packages) to select the right jack control name: https://github.com/alsa-project/alsa-ucm-conf/blob/master/ucm2/sof-hda-dsp/HiFi.conf#L41-L52

tiwai:
Yes, it's a shared functionality on the single headphone pin by jack-retasking.  
I believe PulseAudio deals with this, so the jack control name should remain as is.

perexg:
Resolved in 09bc9c70690b4bbce77f773b9f051b069f4f2fd9 .

juimonen:
@perexg sorry asking you this thing in closed PR... things we're moving pretty fast for ucm2 and I was trying to get legacy ucm work with older sw: So if I want to upstream better working legacy ucm for sof-skl_hda_card, (which is already renamed and in ucm2 dir), what's the place to "re-add" it? alsa-ucm-conf/ucm ?

perexg:
Could you describe your problem with ucm2?

perexg:
sof-skl_hda_card is the old name (before 5.5 kernel); we should not support it..

juimonen:
I'm trying to do this for ubuntu LTS 18 OEM release, which uses 5.0 based kernel, sof 1.3 and pulseaudio v11.1, which are somewhat patched... believe me I really would not like to use this old stuff, but I haven't got so much choice.

perexg:
You can create the SOF files out-of-upstream, of course (maybe a symlink from old sof-skl_hda_card to new name with few changes in HiFi.conf may be sufficient, or patch the kernel). I am trying to push the things to make upstream stable for SOF as I can.

juimonen:
@perexg I checked the master of alsa-ucm-conf, so we don't have backwards compatibility from current? I mean all the names have been changed, there's like nothing in "old" ucm folder. I thought the idea was to leave the old ones there and little by little change to ucm2? Or is there some other mechanism I just didn't get?

perexg:
But there was no upstream for sof-skl_hda_card until I started to dig to it. So this particular case does not fall to "we broke something in the upstream". Simply, Ubuntu did not send this config to the upstream. They cooked something outside.

juimonen:
Well it was me doing it and supposedly sending it when I have it working. There we're quite a lot of use cases I needed to check and make work. So we had this sof-skl_hda_card before renaming it, where was that coming? I mean I'm fine using whatever solution can work with this... I can tell them to take the rename patch to kernel, but then they probably need to use alsa-lib/utils with syntax2 and so on...? I just have to figure out best sw/patch combination for this to work.

perexg:
> Well it was me doing it and supposedly sending it when I have it working. There we're quite a lot of use cases I needed to check and make work. So we had this sof-skl_hda_card before renaming it, where was that coming?

I did this rename.

> I mean I'm fine using whatever solution can work with this... I can tell them to take the rename patch to kernel, but then they probably need to use alsa-lib/utils with syntax2 and so on...?

No, they can just place the old configs to ucm/ tree. They should work like before.

juimonen:
@perexg so can I make PR to add legacy sof-skl_hda_card to this repos "ucm" directory (not ucm2)? Sorry to bother you with this... They would like to see it "upstream" somewhere, I just don't know the correct place...

perexg:
@juimonen I don't see any benefit to add something to upstream which is old and not broadly used. From my view, the upstream code which will support SOF in an universal way is:

1) 5.5 SOF driver (with the updated name) and the legacy/SOF/SST driver auto-detection
2) 1.4 firmware (if Intel makes it work)
3) ucm2

I think that Ubuntu can use those two ways now:

1) maintain own UCM configs out-of-upstream for the older driver / firmware
2) track upstream changes for all components

PR #6, opened 2019-10-25
sof-skl_hda: move hdmi as separate use case

juimonen:
Move hdmi definitions as separate use case and conf file. This will
enable Pulseaudio to create separate profile for hdmi and make routing
much more fluent than with single profile.


juimonen:
#4 needs to be rebased if this goes in

ucm/sof-skl_hda_card/Hdmi.conf
```diff
@@ -0,0 +1,80 @@
+# Use case Configuration for skl-hda-card
+
+SectionVerb {
+
+	EnableSequence [
+		cdev "hw:sofsklhdacard"
+		cset "name='Master Playback Switch' on"
+		cset "name='Capture Switch' on"
```
plbossart: I don't fully get what splitting the config into two verbs will bring?

And specifically on this I am not sure why the capture switch should be on given that this verb is playback only?

lgirdwood: @juimonen verbs are the high level use case, e.g. HiFi as compared to voice call. I guess PA is mapping the verbs as a device class ?

perexg:
Merged to 474f7fd4c18a60ba4c6f7a22005409e8b5b4429c . Tested with the latest SOF driver.

ucm/sof-skl_hda_card/Hdmi.conf
plbossart: @juimonen can you follow-up on the comment? I wonder if we need to do this split for all other UCM files? It's really not clear to me what this change brings, we've always used a single use case really.

juimonen: @plbossart yeah sorry I was trying the whole week to get this working somehow with pulsaudio.
see: https://github.com/juimonen/alsa-ucm-conf-1/tree/dell_test

So exactly as you said, I was only able to get most cases working by separating all outputs to different use cases (will become separate profiles in pulseaudio). With all outputs in single use case pulsaudio just can't change the sinks with jack events. If switching between ports in single hw device (like headphones and speakers usually are) it works, but switching between ports in different hw device, like hdmi doesn't work.

plbossart: humm... what about the inputs? f we want to switch between headset capture and dmic, does this mean dmics need to be in a separate use case?

juimonen: ok, I was little bit fuzzy with my explanation, so the above sink moving doesn't work with jack events. It works when forcing from UI. Now in the dell device I'm using we are not getting (and will not get) mic jack event from headset, so it is not automatically moved, you have to always explicitly move it from UI.

Now this not reacting to jack events is not possibly problematic when connecting devices. User always can open the ui and force it, you will anyway hear audio on current output. Problem is more when removing device, because the sound stops totally.

So this all is kind of also related to the user experience or audio policy and i'm pretty sure pulseaudio doesn't handle all this correctly even with legacy (without UCM)

juimonen: and mic jack event means that we get explicit jack event that there's microphone in the headphone/set. The normal headphone jack event is coming.

juimonen: and now when thinking of this, they probably should be separated anyway :)

perexg: BTW: I did many cleanups in this repo for SOF (including the wrong ConflictingDevice names, removal of this analog part initialization in Hdmi.conf etc.). It would be good, if you can retest the _current_ configs. Thanks.

plbossart: sorry @juimonen I don't get your points.  'not reacting to jack events is not possibly problematic' ?

juimonen: @plbossart so if the outputs are in the same use case, pulseaudio reaction to jack events will change the output (for example switch to headphones from speakers if they are in same use case). if also hdmi is in same use case this will not happen, because it is a different sink in pulseaudio (different alsa device hw0,3 etc instead of hw0,0). But if the sinks are in different profiles (and use cases in ucm) this move will happen.

As said this is not so annoying when you connect, but when you disconnect.

Now, we can leave this to be handled by the UI settings app, maybe some policy module in pulseaudio or similar. With current existing stuff in user space it just doesn't work very smoothly. Is there some conceptual problem of separating the outputs to different use cases?

plbossart: fast forward to SoundWire; we will have different hw devices for headset/headphone (rt711), speakers (rt1308x2), dmics (rt715).
Your solution of splitting the hardware devices by use cases will not scale - or it's becoming difficult to explain how a use case is related to a hardware device.

PR #7, opened 2019-11-26
Sof upstream

juimonen:


juimonen:
@perexg this is something that works pretty well in my testing...
there are couple of issue however:
1) in"sof-hda-dsp: move sound card setup to top level file"
I needed to use cdev, and didn't know how to do that... so the old one is there, I tried different combinations and pulseaudio was just barffing about missing cdev.
2) in "sof-hda-dsp: add headphone microphone"
I wanted to use conditional in enable sequence (you can see it in HiFi.conf) but it just didn't work for me, so in Mics.conf I have just the plain setting

perexg:
>        I needed to use cdev, and didn't know how to do that... so the old one is there, I tried different combinations and pulseaudio was just barffing about missing cdev.

Fixed in https://github.com/alsa-project/alsa-lib/commit/4c89f4eb905a075971537a11798dea37cf723bdf . Thanks for this notice.

>     2. in "sof-hda-dsp: add headphone microphone"
>        I wanted to use conditional in enable sequence (you can see it in HiFi.conf) but it just didn't work for me, so in Mics.conf I have just the plain setting

Conditional can be used only in compound {} blocks not arrays. You need to put If for the whole array definition.

ucm2/sof-hda-dsp/Mics.conf
```diff
@@ -11,11 +49,12 @@ SectionDevice."HeadsetMic" {
 	}
 }
 
-SectionDevice."DigitalMic" {
```
perexg: I'd like to keep DigitalMic here.

ucm2/sof-hda-dsp/Mics.conf
```diff
@@ -11,11 +49,12 @@ SectionDevice."HeadsetMic" {
...
+SectionDevice."Dmic" {
 	Comment "Digital Microphone"
 
 	ConflictingDevice [
-		"HeadsetMic"
+		"Headset Microphone"
```
perexg: Wrong and wrong. It must be UCM device name not comment.

ucm2/sof-hda-dsp/sof-hda-dsp.conf
```diff
@@ -5,7 +5,27 @@ SectionUseCase."HiFi" {
...
+}
+
+SectionUseCase."Hdmi3" {
+	File "Hdmi3.conf"
+	Comment "Play to HDMI3/DP3"
+}
+
+SectionDefaults [
+	cdev "hw:sofsklhdacard"
+	cset "name='Master Playback Switch' on"
```
perexg: The Master volume is just a soft-volume attenuator in this case. It should be set to 100%. Decrease the analog volume control in hw (Speaker Playback Volume, Headphone Playback Volume).

juimonen:
@perexg this is mostly trying to make things work in Pulseaudio as before without ucm. As I see it, if ports are in same device (like speaker and headset hw:0,0 usually) things work pretty well, because pulseaudio is not moving the streams, just setting the port (so essentially setting alsa switches). If ports are in different devices like HDMI, streams are not necessarily moved to new sink if port becomes unavailable. However if the device is in different _profile_ and that profile disappears, streams are moved. So this is a use case mainly when you for example rip off hdmi cable and audio should move to existing output, not play to imaginary hdmi which is not connected.  So me now dividing all HDMI's separately helps in case where you have 2 HDMI connected (in my test devices there's separate HDMI and DP, that you can both connect).

Now, I'm not insisting to divide things into separate use cases / profiles, but I'm not sure is there any harm, so is there some conceptional or real issue existing, that you should not do this.

There's  https://gitlab.freedesktop.org/pulseaudio/pulseaudio/merge_requests/209 to address this but not yet merged.

ucm2/sof-hda-dsp/sof-hda-dsp.conf
juimonen: ok fine by me, I've just taken previous "safe" values found, let me check this doesn't distort the audio or anything.

juimonen:
@perexg reporting this here, if it is bogus, just ignore:

In ubuntu, I was updating to new alsa-lib and alsa-utils to try the new ucm2:

git clone http://git.alsa-project.org/http/alsa-lib.git
cd alsa-lib
./gitcompile
sudo make install

git clone http://git.alsa-project.org/http/alsa-utils.git
cd alsa-utils
./gitcompile
sudo make install

In Ubuntu the alsa-utils compilation started to barf about missing libatopology, even though the alsa-lib was successfully compiled and installed. In Fedora this was working just fine. So somehow the linker was finding old alsa-lib version and not the new one. I got rid of this by violently removing libasound2 packaged files. Also if you start remove libasound2 package with dpkg, it will probably remove half of your user space. I understand using "sudo make install" is messing with the packaged files, but just to let you know and if there's some BKM's for ubuntu compilation (?).

lgirdwood:
@juimonen Ubuntu installs ALSA libraries in "/usr/lib/x86_64-linux-gnu/" so make sure you delete any other ALSA libraries installed in other locations.
You can also do an apt-get reinstall for alsa too if needed to recover your PC.

ucm2/sof-hda-dsp/Mics.conf
juimonen: the comment and name we're the same here (as I renamed headset microphone) device , but I will make device names and comments differ, so it is clear

juimonen: ok

ucm2/sof-hda-dsp/Mics.conf
juimonen: ok

juimonen:
@perexg updated. would be nice if you could try out in X1 to test if the conditionals actually work. They now seem to work for me in dell device.

ucm2/sof-hda-dsp/HiFi.conf
```diff
@@ -1,22 +1,33 @@
...
+			EnableSequence [
+				cset "name='Headphone Playback Switch' on"
+			]
 
-	DisableSequence [
-		cset "name='Headphone Playback Switch' off"
-	]
+			DisableSequence [
+				cset "name='Headphone Playback Switch' off"
+				cset "name='Input Source' 1"
```
perexg: We can use the enum name here. The value '1' is not self-explaining. Use _cset "name='Input Source' Mic"_ or so...

ucm2/sof-hda-dsp/sof-hda-dsp.conf
perexg: Thinking more: We should not probably touch those values. I implemented the master control for PA in https://gitlab.freedesktop.org/pulseaudio/pulseaudio/merge_requests/213 . It seems working nicely here.

The 'Capture Switch' and 'Auto-Mute Mode' settings is fine (although I think that the auto mute settings is available only for some HDA codecs).

perexg:
> Now, I'm not insisting to divide things into separate use cases / profiles, but I'm not sure is there any harm, so is there some conceptional or real issue existing, that you should not do this.

The question is, what's the right way. I need to study / think about this more. We have also modifiers which can be used. I'm not also strict to have more UCM verbs (profiles), it makes a lot of sense. SOF is just the first complex sound card which we need to support in the way like other sound cards without UCM.

> Now, I'm not insisting to divide things into separate use cases / profiles, but I'm not sure is there any harm, so is there some conceptional or real issue existing, that you should not do this.

> There's https://gitlab.freedesktop.org/pulseaudio/pulseaudio/merge_requests/209 to address this but not yet merged.

It appears that it may help to define everything in the one verb (profile) including HDMI and let PA to switch between the possible ports only.

perexg:
> There's https://gitlab.freedesktop.org/pulseaudio/pulseaudio/merge_requests/209 to address this but not yet merged.

It seems that this one is merged to PA master branch now.

ucm2/sof-hda-dsp/HiFi.conf
```diff
@@ -84,6 +72,7 @@ SectionDevice."Speaker" {
 		PlaybackSwitch "Speaker Playback Switch"
 		PlaybackChannels "2"
 		PlaybackPriority "100"
+		JackHWMute "Headphone"
```
perexg: This is PA or the UCM specification issue? The specification says this: _If **JackHWMute** is set, it indicates that when the jack is plugged in, the hardware automatically mutes some other device(s)._

And JackHWMute belongs to the JackControl which is not defined in Speaker device.

juimonen: hmm I don't know actually. With your ucm I was not seeing headset and speakers at the same time ever, so inserting headphones removed the speakers, thus I was not able to manually route to speakers when headset was connected. I was just diffing the differences and this was making it possible. So not sure is it ucm, pulseaudio, or ubuntu ui issue.

So I think this was in your original "Add initial support for generic SOF Skylake driver (DMIC)" import, but was changed during the rename, so actually don't know the originators thinking.

ucm2/sof-hda-dsp/HiFi.conf
juimonen: ok

ucm2/sof-hda-dsp/sof-hda-dsp.conf
juimonen: If I recall correctly this was more for "first boot" situation where pulseaudio doesn't have anything in its databases. So if the PR implements the element chaining that raising the headphone/speaker volume also opens Master, then removing it would be fine.

ucm2/sof-hda-dsp/HiFi.conf
perexg: My changes (trying to keep the one verb only) are in https://github.com/perexg/alsa-ucm-conf/commits/sof , if you like to test. I do tests with the latest alsa-lib and patched PA.

perexg:
Ok, testing my branch more today and it seems that it's the really a way to go. There are two problems remaining:

1) switching to the HDMI playback works correctly only when the PA is restarted after HDMI is plugged in (I'll look into this probably next week); the HDMI PCM device stays in idle for a reason
2) when headphones with a mic are connected, this mic is not used by default (PA keeps DigitalMic, you need to make the switch manually); it's probably because the https://gitlab.freedesktop.org/pulseaudio/pulseaudio/merge_requests/209 code is only for playback; but it's just a guess

juimonen:
@perexg hmm I can't really test atm with latest and greatest as I need to stick for a while with ubuntu 18 LTS. So do you see any commits in this PR which are relevant, cause I can drop/modify it accordingly. I have to say it is pretty difficult to see what issues are caused by differences in ubuntu/fedora UI...

I am after couple of things here:
1) Headset/headphones and Speakers should be visible in the UI same time with each other (I reversed the jack definition between headphones and speakers)
2) System should be able to switch away from HDMI back to available sinks (this was improved by HDMI use case separation)

juimonen:
@perex was trying to test your branch, I get: alsa-ucm.c: Failed to enable ucm device DigitalMic from pulseaudio, everything else seems to go fine, no other message what is actually going wrong... Maybe I'm missing some patch somewhere?

perexg:
No idea. Could you paste logs somewhere? Also, if you have latest alsa-utils (alsaucm), please, paste the UCM configuration dump (alsaucm -c 0 dump text) somewhere to check.

BTW: I fixed the sink state issue (HDMI plug in / plug out problem). I pushed the fix to my PA pull request, too.

juimonen:
@perexg commenting out the following from DigitalMic made it work....
#       ConflictingDevice [                                                                                                                                                                             
#               "HeadsetMic"                                                                                                                                                                            
#       ] 
I should have latest&greates alsa-lib & alsa-utils

juimonen:
ok that paste was _BIG_, sorry about that. So somehow the conflicting device is not good on DigitalMic. And basically it would not be needed... maybe there's someone who wants to use those mics at the same time for some strange reason.

perexg:
The ConflictingDevice section was removed in adeb572487081f4f88318583d6eefb0ea1b61361 for the digital microphone. I also renamed the devices according the documentation in https://github.com/alsa-project/alsa-lib/blob/master/include/use-case.h .

PR #24, opened 2020-04-23
sof-hda-dsp: fix speaker on ThinkPad X1C7

chrisx8:
With the current ucm2 profile at `sof-hda-dsp/HiFi.conf`, built-in speakers on the ThinkPad X1 Carbon Gen 7 sounds muffled/tinny. Muting 'Speaker' and setting 'Speaker' volume to 100 in `alsamixer` fixes this issue.

I applied these fixes in the ucm2 profile, so they don't have to be applied manually.

### DIFF
--- a/ucm2/sof-hda-dsp/HiFi.conf
+++ b/ucm2/sof-hda-dsp/HiFi.conf
@@ -49,7 +49,8 @@ SectionDevice."Speaker" {
 		}
 		True {
 			EnableSequence [
-				cset "name='Speaker Playback Switch' on"
+				cset "name='Speaker Playback Switch' off"
+				cset "name='Speaker Playback Volume' 100%"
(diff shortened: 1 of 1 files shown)


perexg:
Do things working correctly only if the volume is at 100%? Basically, you disable the hardware volume control on DAC and I do have sound on X1 even if the volume is not at 100%.

https://bugzilla.kernel.org/show_bug.cgi?id=207407

perexg:
And this verb is also used for other hardware than X1 carbon.

chrisx8:
Speakers work correctly as long as 'Speaker' is muted. Not setting 'Speaker' to 100% just makes the speakers a bit too quiet.

P.S. I'm also the one who submitted bug report 207407.

perexg:
I think that you see another issue. Let continue to discuss this in the kernel bugzilla.

PR #25, opened 2020-05-03
Add initial support for Realtek ALC1220 TRX40 motherboards

maximlevitsky:
Tested with TRX40 Designare but should work on most TRX40 motherboards
Tested with pulseaudio 13.99.1 (from fedora 31 package)
Tested with kernel 5.7.0-rc3

The following are the notes of the testing and implementation

* Back speaker works in 5.1 mode, however
  pulseaudio doesn't know how to handle mixer controls with
  8 channels, thus 'PlaybackMixerElem' is for now commented out
  This profile at least makes pulseaudio to set volume on this mixer
  element to 100% when it boots
  7.1 mode doesn't work due to lack of jack retasking in the kernel

  Also pulseaudio refuses to set any postive volume on subwoofer on unknown grounds,
  even though on the kernel level it works.

* SPDIF is not tested physically yet. The SPDIF mute mixer control is recognised by
  pulseaudio correctly

* PlaybackPrioriry/CapturePriority seems to have no effect to pulseaudio,
  it doesn't autoswitch inputs when jacks are insterted. You need to switch manually

* When all capture inputs are disconnected, pulseaudio thinks that the UCM2 profile
  is invalid and doesn't use it by default. It can be forced to use it by selecting
  it in the pavucontrol. Or you can for now comment out one of 'JackControl' statements,
  to make it think that one of the jacks is always plugged in.

* Front mic is alway working in mic mode due to lack of jack retasting as well.

* PulseAudio doesn't take advantage of EnableSequence/DisableSequence fully.
  It just enables all the devices regardless of the jack status, and it will disable
  them only when you swith the device profile to Off

* The default volume on capture devices is set to 59% which roughtly repesents
  0 dB amplification according to alsamixer

Signed-off-by: Maxim Levitsky <maximlevitsky@gmail.com>

### DIFF
--- a/ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop-HiFi.conf
+++ b/ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop-HiFi.conf
@@ -1,0 +1,60 @@
+SectionDevice."Speaker" {
+        Comment "Speakers"
+        Value {
+               PlaybackChannels 8
+               PlaybackPriority 100
+               PlaybackPCM "hw:${CardId},0"
+               JackControl "Speaker Jack"
+               PlaybackMixerElem "Speaker"
+        }
+}
+
+SectionDevice."Headphones" {
+        Comment "Front Headphones"
+        Value {
+                PlaybackPriority 300
+                PlaybackPCM "hw:${CardId},1"
+                JackControl "Front Headphone Jack"
+                PlaybackMixerElem "Front Headphone"
+        }
+}
+
+SectionDevice."SPDIF" {
+        Comment "S/PDIF Out"
+        Value {
+                PlaybackPriority 200
+                PlaybackPCM "hw:${CardId},2"
+                PlaybackMixerElem "IEC958"
+        }
+}
+
+SectionDevice."Line" {
+        Comment "Line In"
+        Value {
+                CapturePriority 100
+                CapturePCM "hw:${CardId},0"
+                JackControl "Line Jack"
+                CaptureMixerElem "Line"
+        }
+}
+
+SectionDevice."Mic1" {
+        Comment "Microphone"
+        Value {
+                CapturePriority 200
+                CapturePCM "hw:${CardId},1"
+                JackControl "Mic Jack"
+                CaptureMixerElem "Mic"
+        }
+}
+
+SectionDevice."Mic2" {
+        Comment "Front Microphone"
+        Value {
+                CapturePriority 300
+                CapturePCM "hw:${CardId},2"
+                JackControl "Front Mic Jack"
+                CaptureMixerElem "Front Mic"
+        }
+}
+
--- a/ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop.conf
+++ b/ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop.conf
@@ -1,0 +1,6 @@
+Syntax 2
+Comment "USB-audio on Realtek ALC1220-VB desktop"
+SectionUseCase."HiFi" {
+	File "Realtek-ALC1220-VB-Desktop-HiFi.conf"
+	Comment "Default Alsa Profile"
+}


maximlevitsky:
alsa-info is at http://alsa-project.org/db/?f=795f09597c532c1c84b3c25952728f34a0e44288

perexg:
The config does not pass the ucm-validator test: https://github.com/alsa-project/alsa-tests/tree/master/python/ucm-validator

maximlevitsky:
> The config does not pass the ucm-validator test: https://github.com/alsa-project/alsa-tests/tree/master/python/ucm-validator
> 
> Also, the default volumes and on / off sequences should be removed (PA will control it) if PlaybackMixerElem / CaptureMixerElem values are defined.

It seems to fail at file that I didn't touch:

ucmlib.UcmError: PAZ00/PAZ00.conf: /__w/alsa-ucm-conf/alsa-ucm-conf/alsa-ucm-conf/ucm2/PAZ00/PAZ00.conf:6: Wrong indentation (use tabs to save space!)

I thought it would be nice to have both set the volume to reasonable defaults and then let pulseaudio control it.

Since for main playback channel, PA refuses to work with 8 channel mixer, can we use on/off sequqence as a workaround until pulseaudio is fixed?

perexg:
> It seems to fail at file that I didn't touch:

It's fixed (rebase to latest master - I forgot to commit the fix from my private tree). If you don't want to run ucm-validator out of github, then just do force push to this PR to rerun the github action.

maximlevitsky:
> > It seems to fail at file that I didn't touch:
> 
> It's fixed (rebase to latest master - I forgot to commit the fix from my private tree). If you don't want to run ucm-validator out of github, then just do force push to this PR to rerun the github action.

I will do this in a few hours.

perexg:
> Since for main playback channel, PA refuses to work with 8 channel mixer, can we use on/off sequqence as a workaround until pulseaudio is fixed?

Could you show me the PA error? I would really prefer to fix things rather than keep workarounds somewhere.

maximlevitsky:
> > Since for main playback channel, PA refuses to work with 8 channel mixer, can we use on/off sequqence as a workaround until pulseaudio is fixed?
> 
> Could you show me the PA error? I would really prefer to fix things rather than keep workarounds somewhere.

Apr 30 22:19:38 starship pulseaudio[16510]: W: [pulseaudio] alsa-mixer.c: Volume element 'Speaker' has 8 channels. That's too much! I can't handle that!
Apr 30 22:19:38 starship pulseaudio[16510]: W: [pulseaudio] alsa-ucm.c: Path Speaker is not a volume control

perexg:
Thanks. It would be nice, if you can test this PA code:

 https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/merge_requests/292

maximlevitsky:
> Thanks. It would be nice, if you can test this PA code:
> 
> https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/merge_requests/292

I will gladly do this in the evening, in a few hours!

maximlevitsky:
About the ucm-validator problem - are you sure you pushed the fix? I rebased my commit now and I don't see any updates to the master branch.

About the pulseaudio fix - it works superb.

Note thought that gnome's sound settings seems to be way too broken. Looks like their 7.1 mixer wasn't really tested on real hardware or something. I'll open a bug report to them as soon as I test latest git to eliminate the possibility of outdated version in my fedora 31.
There are at least 2 issues: One is that subwoofer volume is stuck at 0, other is that when you play with balance and fade, the mixer becames out of sync with pulseaudio. When using pavucontrol thought, it works correctly on all channels, and reflects correctly the alsa mixer.

It also often says that it outputs on one device but actually outputs on another when I look at pavucontrol.

About my UCM2 profile, before I send updated pull request, I have few questions:

Do you think that PlaybackPriority/CapturePriority are worth it? Seems that pulseaudio just ignores them and I just picked them more or less to my taste. I might not understand correctly the meaning of this as well.

I did first round of trying to understand the pulseaudio code but I had no luck finding where it makes the UCM2 profile 'unaviable' when all the input jacks are disconnected. This is annoying as hell.
If you have some pointers I'll would be happy to hear.

maximlevitsky:
I see now, this is just my first time I work with pull requests. Its my fork that was up to date, but you did push the fix to the master branch. I updated my branch by force pusing it.

ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop-HiFi.conf
```diff
@@ -0,0 +1,60 @@
...
+               PlaybackPCM "hw:ALC1220VBDT,0"
+               JackControl "Speaker Jack"
+               PlaybackMixerElem "Speaker"
+        }
+}
+
+SectionDevice."Headphone" {
+        Comment "Front Headphone"
+        Value {
+                PlaybackChannels "2"
```
perexg, line 14: Remove 2 channel lines. It's the default now.

ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop-HiFi.conf
```diff
@@ -0,0 +1,60 @@
...
+                PlaybackPCM "hw:ALC1220VBDT,1"
```
perexg, line 15: Use "hw:${CardId},1" for all PCM device specifications (with the correct device index).

ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop-HiFi.conf
```diff
@@ -0,0 +1,60 @@
...
+        Comment "Microphone"
+        Value {
+                CaptureChannels "2"
+                CapturePCM "hw:ALC1220VBDT,1"
+                JackControl "Mic Jack"
+                CaptureMixerElem "Mic"
+        }
+}
+
+SectionDevice."FrontMic" {
```
perexg, line 51: FrontMic is not a valid UCM device name. Use "Mic1" and "Mic2" and write the location to comments.

ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop-HiFi.conf
```diff
@@ -0,0 +1,60 @@
...
+                PlaybackPCM "hw:ALC1220VBDT,1"
+                JackControl "Front Headphone Jack"
+                PlaybackMixerElem "Front Headphone"
+        }
+}
+
+SectionDevice."SPDIF" {
+        Comment "S/PDIF Out"
+        Value {
+                PlaybackPriority 1
```
perexg, line 24: Higher priority means preferred. Use 100 for Speaker, 200 for SPDIF, 300 for Headphones. The logic is that PA should pick the Headphones when connected.

ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop-HiFi.conf
```diff
@@ -0,0 +1,60 @@
...
+        Comment "Speakers"
+        Value {
+               PlaybackChannels "8"
+               PlaybackPCM "hw:ALC1220VBDT,0"
+               JackControl "Speaker Jack"
+               PlaybackMixerElem "Speaker"
+        }
+}
+
+SectionDevice."Headphone" {
```
perexg, line 11: Use "Headphones" here.

ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop-HiFi.conf
```diff
@@ -0,0 +1,60 @@
...
+                CapturePCM "hw:ALC1220VBDT,0"
+                JackControl "Line Jack"
+                CaptureMixerElem "Line"
+        }
+}
+
+SectionDevice."Mic" {
+        Comment "Microphone"
+        Value {
+                CaptureChannels "2"
```
perexg, line 44: Add priority 100 for mic1 and 200 for mic2 (front).

perexg:
> There are at least 2 issues: One is that subwoofer volume is stuck at 0, other is that when you play with balance and fade, the mixer becames out of sync with pulseaudio. When using pavucontrol though, it works correctly on all channels, and reflects correctly the alsa mixer.

I don't know if it's PA issue or another issue. Where do you set balance and fade? In gnome?

> UCM2 profile 'unaviable' when all the input jacks are disconnected. This is annoying as hell.

Hmm: https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/commit/0d50e787f8

You should probably discuss this in the PA project.

tiwai:
About 8 channels: I'm afraid that we need to provide different device definitions per channel setup, and let them conflicting.  Opening 8 channels may conflict with some other retaskable pins, for example.

And I believe that it'd be better to start with a profile that can work with the current PA, i.e. some minimalistic setup. Step by step, we can fill more controls in, once when the fix in PA is actually released. Otherwise we'll be caught with a broken profile until PA's release and its distribution.

maximlevitsky:
> > There are at least 2 issues: One is that subwoofer volume is stuck at 0, other is that when you play with balance and fade, the mixer becames out of sync with pulseaudio. When using pavucontrol though, it works correctly on all channels, and reflects correctly the alsa mixer.
> 
> I don't know if it's PA issue or another issue. Where do you set balance and fade? In gnome?

Yep - it is only broken in gnome mixer - and in pavucontrol is works just fine - After discovering that the 'lock' icon unlocks per channel volumes I could set each channel's volume separately and see
it reflected in alsamixer, including subwoofer.

About the second issue, also  when the gnome's volume balance/fade becomes out of sync with alsamixer, looking at pavucontrol confirms that, it shows the unbalanced volumes that match exactly what alsamixer shows
thus its the gnome sound setting bug as well.

It must be an issue in gnome.

> 
> > UCM2 profile 'unaviable' when all the input jacks are disconnected. This is annoying as hell.
> 
> Hmm: https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/commit/0d50e787f8
> 
> You should probably discuss this in the PA project. EDIT: Perhaps, we can turn this logic off for UCM.

This is an awesome pointer - I will discuss this with PA project

Thanks a lot for the info about the priorities and the rest of the feedback! 
I'll send an updated pull request in the evening.

maximlevitsky:
> About 8 channels: I'm afraid that we need to provide different device definitions per channel setup, and let them conflicting. Opening 8 channels may conflict with some other retaskable pins, for example.
> 
> And I believe that it'd be better to start with a profile that can work with the current PA, i.e. some minimalistic setup. Step by step, we can fill more controls in, once when the fix in PA is actually released. Otherwise we'll be caught with a broken profile until PA's release and its distribution.

It seems to work though - the side channel is just not doing anything and the jack runs as line-in instead.

The only big issue with pulseaudio currently is that it doesn't understand the 8 channel mixer, and the workaround I had is to use  on / off sequences instead to at least set that mixer to 100% when pulseaudio starts. The proposed patch which I tested yesterday fixes this cleanly.

The other issues with pulseaudio which I afraid can't be solved with an UCM profile is that it is just not ready to see multiple 'sound cards', each with only one port which is often not connected, especially when all the capture ports are disabled. It overall just doesn't like it. For instance it refuses to switch to a just plugged input, based on these grounds - I tested and it seems to auto switch between ports of same sound card but not ports of different sound card - regardless of Playback/Capture priority.

ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop-HiFi.conf
maximlevitsky: Done

ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop-HiFi.conf
maximlevitsky: Done, but note that it seems that PA doesn't pick this up. I think it only resolves priority of multiple ports of the same sound card, but here each playback/capture stream appears as a different sound card.

ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop-HiFi.conf
maximlevitsky: Done

ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop-HiFi.conf
maximlevitsky: In this case I think we should give Line-In also some priority, so I gave it priority of 100 and for these I gave 200 and 300 accordingly

ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop-HiFi.conf
maximlevitsky: Done

maximlevitsky:
> > 
> > Hmm: https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/commit/0d50e787f8
> > You should probably discuss this in the PA project. EDIT: Perhaps, we can turn this logic off for UCM.
> 
> This is an awesome pointer - I will discuss this with PA project
> 
> Thanks a lot for the info about the priorities and the rest of the feedback!
> I'll send an updated pull request in the evening.

Manually reverting this commit, finally gets rid of this behavior.
I''l send an email to PA mailing list now.

I updated the pull request with all the feedback. Thanks a lot for the help!

maximlevitsky:
Any update on this pull request?

perexg:
Merged to 48b11a4c33f267c8ffd9a6c584c3e651eda7e5ec with cosmetic changes (indentation, trailing zero in the PCM device).

maximlevitsky:
Thank you very much!

shrisha108:
@maximlevitsky  Hello Maxim,
I'm unlucky owner of Gigabyte TRX40 Master motherboard with this weird sound card. Unfortunately digital out still not functioning properly. I reported bug 209223 at  kernel.org but looks like nobody looked at it yet. Is there is anyway to debug this issue? Could you please give some directions on this? Spent all last week to find anything on this with no results what so ever.

radioactiveman:
@perexg: Your commit introduced a regression for users with ASUS Xonar sound cards. The sound is now simply gone. I have filed a bug report here: https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/issues/1009

Thaodan:
@shrisha108 Did you found a fix? On Arch it doesn't work for me. I tried a Fedora life dvd (33) it worked there if I switched to the headset port.

Thaodan:
Ok found the issue 5.1 analog is broken.

Thaodan:
With that patch https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/merge_requests/389/diffs?commit_id=b530aa4681087bcb315890df31ec91b4eb3cd4cc applied the  other channels work however the 5.1 profile is still dead.

perexg:
I suggest to create PA issue with the debug log attached for the 5.1 profile activated.

Thaodan:
Will do that however the issue is that even on the stereo profile the sound 
from the speakers is shown as coming from the headphones.

shrisha108:
@Thaodan Hey there. No I'm not. :( In pulse audio Volume control app I have two spdifs out and I can physically see red light is on in the cable but no sound.
spdif
And only stereo profile available in configuration. I'm on fedora 33.

perexg:
@shrisha108 : Could you show me URL for your bugzilla.kernel.org bug (or output from alsa-info.sh sound system info script)? Perhaps, your PCI ID is not registered in the USB driver to modify the routing. It's required for TRX40.

Thaodan:
You have the wrong profile for the USB-Audio device.

Thaodan:
> I suggest to create PA issue with the debug log attached for the 5.1 profile activated.
> 
> EDIT: Also, if you see the 5.1 profile, the UCM is not activated on your machine for a reason.

Hm what do you mean by that?

perexg:
> > I suggest to create PA issue with the debug log attached for the 5.1 profile activated.
> > EDIT: Also, if you see the 5.1 profile, the UCM is not activated on your machine for a reason.
> 
> Hm what do you mean by that?

The UCM profile has name "Default Alsa Profile" and /proc/asound/cards entry should looks like:

```
 2 [ALC1220VBDT    ]: USB-Audio - ALC1220-VB-DT
                      Realtek-ALC1220-VB-Desktop
```

If you don't see Realtek-ALC1220-VB-Desktop as the long sound card name, then the USB driver does not set the abstracted routing correctly. Code: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/sound/usb/card.c?h=v5.11-rc1#n372

shrisha108:
@perexg Thank you.
Here is a output of alsa-info.sh http://alsa-project.org/db/?f=be904496b92f88df9256e6552e7147d62d6e0a13

shrisha108:
@Thaodan 
There are three options Digital Stereo IEC958 output, Digital Stereo IEC958 input and Digital Stereo duplex or OFF. Rest is unavailable.

shrisha108:
Guys, is there any chance for fix?

perexg:
@shrisha108 : You have a bit different model and the USB audio driver already contains a quirk for it:

```
 2 [Audio          ]: USB-Audio - Aorus Master Main Audio
                      Gigabyte-Aorus-Master-Main-Audio
 4 [Headphone      ]: USB-Audio - Aorus Master Front Headphone
                      Gigabyte-Aorus-Master-Front-Headphone
```

 It seems that the jack detection does not work as expected:

```
control.22 {
		iface CARD
		name 'Line Out Jack'
		value false
		comment {
			access read
			type BOOLEAN
			count 1
		}
	}
```

```
	control.3 {
		iface CARD
		name 'Headphone - Output Jack'
		value false
		comment {
			access read
			type BOOLEAN
			count 1
		}
	}
```

You may check the Jack detection using `amixer -c <cardnumber> events`. From the above, the card numbers are 2 and 4 for the onboard USB audio devices.

@tiwai : The `Headphone - Output` prefix seems like a wrong control name (at least PA don't have this name in the database). It would be probably nice to rename this to `Headphone` in the USB driver.

Thaodan:
>@tiwai : The Headphone - Output prefix seems like a 
wrong control name (at least PA don't have this name in 
the database). It would be probably nice to rename this 
to Headphone in the USB driver.

It would be also helpful to apply a quirk to rename that 
to be in line with the physical port it represents.

shrisha108:
Thanks so much for input but honestly I'm not certain what and how to do it :(
Here is output of amixer event command:
[shrisha@fedora-home]$ amixer -c 2 events
event add: numid=10,iface=CARD,name='Clock Source 1 Validity'
event add: numid=14,iface=CARD,name='Clock Source 2 Validity'
event add: numid=18,iface=CARD,name='Clock Source 3 Validity'
event add: numid=21,iface=CARD,name='Clock Source 4 Validity'
event add: numid=23,iface=CARD,name='Clock Source 5 Validity'
event add: numid=15,iface=CARD,name='Front Mic Jack'
event add: numid=24,iface=CARD,name='Keep Interface'
event add: numid=7,iface=CARD,name='Line Jack'
event add: numid=22,iface=CARD,name='Line Out Jack'
event add: numid=11,iface=CARD,name='Mic Jack'
event add: numid=16,iface=MIXER,name='Front Mic Capture Switch'
event add: numid=17,iface=MIXER,name='Front Mic Capture Volume'
event add: numid=19,iface=MIXER,name='Line Out Playback Switch'
event add: numid=20,iface=MIXER,name='Line Out Playback Volume'
event add: numid=8,iface=MIXER,name='Line Capture Switch'
event add: numid=9,iface=MIXER,name='Line Capture Volume'
event add: numid=12,iface=MIXER,name='Mic Capture Switch'
event add: numid=13,iface=MIXER,name='Mic Capture Volume'
event add: numid=1,iface=PCM,name='Capture Channel Map'
event add: numid=4,iface=PCM,name='Playback Channel Map'
event add: numid=2,iface=PCM,name='Capture Channel Map',device=1
event add: numid=5,iface=PCM,name='Playback Channel Map',device=1
event add: numid=3,iface=PCM,name='Capture Channel Map',device=2
event add: numid=6,iface=PCM,name='Playback Channel Map',device=3
Ready to listen...

[shrisha@fedora-home]$ amixer -c 4 events
event add: numid=2,iface=CARD,name='Clock Source 4 Validity'
event add: numid=3,iface=CARD,name='Headphone - Output Jack'
event add: numid=4,iface=CARD,name='Keep Interface'
event add: numid=1,iface=PCM,name='Playback Channel Map'
Ready to listen...
I had no problem with detection of analog jacks to front and back audio sockets. Only problem with optical SPDIF. And optical cable is plugged in right now.

perexg:
Sorry, I misread your initial request. So you have problem with the digital output.

```
card 2: Audio [Aorus Master Main Audio], device 0: USB Audio [USB Audio]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 2: Audio [Aorus Master Main Audio], device 1: USB Audio [USB Audio #1]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 2: Audio [Aorus Master Main Audio], device 3: USB Audio [USB Audio #3]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
```

In this case, the UCM configuration is required to describe the digital output device. Let's continue in PR #72 . Could you test? Just copy the added files to the /usr/share/alsa/ucm2/USB-Audio directory and restart pulseaudio. If you have issues, show the verbose pulseaudio debug log (link only) - https://fedoraproject.org/wiki/How_to_debug_PulseAudio_problems .

shrisha108:
Sure, Will do my best

MarijnS95:
On this note, is optical SPDIF supposed to have some form of jack detection? I remember commenting out the section from `Realtek-ALC1220-VB-Desktop-HiFi.conf` way back when to prevent PA from starting with that as default output device or even allow switching to that sink when nothing is connected.

Even now no `IEC958 Jack` or similar appears in `amixer`, just this:

<details>
<summary>amixer -c0 controls</summary>

```console
$ amixer -c0 controls
numid=11,iface=CARD,name='Clock Source 1 Validity'
numid=15,iface=CARD,name='Clock Source 2 Validity'
numid=19,iface=CARD,name='Clock Source 3 Validity'
numid=22,iface=CARD,name='Clock Source 4 Validity'
numid=26,iface=CARD,name='Clock Source 5 Validity'
numid=29,iface=CARD,name='Clock Source 6 Validity'
numid=27,iface=CARD,name='Front Headphone Jack'
numid=16,iface=CARD,name='Front Mic Jack'
numid=30,iface=CARD,name='Keep Interface'
numid=8,iface=CARD,name='Line Jack'
numid=12,iface=CARD,name='Mic Jack'
numid=23,iface=CARD,name='Speaker Jack'
numid=24,iface=MIXER,name='Front Headphone Playback Switch'
numid=25,iface=MIXER,name='Front Headphone Playback Volume'
numid=17,iface=MIXER,name='Front Mic Capture Switch'
numid=18,iface=MIXER,name='Front Mic Capture Volume'
numid=9,iface=MIXER,name='Line Capture Switch'
numid=10,iface=MIXER,name='Line Capture Volume'
numid=13,iface=MIXER,name='Mic Capture Switch'
numid=14,iface=MIXER,name='Mic Capture Volume'
numid=28,iface=MIXER,name='IEC958 Playback Switch'
numid=20,iface=MIXER,name='Speaker Playback Switch'
numid=21,iface=MIXER,name='Speaker Playback Volume'
numid=1,iface=PCM,name='Capture Channel Map'
numid=4,iface=PCM,name='Playback Channel Map'
numid=2,iface=PCM,name='Capture Channel Map',device=1
numid=5,iface=PCM,name='Playback Channel Map',device=1
numid=3,iface=PCM,name='Capture Channel Map',device=2
numid=6,iface=PCM,name='Playback Channel Map',device=2
numid=7,iface=PCM,name='Playback Channel Map',device=3
```
</details>

Is this missing from the kernel driver by chance?

perexg:
I am not aware about hardware which has the plug detection for the optical I/O. For analog, the wires resistence is detected, but for the optical connection, it's not possible to use this method.

Perhaps, we may remove the jack detection for speakers in UCM (remove `JackControl "Speaker Jack"` in config and change priority for SPDIF - lower than Speakers), so the speakers will be selected even if they are not connected. It's standard way for most of hardware. Could you test?

MarijnS95:
@perexg I expected there to be some form of sensing in the plug, but unfortunately even on Windows the "digital out" device says `Ready` instead of `Disconnected` or `Not plugged in`.

I can change priorities but I don't think we need to remove jack detection, that would be annoying for users with only SPDIF plugged in with no gain for users with 3.5mm analog rear plugged in. I'll test and submit a PR if it works out for me!

MarijnS95:
@perexg Done, works!

This might be the right place to discuss some other random issues with this sound card. On boot and after suspend from idle (which hardly ever works anyway on this MSI TRX40 Creator...) volume is extremely loud and sometimes only on the right channel, but hardware controls (`amixer -cXX contents`) report normal volume. A simple volume change gets it back to normal: is the kernel driver simply not restoring a proper/previous volume when enabling the card?

Other times, during normal usage, sound might randomly go to white-noise mode. It seems like there's a sudden software amplification (overflow) going on because the sound does not get much louder, but sounds "normal" when hardware volume is turned down to about 2% volume (in PulseAudio). Switching to and from the off profile resolves this issue. Sometimes it doesn't happend for a full day, just had a boot where it happened about every 5 min. Anyone familiar with this?

(Note again that this is all using hardware voume controlos)

PR #20, opened 2020-04-13
SOF support for Baytrail/CherryTrail

plbossart:
Add symlink and simple test to reuse legacy settings as is.

tested on Baytrail/Asus T100TA only with SOF and legacy drivers, all other platforms are just based on same changes.

sof-bytcr-rt5640 is here:
 http://alsa-project.org/db/?f=6f8be0520e2bbda94cb744ce66232174feec03d4

UCM validator seems to have problems with the conditional statement?

PR #26, opened 2020-05-26
ucm2: Add profile for Chromebook Asus C300

tiwai:
ASUS Chromebook C300 alias Google QUAWKS is an Intel Baytrail platform
with max98090 codec.  This patch adds the basic UCM snippet for the
max98090 codec and HiFi.conf for this model.

Note that MIC2 is used for the headset mic.  If another model with
this codec uses a different ADC port, we'd need to create another
profile snippet and rename the device.

BugLink: https://apibugzilla.suse.com/show_bug.cgi?id=1171492
Signed-off-by: Takashi Iwai <tiwai@suse.de>

### DIFF
--- a/ucm2/chtmax98090/HiFi-Quawks.conf
+++ b/ucm2/chtmax98090/HiFi-Quawks.conf
@@ -1,0 +1,18 @@
+SectionVerb {
+	Value {
+		TQ "HiFi"
+	}
+
+	EnableSequence [
+		<platforms/bytcr/PlatformEnableSeq.conf>
+		<codecs/max98090/EnableSeq.conf>
+		# volume values modified
+		cset "name='Headphone Volume' 25"
+		cset "name='Speaker Volume' 32"
+	]
+}
+
+<codecs/max98090/Headphones.conf>
+<codecs/max98090/Speaker.conf>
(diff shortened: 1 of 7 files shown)


ucm2/chtmax98090/HiFi-Quawks.conf
```diff
@@ -0,0 +1,18 @@
+SectionVerb {
+	Value {
+		TQ "HiFi"
+	}
+
+	EnableSequence [
+		<platforms/bytcr/PlatformEnableSeq.conf>
```
plbossart, line 7: if you ever want to use SOF on this device this would need to change, see https://github.com/alsa-project/alsa-ucm-conf/pull/20

Also I quickly checked the values, they all seem to match what we used so we should just add a default chtmax98090/chtmax98090.conf that calls a basic HiFi.conf and this would work fine on Cyan as well.

perexg: I added the condition to b081ac49ad9cbea298c3dcc6f029f331e369a213 , but the SOF/HiFi.conf should be modified for this platform, too.

Any generic cleanups would be fine, of course.

perexg:
Merged. Thank you.

PR #29, opened 2020-06-02
Db845c

Srinivas-Kandagatla:
Hi Jaroslav Kysela,

I removed Lenovo ucm2 configs for now, are you okay to pull DragonBoard DB845c ucm2 for now till we resolve in general naming issues with DT.  All the drivers and DT are already in mainline since 5.6.

Thanks,
srini
Changes sinc v2:
- remove Lenovo ucm2
- remove cdev references.


perexg:
I am not very happy, that we have not a proper driver name even for the Dragon Boards. DB845c is not really the correct driver name.

I merge this only because we have already DB410c/DB820c in the tree so we need a transition to the new scheme when ready.

As preparation, I moved the path description to the configuration file: https://github.com/alsa-project/alsa-ucm-conf/blob/master/ucm2/ucm.conf , so we can redirect the old drivers to the final path. But this exception list should be minimal.

Merged to master.

PR #34, opened 2020-06-11
sof-hda-dsp: make Headphone Playback Switch conditional

kv2019i:
The Headphone Playback Switch control is not present in all
HDA codecs. Allow the Headphones definition to work also on
such systems.

Signed-off-by: Kai Vehmanen <kai.vehmanen@linux.intel.com>

### DIFF
--- a/ucm2/sof-hda-dsp/HiFi.conf
+++ b/ucm2/sof-hda-dsp/HiFi.conf
@@ -9,13 +9,20 @@ SectionVerb {
 SectionDevice."Headphones" {
 	Comment "Headphones"
 
-	EnableSequence [
-		cset "name='Headphone Playback Switch' on"
-	]
-
-	DisableSequence [
-		cset "name='Headphone Playback Switch' off"
-	]
+	If.headphone_switch {
+		Condition {
+			Type ControlExists
+			Control "name='Headphone Playback Switch'"
+		}
+		True {
+			EnableSequence [
+				cset "name='Headphone Playback Switch' on"
+			]
+			DisableSequence [
+				cset "name='Headphone Playback Switch' off"
+			]
+		}
+	}
 
 	Value {
 		PlaybackPriority 200


perexg:
I think that the whole "Headphones" device should be skipped (not defined) in this case. How you can distinguish Speakers and Headphones without the switch?

kv2019i:
@perexg wrote:
> I think that the whole "Headphones" device should be skipped (not defined) in this case. How you can distinguish Speakers and Headphones without the switch?

This is quite rare case for sure, but I hit this on a development board that has no speaker. It does have a headphone out, and it has working jack detection, just with no "Headphone Playback Switch". If I remove "Headphones" completely, there's no way to get audio out rather than HDMI.

It of course depends on how the codec is set up by BIOS and I wasn't entirely sure whether this is worth fixing in upstream, but it would seem for the generic HDA UCM file, we should be very defensive and make minimal assumptions on codec setup.

Btw, thanks for doing the UCM2 extensions. It's really paying off now and allows to do this kind of variations much more efficiently than trying to cover each combination with a unique UCM1 file. For most part, the generic UCM2 files just work.

kv2019i:
Btw @perexg -- I have one more patch for Speaker missing (didn't send a pull-req yet):
https://github.com/kv2019i/alsa-ucm-conf/commit/0cd82e91b04028ca9eb36ac2769b5d9fa7947c77

I was thinking that even on systems like this where the HDA codec controls are barebones, I'd keep one "hw:CARD,0" device available in the UCM definition for generic SOF HDA. These are likely to be systems with just a headphone connector.

Alsa-mixer contents for one such system:
https://github.com/thesofproject/linux/files/4770259/amixercontents0612.txt

perexg:
Merged to 6b0216d7b9ed5094a7bf28e6f643d90dd7e6fa62 . Thanks.

PR #37, opened 2020-06-30
ucm2: Fixes for moving DragonBoard UCM files.

Srinivas-Kandagatla:
Hi Jaroslav,

Nico just noticed that you have moved and renamed all the DragonBoard ucm2 files.

We have below few issues with that patches:

1> DB820c belongs to APQ8096 SoC, so it should be moved to ucm2/apq8096 folder rather than apq8016.

2> DB410c belongs to APQ8016 SoC, so it should be under ucm2/apq8016 rather than apq8016-sbc.

3> All the board specific ucm conf file names with actual board file are now replaced with SoC names, like DB845c is replace with sdm845. Any particular reason for doing this? Are you suggesting us to use some matching on card long names to include correct board files from top conf file? Can you provide us some pointers on how board specific config files should endup looking like?

This PR fixes (1), and (2).

Thanks,
srini


Srinivas-Kandagatla:
closing this as similar change has been already applied!

PR #39, opened 2020-07-08
ucm2: Add config for Rockchip/rk3399-gru-sound

alpernebbi:
This is the internal card for a number of ChromeOS devices based on the Gru board. Config is taken from ChromeOS sources [1] and modified to mostly look like Rockchip/max98090. Tested on a Samsung Chromebook Plus (rk3399-gru-kevin) running Linux 5.7+.

[1] https://chromium.googlesource.com/chromiumos/overlays/board-overlays/+/75cf7da335c11469956c84ddfa4e2ca73b268441/overlay-kevin/chromeos-base/chromeos-bsp-kevin/files/audio-config/ucm-config/rk3399-gru-sound

Signed-off-by: Alper Nebi Yasak \<alpernebiyasak@gmail.com\>

---

I've tested this on a Debian installation (with alsa-lib version 1.2.2) meaning I had to put the two files to /usr/share/alsa/ucm2/rk3399-gru-sound/ instead of .../Rockchip/rk3399-gru-sound, so I hope the latter is a correct path for them with later versions.

All four devices work, but I couldn't get the Speaker/Headphones to play different things or the Mic/Headset to record simultaneously, so I marked those pairs as "ConflictingDevice"s. Without that, e.g. I see all devices in pulseaudio, but when I play two different audio files to Speaker/Headphones sinks I hear an alternating mix of both files on both devices; and both Mic/Headset sources end up recording only from the headset microphone in an alternating way. 

Pulseaudio doesn't recognize the headset jack, but acpi_listen does report "jack/headphone HEADPHONE plug" events. I'll look into this more, my best guess now is that the kernel doesn't expose the jack in a way it should.

I also uploaded the alsa-info.sh output to cardinfo database:
http://alsa-project.org/db/?f=9b4102975fa9c0f109af6ad687c856c620d76909

### DIFF
--- a/ucm2/ASUS Google Nexus 7 ALC5642/ASUS Google Nexus 7 ALC5642.conf
+++ b/ucm2/ASUS Google Nexus 7 ALC5642/ASUS Google Nexus 7 ALC5642.conf
@@ -1,0 +1,8 @@
+# Use case Configuration for ASUS Google Nexus 7 (2012)
+
+Syntax 2
+
+SectionUseCase."HiFi" {
+	File "HiFi.conf"
+	Comment "Play HiFi quality Music"
+}
--- a/ucm2/ASUS Google Nexus 7 ALC5642/HiFi.conf
+++ b/ucm2/ASUS Google Nexus 7 ALC5642/HiFi.conf
@@ -1,0 +1,105 @@
+# Use case Configuration for ASUS Google Nexus 7 (2012)
+# Based on codecs/rt5640/*
+
+SectionVerb {
+	EnableSequence [
+		# RT5640 Output Configuration
+		cset "name='Speaker Playback Volume' 15"
+		cset "name='HP Playback Volume' 10"
+		cset "name='DAC1 Playback Volume' 175"
+
+		# RT5640 Output Routing
+		cset "name='OUT MIXL DAC L1 Switch' on"
+		cset "name='OUT MIXR DAC R1 Switch' on"
+		cset "name='SDI select' 0"
+		cset "name='DAI select' 0"
+		cset "name='DAC2 Playback Switch' on"
+		cset "name='DIG MIXL DAC L2 Switch' on"
+		cset "name='DAC MIXL INF1 Switch' on"
+		cset "name='DAC MIXR INF1 Switch' on"
+ 		cset "name='Stereo DAC MIXL DAC L1 Switch' on"
+ 		cset "name='Stereo DAC MIXR DAC R1 Switch' on"
+		cset "name='Stereo DAC MIXL DAC L2 Switch' on"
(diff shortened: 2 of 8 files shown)


perexg:
Thank you for your config. The recent version use the kernel module as the primary identifier. There should be symlink from ucm2/module to the top level ucm configuration file (like `ucm2/module/snd_soc_rk3399_gru_sound.conf -> ../Rockchip/rk3399-gru-sound/rk3399-gru-sound.conf`). What's symlink contents of `/sys/class/sound/card0/device/driver` file in this system?

alpernebbi:
I have:
```
$ readlink /sys/class/sound/card0/device/driver
../../../bus/platform/drivers/rk3399-gru-sound
```

And I do see `snd_soc_rk3399_gru_sound` in `lsmod` output. I had completely missed the module dir. So I only need to create that exact symlink?

Another thing that weirds me out is, going by the README it looks like I'd have `rk3399-gru-soun` (?) as the driver name:
```
$ cat /proc/asound/cards
 0 [rk3399grusound ]: rk3399-gru-soun - rk3399-gru-sound
                      rk3399-gru-sound
```

perexg:
> I have:
> 
> ```
> $ readlink /sys/class/sound/card0/device/driver
> ../../../bus/platform/drivers/rk3399-gru-sound
> ```
> 
> And I do see `snd_soc_rk3399_gru_sound` in `lsmod` output. I had completely missed the module dir. So I only need to create that exact symlink?

The ucm2/ucm.conf gets the last part of the path from link using the /sys/class/sound/card0/device/driver path. So the link should be `rk3399-gru-sound.conf -> ./Rockchip/rk3399-gru-sound/rk3399-gru-sound.conf`.

 > Another thing that weirds me out is, going by the README it looks like I'd have `rk3399-gru-soun` (?) as the driver name:
> 
> ```
> $ cat /proc/asound/cards
>  0 [rk3399grusound ]: rk3399-gru-soun - rk3399-gru-sound
>                       rk3399-gru-sound
> ```

Yes, the ALSA driver name is limited (16 bytes including the trailing zero) and the ASoC core tries to be "clever" enough to strip the last characters. The driver should set more nice driver name in this case.

I tried to update README.md to clarify things.

alpernebbi:
> So the link should be `rk3399-gru-sound.conf -> ./Rockchip/rk3399-gru-sound/rk3399-gru-sound.conf`.

I'm assuming that's still in ucm2/module. I did `ln -s ../Rockchip/rk3399-gru-sound/rk3399-gru-sound.conf ucm2/module/rk3399-gru-kevin.conf` and pushed.

perexg:
Merged in 79c21c132da68849c429f7238070567d3dd001dc and https://github.com/alsa-project/alsa-tests/commit/092055a58b9fdfd192a0660b7c3739890ee1edb0 . Than you.

PR #40, opened 2020-07-14
Remove cset value quotation

libinyang:
This patchset removes the quotation marks of the cset values in the ucm2 config files.
In ucm2 files, the cset values should not use quotation marks. Otherwise, the cset won't work.


libinyang:
The first 2 patches are tested. The other 2 patches are not tested, but they should be necessary based on the first 2 patches test result.

PR #41, opened 2020-07-21
amd/renoir/acp: use the machine driver's name for top-level's conf

jason77-wang:
On the machines with amd renoir audio, the /sys/class/sound/card2/
device/driver links to /sys/bus/platform/drivers/acp_pdm_mach.

### DIFF
--- a/ucm2/HDA-Intel/HiFi.conf
+++ b/ucm2/HDA-Intel/HiFi.conf
@@ -9,24 +9,25 @@ If.analog {
 		Type ControlExists
 		Control "name='Master Playback Switch'"
 	}
-	True.Include.analog.File "/HDA-Intel/HiFi-analog.conf"
+	True {
+		Include.analog.File "/HDA-Intel/HiFi-analog.conf"
+		If.acp {
+			Condition {
+				Type String
+				Empty "${var:AcpCardId}"
+			}
+			True {
+				RenameDevice."Mic1" "Mic"
+			}
+			False.Include.acp {
+				Before.SectionDevice "Mic1"
+				File "/HDA-Intel/HiFi-acp.conf"
+			}
+		}
+	}
 }
 
 If.hdmi {
 	Condition { Type String Empty "" }
 	True.Include.hdmi.File "/HDA-Intel/Hdmi.conf"
-}
-
-If.acp {
-	Condition {
-		Type String
-		Empty "${var:AcpCardId}"
-	}
-	True {
-		RenameDevice."Mic1" "Mic"
-	}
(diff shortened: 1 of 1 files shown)


jason77-wang:
Weird, the the ucm_validator reported there are 3 errors in the 3rd patch, but I tested on the lenovo machines with acp, all worked as expected. I have no idea why validator report errors on it.

perexg:
Applied first two patches (thanks). What does not work for HDA HDMI over UCM? It may be good to start with those new devices to move to UCM instead using the pulseaudio probe mechanism.

jason77-wang:
If we could merge this fix https://github.com/alsa-project/alsa-lib/pull/79, then we could revert the 1st patch (amd/renoir/acp: use the machine driver's name for top-level's conf ).

jason77-wang:
> Applied first two patches (thanks). What does not work for HDA HDMI over UCM? It may be good to start with those new devices to move to UCM instead using the pulseaudio probe mechanism.

For the 3rd patch (HDA-Intel: only bind the acp dmic to the soundcard with analog codec), it is ok to let nvidia hdmi sound card use UCM, but for the amd renoir machine, the current ucm will add an acp dmic to nvidia hdmi sound card too. So both Intel sound card and nvidia hdmi sound card have the acp dmic. We could fix it with the 3rd patch or of course we could fix it with adding a condition to include acp dmic. If you prepare the latter solution, I will prepare a new patch. thx. 

And attach a log of 'pacmd list-cards' with the current ucm on the amd renoir machine:

jason77-wang:
This is the log:

[log.txt](https://github.com/alsa-project/alsa-ucm-conf/files/5094012/log.txt)

perexg:
> If we could merge this fix [alsa-project/alsa-lib#79](https://github.com/alsa-project/alsa-lib/pull/79), then we could revert the 1st patch (amd/renoir/acp: use the machine driver's name for top-level's conf ).

Thanks for this report. I just used the wrong sysfs path in ucm2/ucm.conf to determine the kernel module name.

> We could fix it with the 3rd patch or of course we could fix it with adding a condition to include acp dmic. If you prefer the latter solution, I will prepare a new patch. thx.

Yes, it sounds better. Thank you.

jason77-wang:
pushed the modified patch.

To remove the validation error, we could remove the if acp True block since so far the HiFi.conf is only included when there is acp sound card.or ADD a new item in the SUPPRESS_IF.

perexg:
Merged in 4c703a5504deb1772cadfdc783257d9a927b422b . Thank you.

PR #47, opened 2020-08-17
USB-Audio: Dell-WD15-Dock: make input and output volume adjustable

jason77-wang:
Recently we found the input volume is too low for some specific
headset-mic, even we adjust the input volume to max from PA, we still
can't record the sound from that headset-mic. That is because we
change the input or output volume from PA, but the mixer's volume
is not changed, only PA's soft-volume is changed.

Checking the amixer controls for this sound card, it supports
MixerElem, Volume and Switch, So adding them in the ucm.


perexg:
Merged in 15a3ab694c703530e0d9317116be5d28129986e0 . Thank you.

PR #48, opened 2020-08-27
HDA-Intel/HiFi-dual: Add EnableSequence and DisableSequence for Speak…

khfeng:
…er and Headphones

### DIFF
--- a/ucm2/HDA-Intel/HiFi-dual.conf
+++ b/ucm2/HDA-Intel/HiFi-dual.conf
@@ -27,6 +27,14 @@ SectionDevice."Speaker" {
 		PlaybackMixerElem "Speaker"
 	}
 
+	EnableSequence [
+		cset "name='Speaker Playback Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='Speaker Playback Switch' off"
+	]
+
 	ConflictingDevice [
 		"Headphones"
 	]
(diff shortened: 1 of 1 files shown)


jason77-wang:
Since the PA suppose all ucm devices are initially disabled, we could disable the needed ucm devices initially, then we could drop the changes of https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/merge_requests/357

    --- a/ucm2/HDA-Intel/HiFi-dual.conf
    +++ b/ucm2/HDA-Intel/HiFi-dual.conf
    @@ -8,6 +8,8 @@ SectionVerb {
                    cset "name='Front Playback Switch' on"
                    cset "name='Rear-Panel Capture Volume' 100%"
                    cset "name='Rear-Panel Capture Switch' on"
    +               cset "name='Headphone Playback Switch' off"
    +               cset "name='Speaker Playback Switch' off"
            ]
 
            DisableSequence [

perexg:
If PlaybackMixerElem is defined, the PA should set volume/switch automatically for the device. I miss the point of this patch.

khfeng:
> If PlaybackMixerElem is defined, the PA should set volume/switch automatically for the device. I miss the point of this patch.

Does this also apply to alsaucm? The following command doesn't enable/disable Headphones unless with this patch:
```$ alsaucm -c HDA-Intel set _verb HiFi set _enadev Headphones set _disdev Headphones```

khfeng:
On a laptop with sof-hda-dsp, if `EnableSequence` and `DisableSequence` are commented out under `Headphones` and `Speaker`, plugging a headphone no longer mute speaker. PulseAudio doesn't seem to use other means to enable/disable UCM devices.

perexg:
You're right. We need to handle the enable/disable states in UCM. I added your original patch and a change including the correct BootSequence to the repo. Please, report, if you have any issues with it.

PR #49, opened 2020-08-31
ucm2: Qualcomm sdm845 fixes HDMI select card and HiFi set DigitalVol

alimon:
Signed-off-by: Aníbal Limón <anibal.limon@linaro.org>

### DIFF
--- a/ucm2/Qualcomm/sdm845/HDMI.conf
+++ b/ucm2/Qualcomm/sdm845/HDMI.conf
@@ -20,7 +20,7 @@ SectionDevice."HDMI" {
 	Comment "HDMI Digital Stereo Output"
 
 	Value {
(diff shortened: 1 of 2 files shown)


perexg:
Merged in 3048ff7b15a922985499fa04b12b6abe929b8e4a  except the HDMI.conf change which is no-op (and my ucm validator does not like it). Thanks.

perexg:
Wow. I was too quick. You cannot use PlaybackVolume twice. Only the last value is defined then ("RX2 Digital Volume"). Is sufficient to control only this volume for this device?

alimon:
@perexg Right, fixed, thx.

jkysela-rh:
Merged in eda02cf25751cc4792c5b160f42325c695342807 .

PR #57, opened 2020-10-29
Fix initsetting

jason77-wang:
This issue was found when we prepared ubuntu 20.10. When users install the OS to a machine with HDA codec and use sof driver, the audio output is muted after installation. And even we set the master playback switch to on, the output volume from speaker is too low.

This issue could be easily reproduced by:
sudo rm /var/lib/asla/*
rm ~/.config/pulse/* (optional)
sudo /usr/sbin/alsactl -E HOME=/run/alsa -E XDG_RUNTIME_DIR=/run/alsa/runtime restore

### DIFF
--- a/ucm2/HDA-Intel/init.conf
+++ b/ucm2/HDA-Intel/init.conf
@@ -14,7 +14,7 @@ If.master {
 		Control "name='Master Playback Volume'"
 	}
 	True.BootSequence [
-		cset "name='Master Playback Volume' 60%"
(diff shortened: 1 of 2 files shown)


perexg:
I applied only the switch patch as cf267cce6d553d4d436c166c23d3befc5476acef . The volume applies to all hardware and it's difficult to find a good common value. If you like to change the value for the specific model, please, correct the configuration.

PR #60, opened 2020-11-03
ucm2:sof-soundwire: Add Support For Dell Privacy Mic Mute LED Control

GoPerry:
Dell is going to add internal microphone mute led control for Dell privacy system,
which has hardware level protection for audio privacy, when mic mute
state is changed, the ucm2 will help to switch the Mic Mute LED state.

Signed-off-by: perry_yuan <perry_yuan@dell.com>

### DIFF
--- a/ucm2/sof-soundwire/rt715.conf
+++ b/ucm2/sof-soundwire/rt715.conf
@@ -5,10 +5,12 @@ SectionDevice."Mic" {
 
 	EnableSequence [
 		cset "name='PGA5.0 5 Master Capture Switch' 1"
+		cset "name='rt715 Micmute Led Mode' On"
 	]
 
 	DisableSequence [
 		cset "name='PGA5.0 5 Master Capture Switch' 0"
+		cset "name='rt715 Micmute Led Mode' Off"
 	]
 
 	Value {


ucm2/sof-soundwire/rt715.conf
```diff
@@ -5,10 +5,12 @@ SectionDevice."Mic" {
 
 	EnableSequence [
 		cset "name='PGA5.0 5 Master Capture Switch' 1"
+        cset "name='rt715 Micmute Led Mode' On"
```
superm1, line 8: it looks like this is using spaces, but the file normally uses tabs to me.

superm1:
@GoPerry based upon LKML feedback, I think it's appropriate to close this PR and switch around the kernel driver to flip based upon the capture switch.  However you'll need to get fixed UCM from https://github.com/thesofproject/linux/issues/2544 on your SUT.

ucm2/sof-soundwire/rt715.conf
GoPerry: I will update the patch, check later.

GoPerry:
I am going to close this PR for leveraging the new "Main Capture Switch"

PR #63, opened 2020-11-20
ucm2: add support to for Qualcomm RB5 Platform

Srinivas-Kandagatla:
This patchset adds support to The Qualcomm Robotics RB5 Platform Audio ports.
RB5 has 2 WSA881X smart speakers attached via Soundwire, HDMI and a DMIC0 on the board.
All the audio related patches are merged in ASoC tree.

Also to note that the driver name and long card names are fixed properly on all new and old machine drivers so that we can organise the ucm2 directories correctly. That is the reason for moving sm8250 out of Qualcomm folder.

Thanks,
srini


Srinivas-Kandagatla:
hi @perexg , 
Do you have any comments on this patch?
The validation is failing as its unable to find the alsainfo file for this board which is also available in alsa-test pull request.

Many thanks,
srini

PR #70, opened 2020-12-18
Move SDM845 out of Qualcomm folder and add C360

Srinivas-Kandagatla:
With driver_name fix in sdm845 machine driver  https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?h=v5.10&id=3f48b6eba15ea342ef4cb420b580f5ed6605669f 
/proc/asound/cards get populated correctly with proper driver name and card long name. With this in place we should be able to rearrange sdm845 directory to be inline with others and reuse some of the common code.
Without this all the sdm845 based machines will get a generic ucm description which will not work, one of such instance is Lenovo C360 Laptop. This patchset also add support to Lenovo C360 support

Thanks,
srini


perexg:
The Qualcomm directory was added to make the configuration directory tree more descriptive. The string _sdm845_ string is not very descriptive, but _Qualcomm/sdm845_ path looks much better in my eyes.

I am playing with an idea to make the lookups like udev does for block devices (by-id / by-path / by-uuid). So the driver / name redirection may be done via a symlink directory like for module names (ucm2/modules). We can change the lookups via ucm2/ucm.conf file now.

It's not only against the Qualcomm drivers. I don't like the old Intel names (_chtnau8824_ for example) which may be moved to _Intel/_ subtree.

It's on my to do list, I'll try to do this change when my time permits. If you keep the Qualcomm/sdm845 tree, I'll accept your change now (assuming that the module name lookup works for you).

Srinivas-Kandagatla:
> The Qualcomm directory was added to make the configuration directory tree more descriptive. The string _sdm845_ string is not very descriptive, but _Qualcomm/sdm845_ path looks much better in my eyes.
> 
> I am playing with an idea to make the lookups like udev does for block devices (by-id / by-path / by-uuid). So the driver / name redirection may be done via a symlink directory like for module names (ucm2/modules). We can change the lookups via ucm2/ucm.conf file now.
> 
> It's not only against the Qualcomm drivers. I don't like the old Intel names (_chtnau8824_ for example) which may be moved to _Intel/_ subtree.
> 
> It's on my to do list, I'll try to do this change when my time permits. If you keep the Qualcomm/sdm845 tree, I'll accept your change now (assuming that the module name lookup works for you).
Yes, sdm845 as it is, is not very descriptive.. like any other  SoC names, namespacing it with Vendor name sounds good to me!  Am Okay with it, as long as we do not endup with using single ucm conf (specially for Qcom case) for multiple machines.

lumag:
@perexg unfortunately module lookups won't work if the driver is built in into the kernel. Or if the same driver covers multiple cards (like this PR).
I'd suggest to have additional in the ucm.conf, like `ucm2/module/${KernelModule}/${CardLongName}.conf` (instead/in addition to current `ucm2/module/${KernelModule}.conf`)?

```diff
--- alsa-ucm-conf-1.2.4.orig/ucm2/ucm.conf
+++ alsa-ucm-conf-1.2.4/ucm2/ucm.conf
@@ -53,6 +53,10 @@ If.driver {
 			False {
 				Define.KernelModulePath "class/sound/card${CardNumber}/device/driver/module"
 				Define.KernelModule "$${sys:$KernelModulePath}"
+				UseCasePath.modulelongname {
+					Directory "module/${var:KernelModule}"
+					File "${CardLongName}.conf"
+				}
 				UseCasePath.module {
 					Directory "module"
 					File "${var:KernelModule}.conf"
```

As for the monolithic kernel builds, would it be possible to have include e.g. `ucm2/driver/${CardDriver}/${CardLongName}.conf` and `ucm2/driver/${CardDriver}.conf` into UseCase path? Doing so would allow one to use driver name to locate the directory in addition to using modules.

perexg:
Obsoleted by 90ee4e03095ec6917a9711782282ba6c4205364c .

PR #67, opened 2020-12-15
ucm2: sof-soundwire - add support for new main capture switch and main capture volume control

GoPerry:
This patch was added for its new main capture switch and main capture
volume control which combine the ADC 07 / ADC 27 Capture and Volume
switch ,the new mute switch and volume control will be handled by codec
rt715`s driver mute/volume kcontrol instead of PGA5.

BugLink: https://github.com/thesofproject/linux/issues/2544
Signed-off-by: Perry Yuan <perry_yuan@dell.com>

PR #71, opened 2020-12-23
Sdca support

libinyang:
Add supports for SDW sdca codec devices

### DIFF
--- a/ucm2/codecs/rt715-sdca/init.conf
+++ b/ucm2/codecs/rt715-sdca/init.conf
@@ -1,0 +1,9 @@
++# RT715-sdca (aka RT714) specific volume control settings
+
+BootSequence [
+	cset "name='rt714 FU0A Capture Switch' 0"
+	cset "name='rt714 ADC 22 Mux' 'DMIC3'"
+	cset "name='rt714 ADC 23 Mux' 'DMIC4'"
+	cset "name='rt714 FU02 Capture Switch' 1"
+	cset "name='rt714 FU02 Capture Volume' 124"
+]
--- a/ucm2/sof-soundwire/rt1316-1.conf
+++ b/ucm2/sof-soundwire/rt1316-1.conf
@@ -1,0 +1,22 @@
+# Use case Configuration for sof-soundwire card
+
+SectionDevice."Speaker" {
+	Comment	"Speaker"
+
+	EnableSequence [
+		cset "name='rt1316-1 DAC L Switch' 1"
+		cset "name='rt1316-1 DAC R Switch' 1"
+		cset "name='Speaker Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='rt1316-1 DAC L Switch' 0"
+		cset "name='rt1316-1 DAC R Switch' 0"
+		cset "name='Speaker Switch' off"
+	]
+
+	Value {
+	      PlaybackPriority 100
+	      PlaybackPCM "hw:${CardId},2"
+	}
+}
--- a/ucm2/sof-soundwire/rt1316-2.conf
+++ b/ucm2/sof-soundwire/rt1316-2.conf
@@ -1,0 +1,29 @@
+# Use case Configuration for sof-soundwire card
+
+SectionDevice."Speaker" {
+	Comment	"Speaker"
+
+	EnableSequence [
+		cset "name='rt1316-1 RX Channel Select' L,L"
+		cset "name='rt1316-2 RX Channel Select' R,R"
+
+		cset "name='rt1316-1 DAC L Switch' 1"
+		cset "name='rt1316-1 DAC R Switch' 1"
+		cset "name='rt1316-2 DAC L Switch' 1"
+		cset "name='rt1316-2 DAC R Switch' 1"
+		cset "name='Speaker Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='rt1316-1 DAC L Switch' 0"
+		cset "name='rt1316-1 DAC R Switch' 0"
+		cset "name='rt1316-2 DAC L Switch' 0"
+		cset "name='rt1316-2 DAC R Switch' 0"
+		cset "name='Speaker Switch' off"
+	]
+
+	Value {
+	      PlaybackPriority 100
+	      PlaybackPCM "hw:${CardId},2"
+	}
+}
--- a/ucm2/sof-soundwire/rt711-sdca.conf
+++ b/ucm2/sof-soundwire/rt711-sdca.conf
@@ -1,0 +1,40 @@
+# Use case Configuration for sof-soundwire card
+
+SectionDevice."Headphones" {
+	Comment	"Headphones"
+
+	EnableSequence [
+		cset "name='Headphone Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='Headphone Switch' off"
+	]
+
+	Value {
+	      PlaybackPriority 200
+	      PlaybackPCM "hw:${CardId}"
+	      JackControl "Headphone Jack"
+	}
+}
+
+SectionDevice."Headset" {
+	Comment "Headset Microphone"
+
+	EnableSequence [
+		cset "name='PGA2.0 2 Master Capture Switch' 1"
+	]
+
+	DisableSequence [
+		cset "name='PGA2.0 2 Master Capture Switch' 0"
+	]
+
+	Value {
+		CapturePriority 200
+		CapturePCM "hw:${CardId},1"
+		JackControl "Headset Mic Jack"
+		CaptureSwitch "PGA2.0 2 Master Capture Switch"
+		CaptureVolume "PGA2.0 2 Master Capture Volume"
+		CaptureMixerElem "PGA2.0 2 Master"
+	}
+}
--- a/ucm2/sof-soundwire/rt715-sdca.conf
+++ b/ucm2/sof-soundwire/rt715-sdca.conf
@@ -1,0 +1,21 @@
+# Use case Configuration for sof-soundwire card
+
+SectionDevice."Mic" {
+	Comment	"SoundWire microphones"
+
+	EnableSequence [
+		cset "name='rt714 FU02 Capture Switch' 1"
+	]
+
+	DisableSequence [
+		cset "name='rt714 FU02 Capture Switch' 0"
+	]
+
+	Value {
+	      CapturePriority 100
+	      CapturePCM "hw:${CardId},4"
+	      CaptureSwitch "rt714 FU02 Capture Switch"
+	      CaptureVolume "rt714 FU02 Capture Volume"
+	      CaptureMixerElem "rt714 FU02"
+	}
+}
--- a/ucm2/codecs/rt711-sdca/init.conf
+++ b/ucm2/codecs/rt711-sdca/init.conf
@@ -1,0 +1,8 @@
+# RT711-sdca specific volume control settings
+
+BootSequence [
+	cset "name='rt711 FU05 Playback Volume' 87"
+	cset "name='rt711 ADC 22 Mux' 'MIC2'"
+	cset "name='rt711 FU0F Capture Volume' 57"
+	cset "name='rt711 FU0F Capture Switch' 1"
+]
--- a/ucm2/sof-soundwire/HiFi.conf
+++ b/ucm2/sof-soundwire/HiFi.conf
@@ -9,7 +9,7 @@ If.spkdev {
 		Type String
 		Empty "${var:SpeakerCodec1}"
 	}
-	False.Include.spkdev.File "/sof-soundwire/${var:SpeakerCodec1}-${var:SpeakerChannels1}.conf"
+	False.Include.spkdev.File "/sof-soundwire/${var:SpeakerCodec1}-${var:SpeakerAmps1}.conf"
 }
 
 If.micdev {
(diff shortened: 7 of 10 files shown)


libinyang:
@perexg Ping for review. This PR has been tested on sdca codecs.

libinyang:
BTW: we are trying to remove PGA in rt715-sdca. But there is an issue in codec driver. After it is fixed, we will switch to 'rt714 FU02' for PGA

libinyang:
update the patch to use 'rt714 FU02' as this bug is fixed by https://github.com/thesofproject/linux/pull/2670

perexg:
Basically, the patch looks ok. Can we have also output from 'alsa-info.sh --no-upload' for the added sdca devices to improve the validator?

libinyang:
> Basically, the patch looks ok. Can we have also output from `alsa-info.sh --no-upload` for the added sdca devices to improve the validator?

@perexg Do you mean submit the alsa-info.sh output to alsa-tests? I will do it.

I have another question about the UCM for the rt715 sdca codec. It perplexes me. We used PGA control as the volume/mute control before. But currently, we are trying to use rt715's kcontrol in non-sdca codec, such as 'rt715 Main Capture Switch' in https://github.com/alsa-project/alsa-ucm-conf/pull/67. I also change to use "rt714 FU02" to replace the PGA control in this PR. Do you think we should use "rt714 FU02" or PGA kcontrols? 
My concern is that it may impact on the MIC LED. Realtek's codec driver won't touch the MIC LED based on my test. So if we use "rt714 FU02", the MIC LED may not work on the old platforms. 
Or we use "rt714 FU02" in the PR, and if we met some platforms which MIC LED doesn't work, then we can add patches for that platform. What do you think?

GoPerry:
@libinyang 
Will the new SDCA Codec will not use the rt715 sdca ucm2 conf ?
The new Intel platforms and sdca codec will only use the new ucm2 conf, we can only use the new sdca conf replacing the PGA kcontrol,is it right ?

the non sdca codec, for example rt715  ucm conf will need to have the compatible conf for the other platforms which has the none sdca codec.
[
https://github.com/alsa-project/alsa-ucm-conf/pull/67](url)

I have submmited the initial PR for the compatible case.
If @perexg agree to use this matching method, i will test that patch on my system.

perexg:
My opinion is that the analog volume control is preferred than the PCM stream modifications. So PGA controls should not be used, if we have the analog controls.

But the situation for Dell device is a bit complicated, because we have multiple controls for the microphone array (2 stereo controls for four microphones) and we cannot describe this in UCM at the moment. I proposed to avoid the kernel driver change (which is incomplete anyway) and use the current scenario - use PGA for the volume control and set the FU registers to a sane maximal value per machine. We can do DMI checks at the UCM layer. My plan is to add the translation code for two stereo controls -> one four channel controls to alsa-lib in the near future so we can switch to FU then.

The LED problem is driver specific. I believe that all controls (FU, PGA in this case) on the path which effectively turns off capture in the kernel space (zero samples), should invoke LED off. See my proposal on the alsa-devel mailing list: https://lore.kernel.org/alsa-devel/1765b000-ac56-7373-b544-b5d9daa1565b@perex.cz/T/#t .

libinyang:
> My opinion is that the analog volume control is preferred than the PCM stream modifications. So PGA controls should not be used, if we have the analog controls.
> 
> But the situation for Dell device is a bit complicated, because we have multiple controls for the microphone array (2 stereo controls for four microphones) and we cannot describe this in UCM at the moment. I proposed to avoid the kernel driver change (which is incomplete anyway) and use the current scenario - use PGA for the volume control and set the FU registers to a sane maximal value per machine. We can do DMI checks at the UCM layer. My plan is to add the translation code for two stereo controls -> one four channel controls to alsa-lib in the near future so we can switch to FU then.
> 
> The LED problem is driver specific. I believe that all controls (FU, PGA in this case) on the path which effectively turns off capture in the kernel space (zero samples), should invoke LED off. See my proposal on the alsa-devel mailing list: https://lore.kernel.org/alsa-devel/1765b000-ac56-7373-b544-b5d9daa1565b@perex.cz/T/#t .

@perexg Got it. I will use PGA in my updated patches. Thanks.

libinyang:
> @libinyang
> Will the new SDCA Codec will not use the rt715 sdca ucm2 conf ?

@GoPerry 
SDCA rt715 codec will use rt715 sdca UCM.

> The new Intel platforms and sdca codec will only use the new ucm2 conf, we can only use the new sdca conf replacing the PGA kcontrol,is it right ?

sdca codec will use xxx-sdca.conf. 

> 
> the non sdca codec, for example rt715 ucm conf will need to have the compatible conf for the other platforms which has the none sdca rt715 codec.

Yes, this is why we needs separated sdca codec UCM.

> 
> I have submmited the initial PR for the compatible case.
> If @perexg agree to use this matching method, i will test that patch on my system.

libinyang:
Patch is updated and alsa-info is submitted to https://github.com/alsa-project/alsa-tests/pull/15

ucm2/codecs/rt715-sdca/init.conf
```diff
@@ -0,0 +1,9 @@
++# RT715-sdca (aka RT714) specific volume control settings
+
+BootSequence [
+	cset "name='rt714 FU0A Capture Switch' 0"
+	cset "name='rt714 ADC 22 Mux' 'DMIC3'"
+	cset "name='rt714 ADC 23 Mux' 'DMIC4'"
+	cset "name='rt714 FU02 Capture Switch' 1"
+	cset "name='rt714 FU02 Capture Volume' 124 124"
```
perexg, line 8: Perhaps, the initialization for the second stereo pair is missing here? In the kernel patch (https://lore.kernel.org/alsa-devel/7d74544f-77f2-db01-db3d-d1d8a56d576d@linux.intel.com/T/#m9d8869261e4531f537336da6b81085c1567034a9) the four stereo controls are set:

```
+	unsigned int capture_reg_H[] = {RT715_SET_GAIN_MIC_ADC_H,
+		RT715_SET_GAIN_LINE_ADC_H, RT715_SET_GAIN_MIX_ADC_H,
+		RT715_SET_GAIN_MIX_ADC2_H};
+	unsigned int capture_reg_L[] = {RT715_SET_GAIN_MIC_ADC_L,
+		RT715_SET_GAIN_LINE_ADC_L, RT715_SET_GAIN_MIX_ADC_L,
+		RT715_SET_GAIN_MIX_ADC2_L};
```

The stereo controls are:

```
        SOC_DOUBLE_R_EXT_TLV("ADC 07 Capture Volume", RT715_SET_GAIN_MIC_ADC_H,
                        RT715_SET_GAIN_MIC_ADC_L, RT715_DIR_IN_SFT, 0x3f, 0,
                        rt715_set_amp_gain_get, rt715_set_amp_gain_put,
                        in_vol_tlv),
        SOC_DOUBLE_R_EXT_TLV("ADC 08 Capture Volume", RT715_SET_GAIN_LINE_ADC_H,
                        RT715_SET_GAIN_LINE_ADC_L, RT715_DIR_IN_SFT, 0x3f, 0,
                        rt715_set_amp_gain_get, rt715_set_amp_gain_put,
                        in_vol_tlv),
        SOC_DOUBLE_R_EXT_TLV("ADC 09 Capture Volume", RT715_SET_GAIN_MIX_ADC_H,
                        RT715_SET_GAIN_MIX_ADC_L, RT715_DIR_IN_SFT, 0x3f, 0,
                        rt715_set_amp_gain_get, rt715_set_amp_gain_put,
                        in_vol_tlv),
        SOC_DOUBLE_R_EXT_TLV("ADC 27 Capture Volume", RT715_SET_GAIN_MIX_ADC2_H,
                        RT715_SET_GAIN_MIX_ADC2_L, RT715_DIR_IN_SFT, 0x3f, 0,
                        rt715_set_amp_gain_get, rt715_set_amp_gain_put,
                        in_vol_tlv),
```

If you need only one control, we can use the hardware rt714 register controls now. My impression was that there are more microphones then 2 are connected... Perhaps, I need to study, how the FU controls are build for rt714/715 (they come from topology files?).

perexg: Ok, you have rt715-scda.c only in your development repo. It is not upstreamed and the FU02 (and maybe FU0E/FU0C boost) controls are problematic in my eyes - it contains the stereo volume/switch mapping/coupling to multiple hardware registers/inputs.

I created https://github.com/thesofproject/linux/issues/2696 .

perexg:
Until thesofproject/linux#2696 is resolved, we cannot continue. If FU SCDA controls will be multichannel, I would prefer to use them, otherwise see my https://github.com/alsa-project/alsa-ucm-conf/pull/71#issuecomment-761118130 .

libinyang:
> Until [thesofproject/linux#2696](https://github.com/thesofproject/linux/issues/2696) is resolved, we cannot continue. If FU SCDA controls will be multichannel, I would prefer to use them, otherwise see my [#71 (comment)](https://github.com/alsa-project/alsa-ucm-conf/pull/71#issuecomment-761118130) .

@perexg The patch for rt715-sdca is in https://github.com/thesofproject/linux/ already. And I suppose this will be soon upstreamed. And thanks to raise the issue https://github.com/thesofproject/linux/issues/2696. In our current design, FU02 will control multiple channels. I would like to pending this PR and let's talk more details in  https://github.com/thesofproject/linux/issues/2696.

libinyang:
Patch is updated based on @jack-cy-yu 's latest rt715 codec patch https://github.com/jack-cy-yu/linux/commit/17706c534bf49e3208927c106f48dc3c0d864846

libinyang:
@perexg This patch is based on @jack-cy-yu 's latest sdca codec patch. Could you please help review? Thanks.

ucm2/sof-soundwire/rt1316-1.conf
```diff
@@ -0,0 +1,22 @@
...
+	DisableSequence [
+		cset "name='rt1316-1 DAC L Switch' 0"
+		cset "name='rt1316-1 DAC R Switch' 0"
+		cset "name='Speaker Switch' off"
+	]
+
+	Value {
+	      PlaybackPriority 100
+	      PlaybackPCM "hw:${CardId},2"
+	}
```
perexg, line 21: No hw volume control?

ucm2/sof-soundwire/rt1316-2.conf
```diff
@@ -0,0 +1,29 @@
...
+		cset "name='rt1316-1 DAC L Switch' 0"
+		cset "name='rt1316-1 DAC R Switch' 0"
+		cset "name='rt1316-2 DAC L Switch' 0"
+		cset "name='rt1316-2 DAC R Switch' 0"
+		cset "name='Speaker Switch' off"
+	]
+
+	Value {
+	      PlaybackPriority 100
+	      PlaybackPCM "hw:${CardId},2"
```
perexg, line 27: No hw volume control?

ucm2/sof-soundwire/rt711-sdca.conf
```diff
@@ -0,0 +1,39 @@
...
+	DisableSequence [
+		cset "name='PGA2.0 2 Master Capture Switch' 0"
+	]
+
+	Value {
+		CapturePriority 200
+		CapturePCM "hw:${CardId},1"
+		JackControl "Headset Mic Jack"
+		CaptureSwitch "PGA2.0 2 Master Capture Switch"
+		CaptureVolume "PGA2.0 2 Master Capture Volume"
```
perexg, line 37: Missing MixerElem for PA and no hw volume control?

ucm2/sof-soundwire/rt715-sdca.conf
```diff
@@ -0,0 +1,20 @@
...
+
+	DisableSequence [
+		cset "name='rt714 FU02 Capture Switch' 0"
+	]
+
+	Value {
+	      CapturePriority 100
+	      CapturePCM "hw:${CardId},4"
+	      CaptureSwitch "rt714 FU02 Capture Switch"
+	      CaptureVolume "rt714 FU02 Capture Volume"
```
perexg, line 18: Missing MixerElem for PA.

perexg (changes requested):

ucm2/sof-soundwire/rt715-sdca.conf
libinyang: For the MixerElem, is `CaptureMixerElem "rt714 FU02"` good enough?
I checked how to set control for MixerElem in alsa-lib. It seems very little information about it. I don't know what is the best solution to set the MixerElem. Could you give a little hints?

ucm2/sof-soundwire/rt711-sdca.conf
libinyang: I didn't find a proper switch/volume kcontrol for it on my platform.

libinyang:
@perexg Patches are updated based on my understanding. Please help review whether my update is OK or not. I'm not so confident about the MixerElem setting is ok or not. For the rt711-sdca hw control, "rt711 FU0F" should be the right control. But the switch doesn't work on my platform. So I still use the PGA kcontrol in this version.

libinyang:
ping @perexg

KaiChuan-Hsieh:
The commit test passed on machine with sdca codec.

ucm2/sof-soundwire/rt1316-1.conf
perexg: I'd like double check: The codec does not seem any volume control, but do we have any DSP (PGA) controls which can modify the stream volume to use them here?

perexg:
I applied this to ucm repo, but I still think that the hw controls should be improved for the best experience.

PR #73, opened 2021-01-14
USB-Audio/ALC1220: Bump analog Speaker priority over S/PDIF

MarijnS95:
Suggestion from https://github.com/alsa-project/alsa-ucm-conf/pull/25#issuecomment-760141633

The `S/PDIF` port does not have any jack sensing, and with a priority higher than the `Speakers` it will always be selected by default instead of the `Speakers` even if unplugged.

Swapping the priorities around allows analog `Speakers` to be selected first, _if_ they are plugged in. Otherwise `S/PDIF` is used.

The sections are reordered to adhere to this priority, including some indentation fixes.

### DIFF
--- a/ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop-HiFi.conf
+++ b/ucm2/USB-Audio/Realtek-ALC1220-VB-Desktop-HiFi.conf
@@ -1,14 +1,3 @@
-SectionDevice."Speaker" {
-	Comment "Speakers"
-	Value {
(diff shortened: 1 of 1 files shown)


perexg:
Pushed via 062e4710c524283c5d11cbf99931b24c6d963f98 (excluding the device shuffling - it's not required).

MarijnS95:
@perexg Would've been nice if you fixed the indentation of the lines touched here, though :no_mouth:

PR #74, opened 2021-01-18
chtnau8824: Add support for using the SOF driver

perexg:
The old (and currently the default) SST driver uses TDM 4 slots 24 bit
as wire format to the codec. Where as the new SOF driver uses standard
I2S 2 channel 24 bit.

Normally this should not impact the UCM settings, but on the NAU8824
the "DAC Right Channel Source" mixer setting must be set to 1 when
using TDM 4 slots and to 0 when using I2S 2 channel mode.

Getting this wrong (in either case) results in the right channel not
outputting any sound.

This commit introduces a RightOutputChannel variable which gets
set to 0/1 depending on the driver and then uses that for the
"DAC Right Channel Source" mixer setting so that we do the right
thing depending on the driver.

This has been tested on the following devices:

Medion E2215T:    Stereo speakers, analog mic
Medion E2228T:    Stereo speakers, stereo digital mics
Cube iWork 8 Air: Mono speaker, analog mic

### DIFF
--- a/ucm2/chtnau8824/HiFi.conf
+++ b/ucm2/chtnau8824/HiFi.conf
@@ -1,6 +1,7 @@
 Define.Speaker "Speaker"
 Define.Mic "InternalMic"
 Define.SST "yes"
+Define.RightOutputChannel 1
 
 If.Controls {
 	Condition {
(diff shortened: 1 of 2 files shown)


jwrdegoede:
I can confirm that this simplified version works as it should, thanks.

perexg:
Merged in 96e1c19a6bcc1e353f677966d4d93399204af0ec .

PR #75, opened 2021-01-21
SOF SoundWire: Improve dmic detection

plbossart:
DMIC support seems incomplete for HDaudio (not handling all possibilities for cfg-dmics) and missing completely for SoundWire.

This is an untested draft just to get UCM3 feedback from @perexg and tests from @aigilea on an HP Spectre x360 device.

Edit: adding @kv2019i @RanderWang @bardliao and @libinyang for reviews.

### DIFF
--- a/ucm2/sof-soundwire/dmic.conf
+++ b/ucm2/sof-soundwire/dmic.conf
@@ -1,0 +1,21 @@
+SectionDevice."Mic" {
+	Comment "Digital Microphone"
+
+	Value {
+		CapturePriority 100
+		CapturePCM "hw:${CardId},3"
+		If.chn {
+			Condition {
+				Type RegexMatch
+				Regex "[34]"
+				String "${var:Mics1}"
+			}
+			True {
+				CaptureChannels 4
+			}
+		}
+		CaptureMixerElem "Dmic0"
+		CaptureVolume "Dmic0 Capture Volume"
+		CaptureSwitch "Dmic0 Capture Switch"
+	}
+}
--- a/ucm2/sof-soundwire/sof-soundwire.conf
+++ b/ucm2/sof-soundwire/sof-soundwire.conf
@@ -11,6 +11,7 @@ Define {
 	SpeakerAmps1 "0"
 	HeadsetCodec1 ""
 	MicCodec1 ""
+	Mics1 "0"
 }
 
 DefineRegex {
@@ -32,6 +33,10 @@ DefineRegex {
 	}
 	MicCodec {
 		Regex " mic:([a-z0-9]+)"
+		String "${CardComponents}"
+	}
+	Mics {
+		Regex " cfg-mics:([1-9][0-9]*)"
 		String "${CardComponents}"
 	}
 }


aigilea:
@plbossart `If.1mic` block is not closed in `sof-hda-dsp/HiFi.conf`.
I've applied the patch to files in `/usr/share/alsa/ucm2/` and rebooted, but can't see any changes (no real pcms in `arecord -L`, no sources in pulseaudio).

plbossart:
Thanks @aigilea, I don't know what happened here, appreciate the checks. I pushed something that passes the test, except for the following reports (missing configs)?
````
./ucm.py --level 0 all ../../../alsa-ucm-conf/ucm2
./ucm.py --level 0 configs ../../../alsa-ucm-conf/ucm2 configs 
ERR: sof-hda-dsp/sof-hda-dsp.conf@HiFi.conf: 'SectionDevice'.'Mic1'.'Value'.'If'.'3mic'.'Condition' - True block not executed
ERR: sof-soundwire/sof-soundwire.conf@HiFi.conf: 'If'.'dmicdev'.'False'.'Include'.'dmicdev'.'SectionDevice'.'Mic1'.'Value'.'If'.'3mic'.'Condition' - True block not executed
ERR: sof-soundwire/sof-soundwire.conf@HiFi.conf: 'If'.'dmicdev'.'False'.'Include'.'dmicdev'.'SectionDevice'.'Mic1'.'Value'.'If'.'4mic'.'Condition' - True block not executed
ERR: sof-soundwire/sof-soundwire.conf@HiFi.conf: 'If'.'dmicdev'.'False'.'Include'.'dmicdev'.'SectionDevice'.'Mic1'.'Value'.'If'.'vol'.'Condition' - True block not executed
ERR: total errors: 4
make: *** [Makefile:19: configs] Error 1

````

File: ucm2/sof-hda-dsp/HiFi.conf [hunk 1]
```diff
@@ -14,7 +14,36 @@ SectionDevice."Mic1" {
...
-		If.chn {
+		If.1mic {
+			Condition {
+			Type String
+			Haystack "${CardComponents}"
+			Needle "cfg-dmics:1"
+		}
+		True {
+			CaptureChannels 2
+		}
```
libinyang, line 25: Here missed a '}'.

ucm2/sof-hda-dsp/HiFi.conf
```diff
@@ -14,7 +14,36 @@ SectionDevice."Mic1" {
 	Value {
 		CapturePriority 100
 		CapturePCM "hw:${CardId},6"
-		If.chn {
+		If.1mic {
+			Condition {
+			Type String
+			Haystack "${CardComponents}"
+			Needle "cfg-dmics:1"
+		}
```
libinyang, line 22: The alignment is wrong

ucm2/sof-soundwire/HiFi.conf
```diff
@@ -28,4 +28,13 @@ If.hsdev {
 	False.Include.hsdev.File "/sof-soundwire/${var:HeadsetCodec1}.conf"
 }
 
+If.dmicdev {
```
perexg, line 31: This should be merged to `If.micdev`. I assume that there will be never mix of SDW and direct DMIC.

ucm2/sof-soundwire/sof-soundwire.conf
```diff
@@ -34,6 +35,11 @@ DefineRegex {
 		Regex " mic:([a-z0-9]+)"
 		String "${CardComponents}"
 	}
+	# FIXME: the case cfg-dmics=0 should fail.
+	Dmics {
+		Regex " cfg-dmics:([0-9]+)"
```
perexg, line 40: Better regex: `Regex " cfg-dmics:([1-9][0-9]*)"` . I think that you can remove the FIXME with it. BTW: Nice regex online tool: https://regex101.com/

ucm2/sof-soundwire/dmic.conf
```diff
@@ -0,0 +1,46 @@
+SectionDevice."Mic1" {
+	Comment "Digital Microphone"
+
+	Value {
+		# FIXME: value is not 100 to avoid duplicate priority reported
+		# by ucm-validator		
+		CapturePriority 101
+		CapturePCM "hw:${CardId},3"
+		If.3mic {
```
perexg, line 9: Use `Regex` condition here with `cfg-dmics:[34]` string here. It simplifies the config.

ucm2/sof-hda-dsp/HiFi.conf
```diff
@@ -14,7 +14,17 @@ SectionDevice."Mic1" {
...
+			Condition {
+				Type String
+				Haystack "${CardComponents}"
+				Needle "cfg-dmics:3"
+			}
+			True {
+				CaptureChannels 4
+			}
+		}
+		If.4mic {
```
perexg, line 27: Use Regex condition here with cfg-dmics:[34] string here. It simplifies the config (keep one If).

perexg (changes requested):

ucm2/sof-soundwire/sof-soundwire.conf
plbossart: ack, done

ucm2/sof-hda-dsp/HiFi.conf
plbossart: ack, done.

ucm2/sof-soundwire/dmic.conf
plbossart: ack, done.

ucm2/sof-soundwire/HiFi.conf
plbossart: It's likely indeed, but I don't get what you meant by merging. We use

rt715-> mic:rt715 and if found we use a dedicated init.conf
dmic -> cfg-dmics:1..4

perexg: I assume that having both 'mic:XXX and cfg-dmic:1+' is invalid, right? So add the second If inside If.micdev something like:

<pre>
If.micdev {
        Condition {
                Type String
                Empty "${var:MicCodec1}"
        }
        False.Include.micdev.File "/sof-soundwire/${var:MicCodec1}.conf"
        True.If {
                Condition {
                        Type String
                        Empty "${var:Dmics1}"
                }
                False.Include.dmic.File "/sof-soundwire/dmic.conf"
        }
}
</pre>

You're true. It's not direct merging.

ucm2/sof-hda-dsp/HiFi.conf
```diff
@@ -18,7 +18,7 @@ SectionDevice."Mic1" {
 			Condition {
 				Type String
```
perexg, line 19: This must be changed to RegexMatch and Haystack should be renamed to String. See `If.hs_init` for the RegexMatch condition example.

ucm2/sof-soundwire/sof-soundwire.conf
```diff
@@ -11,6 +11,7 @@ Define {
 	SpeakerAmps1 "0"
 	HeadsetCodec1 ""
 	MicCodec1 ""
+	Dmics1 "0"
```
perexg, line 14: Apparently, this string should be empty by default (Empty condition for include).

plbossart: the machine driver will set cfg-dmics=0, are you saying we should only conditionally set the component string when we have a non-zero value?

plbossart: we could also write 'mic:dmic' if that was desired, so that there's a built-in mutual exclusion in the driver.

ucm2/sof-soundwire/HiFi.conf
plbossart: it depends what we want. If mic:xyz is the name of the input, we could have mic:rt715 or mic:dmic.
cfg-dmics was added specifically for dmic but maybe it should be cfg-mic in case we want to deal with other configurations (e.g. for cfg-spk we don't care what type of amp is used).

perexg: My configuration works with both `mic:dmic` and `cfg-dmic:2` option without `mic:dmic`. If both are passed by the driver, then the second if is not required (dmic will be extracted from `mic:dmic`, right?).

But with `mic:rt715 cfg-dmics:4` we are in trouble, but I assume that it's invalid hw combo (direct dmics + SDW dmics), right?

perexg:
I applied the "sof-hda-dsp: update handling of cfg-dmics strings" as 3bea51c3188257c5bc0eaaed922d1f714e3508c7 .

My modified version of the sof-soundwire config is here: https://github.com/perexg/alsa-ucm-conf/commit/5490428d7fe741bcfe17d3dd138f15e3d1359d60

I assume that `mic:dmic` exists in the components string.

I would also remove the support for v1.3 firmware for SDW (If.vol in sof-soundwire/dmic.conf). I believe that nobody uses this version for SDW devices?

plbossart:
@perexg I would need to modify the machine driver (which is fine, it's not released yet).

So that we've clear, if dmics are detected you want both 'mic:dmic cfg-dmics:[1234]'', yes?

perexg:
> @perexg I would need to modify the machine driver (which is fine, it's not released yet).
> 
> So that we've clear, if dmics are detected you want both 'mic:dmic cfg-dmics:[1234]'', yes?

Yes, it looks as the best alternative. One is for the internal microphone type, the second is for the microphone channels (perhaps `mic:dmic cfg-mics:[1234]` may be better - we may need to know even the used microphone channels for SDW codecs, too).

plbossart:
> * use `cfg-mics` instead `cfg-dmics`

we used cfg-dmics for HDaudio platforms, and that's the same hardware. It'd be odd to have different definitions for the same thing.

perexg:
> > * use `cfg-mics` instead `cfg-dmics`
> 
> we used cfg-dmics for HDaudio platforms, and that's the same hardware. It'd be odd to have different definitions for the same thing.

The difference is that cfg-dmics on HDA platforms identifies the internal microphone type and channels count. For sof-soundwire, we need to carry only internal microphone channels and it make sense to share `cfg-mics:` identifier with the SDW microphone codecs (on the future demand).

plbossart:
@perexg https://github.com/thesofproject/linux/pull/2701/commits/c3c75fc65a1ea1ec3c1bdff5ac1728f9054d077b is what you want, yes?

perexg:
> @perexg [thesofproject/linux@c3c75fc](https://github.com/thesofproject/linux/commit/c3c75fc65a1ea1ec3c1bdff5ac1728f9054d077b) is what you want, yes?

Yep, it looks fine.

ucm2/sof-soundwire/dmic.conf
```diff
@@ -0,0 +1,36 @@
+SectionDevice."Mic1" {
+	Comment "Digital Microphone"
+
+	Value {
+		# FIXME: value is not 100 to avoid duplicate priority reported
+		# by ucm-validator		
+		CapturePriority 101
+		CapturePCM "hw:${CardId},3"
```
libinyang, line 8: @plbossart I checked our tplgs of sdw with DMIC and found the PCM is not aligned.
1. on sof-cml-rt700 and sof-icl-rt700, it will use PCM2 for DMIC01 and PCM3 for DMIC16k.
2. on sof-tgl-rt5682 and sof-tgl-rt711-i2s-rt1308, it will use PCM3 for DMIC01 and PCM4 for DMIC16k. (The PIPELINE_PCM_ADD and PCM_PLAYBACK_ADD is not aligned in sof-tgl-rt5682)
3. on sof-tgl-sdw-max98373-rt5682, it will use PCM4 for DMIC01 and  PCM5 for DMIC16k (used for keyword detection)
So it seems the dmic PCM is not aligned. It is better to be aligned. What do you think? Or should we not touch the old platforms?

Also I checked rt715 codec DMIC, it will always use PCM4.

plbossart: I don't know how we can align @libinyang 
1. cml-rt700 does not have an amplifier
2. sof-tgl-rt5682 and sof-tgl-rt711-i2s-rt1308 have an external amplifier
3.sof-tgl-sdw-max98373-rt5682 has SoundWire.

The real problem is actually the device numbers, and in a number of cases we used 99 as a temporary solution (which has remained for 2 years). We also don't have consistency on the definition of pipelines, which means e.g. that the tgl-rt711 does not have the right controls for DMIC.

aigilea:
`alsaucm` gives me the following error with the current config:
```
ALSA lib main.c:1014:(snd_use_case_mgr_open) error: failed to import sof-soundwire use case configuration -17
alsaucm: error failed to open sound card sof-soundwire: File exists
```
It completes with no error if `If.dmicdev` block is removed from `HiFi.conf`.
Without that block I can see pcms in `arecord -L` and even record from them. The resulting file is very quiet (I've set maximum gain for dmics in alsamixer), but it works!

`dmic.conf` is still used somehow, even without that block (which is the only include for `dmic.conf` from other config files) as removing it causes following error:
```
ALSA lib utils.c:335:(uc_mgr_config_load) could not open configuration file /usr/share/alsa/ucm2//sof-soundwire/dmic.conf
ALSA lib parser.c:95:(uc_mgr_config_load_file) error: failed to open file /usr/share/alsa/ucm2//sof-soundwire/dmic.conf : -2
ALSA lib main.c:1014:(snd_use_case_mgr_open) error: failed to import sof-soundwire use case configuration -2
alsaucm: error failed to open sound card sof-soundwire: No such file or directory
```

plbossart:
@aigilea Thanks for testing, this looks encouraging.

I have no idea what the 'File exists' error message is telling us. @perexg we could use your help on this one.

Can you push your changes or just attach a zip file of your ucm2/sof-soundwire configuration. I don't know which solution you are using, there's been multiple changes between me and @perexg 

For the volume issue, I kind of expected it. We modified the topology for the DMIC case on HDaudio devices to improve the volume level, but the same fix was not applied in the SoundWire topology, and it doesn't have the same controls. It's no big deal but one step at a time. We'll make sure this device is fully supported.

aigilea:
@plbossart [Here](https://github.com/alsa-project/alsa-ucm-conf/files/5880398/sof-soundwire.zip)'s what I'm using now.

plbossart:
@aigilea indeed you were using an old version with a part that looked suspicious, can you try with the update of this PR (commit from @perexg)

if you don't mind uploading the alsa-info results with the last driver, that would help for UCM validation (device numbers and existence of controls). Thanks!

aigilea:
@plbossart I've updated to perexg's config, no new problems. `alsaucm` completes with no error, pulseaudio [refuses](https://github.com/alsa-project/alsa-ucm-conf/files/5881101/pulseaudio5.txt) to work with the card.

`alsa-info` result:
http://alsa-project.org/db/?f=4c6f69dc528794fb99900138f468d04c6fec2014

plbossart:
Thanks @aigilea. No idea what happens on this one

````
I: [pulseaudio] module-alsa-card.c: Found UCM profiles
I: [pulseaudio] alsa-ucm.c: Set ucm verb to HiFi
I: [pulseaudio] alsa-util.c: Device hw:sofsoundwire,7 doesn't support 44100 Hz, changed to 48000 Hz.
I: [pulseaudio] alsa-util.c: Device hw:sofsoundwire,6 doesn't support 44100 Hz, changed to 48000 Hz.
I: [pulseaudio] alsa-util.c: Device hw:sofsoundwire,5 doesn't support 44100 Hz, changed to 48000 Hz.
I: [pulseaudio] alsa-util.c: Device hw:sofsoundwire doesn't support 44100 Hz, changed to 48000 Hz.
I: [pulseaudio] alsa-util.c: Device hw:sofsoundwire,2 doesn't support 44100 Hz, changed to 48000 Hz.
I: [pulseaudio] alsa-util.c: Device hw:sofsoundwire,1 doesn't support 44100 Hz, changed to 48000 Hz.
I: [pulseaudio] (alsa-lib)conf.c: Unknown parameter 1
I: [pulseaudio] (alsa-lib)conf.c: Parse arguments error: No such file or directory
I: [pulseaudio] (alsa-lib)pcm.c: Unknown PCM plug:hw:sofsoundwire,3
I: [pulseaudio] alsa-util.c: Error opening PCM device plug:hw:sofsoundwire,3: No such file or directory
E: [pulseaudio] module-alsa-card.c: Failed to find a working profile.
````

Rather cryptic what this parameter is.

**Comment by perexg** [comment]
@aigilea : Could you show the debug output from PA? (-vvv argument)?

**Comment by aigilea** [comment]
@perexg [here](https://github.com/alsa-project/alsa-ucm-conf/files/5882038/pulseaudio6.txt), but boils down to the same "Unknown parameter 1"

**Comment by perexg** [comment]
The problem is here:

<pre>
D: [pulseaudio] alsa-ucm.c: Got CapturePriority for device Mic1: 101

D: [pulseaudio] alsa-util.c: Trying hw:sofsoundwire,3 with SND_PCM_NO_AUTO_FORMAT ...
D: [pulseaudio] alsa-util.c: Managed to open hw:sofsoundwire,3
D: [pulseaudio] alsa-util.c: snd_pcm_hw_params_set_channels(2) failed: Invalid argument
</pre>

The channel count is invalid and the capture priority comes from old configuration from @plbossart . You're using a wrong sof-soundwire/dmic.conf file. Use the last alsa-ucm-conf (from this repo) plus this patch on top: https://github.com/alsa-project/alsa-ucm-conf/commit/d0ef97c77c77b687f364b208cade88c9369b278d.patch

**Comment by aigilea** [comment]
@perexg sorry, wrong log
[pulseaudio7.txt](https://github.com/alsa-project/alsa-ucm-conf/files/5882153/pulseaudio7.txt)
no difference

**Comment by perexg** [comment]
<pre>
Components	: 'HDA:80862812,80860101,00100000 cfg-spk:2 cfg-amp:1 mic:dmic cfg-mics:2 hs:rt711 spk:rt1308'
</pre>

PA asks correctly for 2 DMIC channels, but the driver rejects. @plbossart ?

**Comment by aigilea** [comment]
Wow that was insanely fast @perexg . But at least I had fun finding this out as well :)

`Unknown parameter 1` happens when it tries to open `plug:hw:sofsoundwire,3` pcm because `pcm.plug` key in `alsa.conf` expects one string as a parameter and there's extra int here so this name is malformed.
This bogus `plug:` comes from `pa_alsa_open_by_device_string` when it tries to reopen pcm with extra `plug:` after it failed to open it normally.
It fails to open `hw:sofsoundwire,3` normally because `snd_pcm_hw_params_set_channels` fails.
Patching it to never fail allows pulseaudio to start, but it fails later anyway when it receives a frame of incorrect size which is expected.

This happens with arecord as well `arecord -D hw:CARD=sofsoundwire,DEV=4 -d 5 -f S16_LE sample4.wav` with `hw:` devices (and doesn't happen with `plughw:` devices).
```
Recording WAVE 'sample4.wav' : Signed 16 bit Little Endian, Rate 8000 Hz, Mono
arecord: set_params:1349: Channels count non available
```

**Comment by plbossart** [comment]
Thanks @perexg and @aigilea for the information.

That 2ch issue rings a bell, I think we had this problem in the past and that was solved with a topology upgrade.
@kv2019i @juimonen @singalsu, do you recall what we had to do?

**Comment by plbossart** [comment]
Added a topology change to align with HDAudio https://github.com/thesofproject/sof/pull/3790
Completely untested/unproven, I may have the wrong devices or link IDs.

topology file to unzip and install in /lib/firmware/intel/sof-tplg
[sof-tgl-rt711-rt1308-2ch.tplg.gz](https://github.com/alsa-project/alsa-ucm-conf/files/5883673/sof-tgl-rt711-rt1308-2ch.tplg.gz)

**Comment by perexg** [comment]
> `Unknown parameter 1` happens when it tries to open `plug:hw:sofsoundwire,3` pcm because `pcm.plug` key in `alsa.conf` expects one string as a parameter and there's extra int here so this name is malformed.
> This bogus `plug:` comes from `pa_alsa_open_by_device_string` when it tries to reopen pcm with extra `plug:` after it failed to open it normally.

Yes, adding the plug works for the simple device names without additional arguments. The correct plug usage should be:

`plug:SLAVE='hw:sofsoundwire,3'`

The SLAVE argument can pass device name string which can carry arguments.

**Comment by aigilea** [comment]
@plbossart with this file
in [dmesg](https://github.com/alsa-project/alsa-ucm-conf/files/5885757/dmesgj.txt):
```
[    6.392924] sof-audio-pci 0000:00:1f.3: tplg: load pcm DMIC16kHz 4
[    6.392927] sof-audio-pci 0000:00:1f.3: can't find host comp to bind pcm
[    6.392929] sof-audio-pci 0000:00:1f.3: error: can't bind pcm to host
[    6.392931] sof-audio-pci 0000:00:1f.3: ASoC: DAI loading failed
[    6.392931] sof-audio-pci 0000:00:1f.3: ASoC: topology: could not load header: -22
```
`ls /dev/snd`
```
seq  timer
```

File: ucm2/sof-soundwire/dmic.conf (hunk 9 above)
Comment (libinyang) [reply ucm2/sof-soundwire/dmic.conf:6]: > 1. an amplifier

@plbossart All sof-cml-rt700,  sof-icl-rt700, sof-tgl-rt5682, sof-tgl-rt711-i2s-rt1308, sof-tgl-sdw-max98373-rt5682 will use sof-sdw machine driver. So they will use dmic.conf based on this PR. But the DMIC PCM is not aligned. The will cause issues, like https://github.com/alsa-project/alsa-ucm-conf/pull/75#issuecomment-768471801. 

Mostly DMIC01 is used for 4 channels and DMIC16K is used for 2 channels. So the dmic.conf using CapturePCM "hw:${CardId},3" will cause 
1. platforms of sof-cml-rt700 and sof-icl-rt700 will use DMIC16k, which has 2 channels, 
2. platforms sof-tgl-rt5682 and sof-tgl-rt711-i2s-rt1308 will use DMIC01, which has 4 channels.
3. platforms of  sof-tgl-sdw-max98373-rt5682 will failed to find the PCM3, which will cause PA initialization fail.

As PCM num is separated from pipeline num, I suggest we use pcm #99 for DMIC01 (4 channel), pcm #100 for DMIC16k. What do you think?

**Comment by perexg** [comment]
I merged the sdw dmic patch in 889c5b9809f1635005d77871bb9abc09df6a3cc3 to alsa-ucm-conf . It appears that UCM is fine, the topology configuration requires more love..

File: ucm2/sof-soundwire/dmic.conf (hunk 9 above)
Comment (plbossart) [reply ucm2/sof-soundwire/dmic.conf:6]: > @plbossart with this file
> in [dmesg](https://github.com/alsa-project/alsa-ucm-conf/files/5885757/dmesgj.txt):
> 
> ```
> [    6.392924] sof-audio-pci 0000:00:1f.3: tplg: load pcm DMIC16kHz 4
> [    6.392927] sof-audio-pci 0000:00:1f.3: can't find host comp to bind pcm
> [    6.392929] sof-audio-pci 0000:00:1f.3: error: can't bind pcm to host
> [    6.392931] sof-audio-pci 0000:00:1f.3: ASoC: DAI loading failed
> [    6.392931] sof-audio-pci 0000:00:1f.3: ASoC: topology: could not load header: -22
> ```
> 
> `ls /dev/snd`
> 
> ```
> seq  timer
> ```

Yes, I see the same error for the TGL reference platforms which looks very similar to your device.
https://sof-ci.01.org/sofpr/PR3790/build8029/devicetest/
We may have both a problem with device numbers and dailinks.

**Comment by plbossart** [comment]
@aigilea can you apply this patch on top of your tree https://github.com/thesofproject/linux/commit/ff243ba9b960074017b0b1829b04eb95c8edf6d8.patch and enable dynamic debug for sdw_sof?

We create the dailinks dynamically with this machine driver and I don't know what values are expected.

**Comment by aigilea** [comment]
@plbossart done, [dmesg](https://github.com/alsa-project/alsa-ucm-conf/files/5887665/dmesgk.txt):
```
[    6.346348] sof_sdw sof_sdw: create dai link SDW0-Playback, id 0
[    6.346358] sof_sdw sof_sdw: create dai link SDW0-Capture, id 1
[    6.346360] sof_sdw sof_sdw: create dai link SDW1-Playback, id 2
[    6.346361] sof_sdw sof_sdw: create dai link dmic01, id 3
[    6.346362] sof_sdw sof_sdw: create dai link dmic16k, id 4
[    6.346363] sof_sdw sof_sdw: create dai link iDisp1, id 5
[    6.346364] sof_sdw sof_sdw: create dai link iDisp2, id 6
[    6.346366] sof_sdw sof_sdw: create dai link iDisp3, id 7
[    6.346367] sof_sdw sof_sdw: create dai link iDisp4, id 8
```

plbossart:
@aigilea I made a stupid mistake in the topology edit, this should work better 
[sof-tgl-rt711-rt1308-2ch.tplg.gz](https://github.com/alsa-project/alsa-ucm-conf/files/5890375/sof-tgl-rt711-rt1308-2ch.tplg.gz)

plbossart:
@perexg we may have to do another change for the PCM device number. The same machine driver/UCM is used by different devices/topologies, and unfortunately they don't all use the same device number. We may need to align all the SoundWire+DMIC cases to all use the same device number. In the past we used 99 for DMIC and 100 for DMIC16k. This was a value I chose randomly to avoid conflicts some 2+ years ago, and that has trickled in multiple solutions...

I am not aware of any means for the machine driver to figure out what PCM device is used, it only exposes the dailinks/BE and the PCM devices/FE are defined in the topology file.

aigilea:
@plbossart holy s, everything works, including speakers. There's only one sink for microphone and the gain is still bad and pulseaudio lists that microphone and speakers as "availability unknown", but they all do work.

plbossart:
Thanks for testing @aigilea. I don't know what "availability unknown" could possibly mean. @perexg and @libinyang any idea?

plbossart:
For the gain issue, I will have to ask @singalsu and @juimonen. We must have fixed this for HDaudio+DMIC some 1.5 years back and I don't recall anything.

@aigilea Do you see a control called "Dmic0 Capture Volume" and can the value be changed? I don't see a default value for it, maybe that's something missing?

aigilea:
@plbossart thank you for the hard work, testing is a pleasure to do, especially when it's about your problem.

I can see "Dmic0 CAPTURE" stereo control in alsamixer, I've maxed all controls on the capture screen. Pulseaudio source volume is maxed as well, I guess it's the same thing.

It's not _that_ bad actually, but speaking loudly right to where I think microphone is gets only third to half of the scale in the recorder while it should clip in this case IMO.
I don't have windows currently to check if maybe it's ok for this hardware, but if needed I can find another SSD and install it.

perexg:
I think that all is resolved now (via https://github.com/alsa-project/alsa-ucm-conf/pull/77). If something is missing, please, open a new PR.

singalsu:
@aigilea Raw unprocessed microphone signal is usually left to quite low level. Even loud voice should not clip. Noise suppressors, AEC etc. work a lot better with linear signal with headroom for ambient noise, portable device handling noise etc. In telephony (ITU-T domain) the RMS level of active parts for normal conversation should be around -22 dBFS or -16 dBm0. That corresponds to steady sine wave with amplitude 0.11 times full scale.

PR #76, opened 2021-01-30
Extend card search paths

lumag:
On Qualcomm platforms is is typical for a single kernel module to support several sound cards. Add support for using `ucm2/module/${KernelModule}/${CardLongName}.conf` as a use case config file.

See https://github.com/alsa-project/alsa-ucm-conf/pull/70 for the reference.


perexg:
Obsoleted with 4cfceb1e257a91cf5c1d29f28d1305607bbf58eb .

PR #78, opened 2021-02-03
Extend use case search paths to use kernel driver name

lumag:
This is a companion (or replacement) for #76 . This adds lookup for the use case search path using the in-kernel driver name. This allows easy lookup both for the setups which have the sound card/platform driver built as a module or builtin into the kernel.


perexg:
I would prefer the solution in https://github.com/alsa-project/alsa-ucm-conf/pull/79 . Any comments are welcome. Thank you for your ideas.

perexg:
Obsoleted by 4cfceb1e257a91cf5c1d29f28d1305607bbf58eb .

PR #85, opened 2021-02-27
broadwell-rt286, bdw-rt5677: Fix inconsistency in card detection

dpward:
This conveniently provides a workaround for a bug in kernel 5.10+:
https://bugzilla.kernel.org/show_bug.cgi?id=211985


perexg:
Merged. Thank you.

PR #89, opened 2021-04-19
[RFC] ucm2: rt711-sdca: remove 'rt711 FU0F Capture Switch' setting

libinyang:
The rt711-sdca codec driver removes the 'rt711 FU0F Capture Switch'
kcontrol. So we need to remove the setting of 'rt711 FU0F Capture Switch'
correspondingly.

Signed-off-by: Libin Yang <libin.yang@intel.com>


libinyang:
The codec driver patch is not merged. So the UCM patch is just for discussion now.

libinyang:
remove the PR as realtek has already fix the FU0F issue.

PR #92, opened 2021-05-20
Add UCMs for Google Asus Nexus 7 and  Acer Iconia A500

digetx:
@perexg As was discussed previouly at https://www.spinics.net/lists/alsa-devel/msg126382.html I'm opening PR with the UCMs for  Nexus 7 and A500. Both these devices are supported by the mainline Linux kernel. 

You're right that Tegra ASoC driver doesn't set card's name properly. In particular I found that we need to set the card's `driver_name` to `tegra`, then UCM get a proper lookup path.

I rebased UCMs on top of the recent alsa-ucm-conf master and switched Nexus 7 UCM to use generic RT5640 UCM. Everything works great once the card name of the kernel driver is corrected. I'll send the kernel patches ASAP for review.

Nexus 7 alsa-info: https://alsa-project.org/db/?f=0a29f1dab39ec5b33e01215dee42dbfd1dbbf08e
A500 alsa-info: https://alsa-project.org/db/?f=bde4f7938a39b1d26904b414969fc811e726242f

The alsa-info is taken using the updated kernel ASoC driver.

### DIFF
--- a/ucm2/codecs/rt5640/DigitalMics.conf
+++ b/ucm2/codecs/rt5640/DigitalMics.conf
@@ -1,9 +1,17 @@
 SectionDevice."Mic" {
 	Comment "Internal Digital Microphones"
 
-	ConflictingDevice [
-		"Headset"
-	]
+	If.have-headset {
+		Condition {
+			Type String
+			Empty "${var:HaveHeadsetMic}"
+		}
+		False {
+			ConflictingDevice [
+				"Headset"
+			]
+		}
+	}
 
 	EnableSequence [
 		cset "name='Mono ADC MIXL ADC2 Switch' on"
--- a/ucm2/codecs/rt5640/DigitalMics.conf
+++ b/ucm2/codecs/rt5640/DigitalMics.conf
@@ -10,7 +10,6 @@ SectionDevice."Mic" {
 		cset "name='Mono ADC MIXR ADC2 Switch' on"
 		cset "name='Stereo ADC MIXL ADC2 Switch' on"
 		cset "name='Stereo ADC MIXR ADC2 Switch' on"
-		cset "name='Internal Mic Switch' on"
 
 	]
 
@@ -19,7 +18,6 @@ SectionDevice."Mic" {
 		cset "name='Mono ADC MIXR ADC2 Switch' off"
 		cset "name='Stereo ADC MIXL ADC2 Switch' off"
 		cset "name='Stereo ADC MIXR ADC2 Switch' off"
-		cset "name='Internal Mic Switch' off"
 
 	]
 
--- a/ucm2/Tegra/rt5640/ASUS Google Nexus 7 ALC5642.conf
+++ b/ucm2/Tegra/rt5640/ASUS Google Nexus 7 ALC5642.conf
@@ -1,0 +1,15 @@
+# Use case Configuration for ASUS Google Nexus 7 (2012)
+
+Define.HaveAif1 "yes"
+Define.HaveAif2 ""
+Define.HaveSpeaker "yes"
+Define.HaveHeadsetMic ""
+Define.HaveInternalMic "yes"
+
+Include.hp.File "/codecs/rt5640/HeadPhones.conf"
+Include.mspk.File "/codecs/rt5640/Speaker.conf"
+Include.dmic.File "/codecs/rt5640/DigitalMics.conf"
+
+SectionVerb {
+	Include.e.File "/codecs/rt5640/EnableSeq.conf"
+}
--- a/ucm2/Tegra/wm8903/Acer Iconia Tab A500 WM8903.conf
+++ b/ucm2/Tegra/wm8903/Acer Iconia Tab A500 WM8903.conf
@@ -1,0 +1,114 @@
+# Use case Configuration for Acer Iconia Tab A500
+
+SectionVerb {
+	EnableSequence [
+		# WM8903 Output Configuration
+		cset "name='Digital Playback Volume' 80"
+		cset "name='Headphone Volume' 42"
+		cset "name='Line Out Volume' 42"
+		cset "name='DAC Boost Volume' 0"
+
+		# WM8903 Output Routing
+		cset "name='Left Output Mixer DACL Switch' on"
+		cset "name='Left Output Mixer DACR Switch' off"
+		cset "name='Right Output Mixer DACL Switch' off"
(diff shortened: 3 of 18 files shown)


ucm2/codecs/rt5640/DigitalMics.conf
```diff
@@ -10,7 +10,15 @@ SectionDevice."Mic" {
 		cset "name='Mono ADC MIXR ADC2 Switch' on"
 		cset "name='Stereo ADC MIXL ADC2 Switch' on"
 		cset "name='Stereo ADC MIXR ADC2 Switch' on"
-		cset "name='Internal Mic Switch' on"
+		If.Controls {
```
perexg, line 21: At first of all, thank you for your work. It seems like a right track to make things more maintainable. The BayTrail (Intel) specific controls should be moved to the upper level configs. @plbossart , FYI,

Also, `If` blocks cannot be in the sequences. You can define [] arrays multiple times and the sequences will be added to the end of array. So this is fine:

````
EnableSequence [
   sequence3
]
If.x {
   True.EnableSequence [
      sequence1
   ]
}
If.y {
  False.EnableSequence [
     sequence2
  ]
}
````

Note that Include / If blocks are evaluated as first (see the sequence numbers). That's the reason why we have some dummy (empty) string `If` blocks in the configs.

**Review by perexg — CHANGES_REQUESTED** [review]

**Comment by digetx** [comment]
Do you also mean that it's also possible to move the device sequences to `ucm2/bytcr-rt5640/HiFi.conf`?

```
SectionDevice."Headset" {
	If.HeadsetControls {
		Condition {
			Type ControlExists
			Control "name='Headset Mic Switch'"
		}
		True {
			cset "name='Headset Mic Switch' off"
		}
	}

	EnableSequence [
		If.Controls {
			Condition {
				Type ControlExists
				Control "name='Headset Mic Switch'"
			}
			True {
				cset "name='Headset Mic Switch' on"
			}
		}
	]

	DisableSequence [
		If.Controls {
			Condition {
				Type ControlExists
				Control "name='Headset Mic Switch'"
			}
			True {
				cset "name='Headset Mic Switch' off"
			}
		}
	]
}
```

**Comment by perexg** [comment]
> Do you also mean that it's also possible to move the device sequences to `ucm2/bytcr-rt5640/HiFi.conf`?

Yes, but the config should be like (`ucm2/bytcr-rt5640/HiFi-Components.conf`):

````
If.hsmic {
	Condition { Type String Empty "" }
	True {
		Include.hsmic.File "/codecs/rt5640/HeadsetMic.conf"
		If.switch {
			Condition {
				Type ControlExists
				Control "name='Headset Mic Switch'"
			}
			True {
				EnableSequence [
					cset "name='Headset Mic Switch' on"
				]
				DisableSequence [
					cset "name='Headset Mic Switch' off"
				]
			}
		}
	}
}
````

In this case `Include` is evaluated at first, so the cset will be appended to the corresponding Enable/Disable sequences.

But in this case, we know that `Headset Mic Switch` exists for this platform (bytcr-rt5640), so you can remove the control `If` completely and keep only Enable/Disable sequences.

**Comment by digetx** [comment]
@perexg I'm getting `I: [pulseaudio] (alsa-lib)main.c: error: failed to import hw:0 use case configuration -17` with:

```patch
If.spk {
	Condition {
		Type String
		Haystack "${CardComponents}"
		Needle "cfg-spk:2"
	}
	True {
		Include.spk.File "/codecs/rt5640/Speaker.conf"
		Define.HaveSpeaker "yes"

+		SectionDevice."Speaker" {
+			EnableSequence [
+				cset "name='Speaker Switch' on"
+			]
+
+			DisableSequence [
+				cset "name='Speaker Switch' off"
+			]
+		}
	}
}
```

alsa-lib doesn't like `SectionDevice` specified within conditions.  What should we do about it?

digetx:
Also, second `SectionDevice` definition overrides the previous definition, instead of appending it. I thought that  this may work:

```patch
If.spk {
	Condition {
		Type String
		Haystack "${CardComponents}"
		Needle "cfg-spk:2"
	}
	True {
		Include.spk.File "/codecs/rt5640/Speaker.conf"
+		Include.spk.File "/bytcr-rt5640/Speaker.conf"
		Define.HaveSpeaker "yes"
	}
}
```
but it's not working as expected.

Apparently syntax upgrade is required here. We could temporally use the version that I proposed with the conditions within the codec and move them to `/bytcr-rt5640` once syntax will allow to do that.

perexg:
Oops, I fixed this in 6b728405697e38339773a6901ca8f557c89f0088 . The merging was previously handled in only one level, thus:

````
SectionDevice."Speaker" {
	Comment "Speakers"
	EnableSequence [
		shell "echo 1st"
	]
	Value {
		PlaybackChannels 2
		PlaybackPriority 200
		PlaybackPCM "hw:0"
	}
}

If.0 {
	# Does not work before alsa-lib's commit 6b728405697e38339773a6901ca8f557c89f0088
	Condition { Type String Empty "" }
	True {
		SectionDevice."Speaker" {
			EnableSequence [
				shell "echo 2nd"
			]
		}
	}
}

SectionDevice."Speaker" {
	EnableSequence [
		shell "echo 3rd"
	]
}

SectionDevice."Speaker" {
	If.1 {
		Condition { Type String Empty "" }
		True {
			EnableSequence [
				shell "echo 4rd"
			]
		}
	}
}
````

````
$ alsaucm -c test set _verb HiFi set _enadev Speaker
1st
3rd
2nd
4rd
````

Note the order.

If `If.0` block is commented (or removed), it works with the previous master:

````
$ alsaucm -c test set _verb HiFi set _enadev Speaker
1st
3rd
4rd
````

digetx:
@perexg Thank you very much, it works now using the latest alsa-lib master. Will you backport that change to a stable alsa-lib version or bump the syntax version?

ucm2/codecs/rt5640/DigitalMics.conf
digetx: I pushed the updated version that implements yours suggestion, it works using the latest alsa-lib master. Please take a look.

perexg:
I've merged your changes with the 4884ef82c2ea27942a51411d0606f5f0866dcd2e fix / cleanups for bytcr-rt5640 . The internal variables (define) sections must be evaluated before the device includes. Thank you for your contribution.

digetx:
@perexg Thank you, everything working great using the latest alsa-ucm-conf + alsa-lib master.

PR #101, opened 2021-06-17
HDA-Intel: add Boost volume control for Headset Mic

jason77-wang:
Ubuntu Linux users complain even they set the input volume to maximum,
they still can't record sound with a pretty satisfied volume, they
need to run commandline commands to adjust "Mic Boost/Headset Mic
Boost/Headphone Mic boost" value to increase the input volume.

That is because the current ucm only defines "Capture Volume" to
control the input volume. Here we add Boost control, then users could
adjust both "Capture Volume" and "Boost Volume" through Pulseaudio and
Gnome.

BugLink: https://bugs.launchpad.net/bugs/1930188


perexg:
Applied in 89c4e449ab80f8b5f083743e109301ff595db92d . Thank you.

PR #102, opened 2021-06-21
Add UCM for Librem 5 Devkit

agx:
The DTS configuration is in Linux mainline since 5.12:

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/boot/dts/freescale/imx8mq-librem5-devkit.dts?h=v5.12#n166

This is based on initial work by Bob Ham.

Signed-off-by: Guido Günther <agx@sigxcpu.org>

### DIFF
--- a/ucm2/NXP/iMX8/Librem_5_Devkit/HiFi.conf
+++ b/ucm2/NXP/iMX8/Librem_5_Devkit/HiFi.conf
@@ -1,0 +1,74 @@
+SectionVerb {
+	EnableSequence [
+		cset "name='Digital Input Mux' I2S"
+		cset "name='PCM Playback Volume' 144"
+
+		cset "name='AVC Switch' off"
+		cset "name='Headphone Playback ZC Switch' off"
+
+		cset "name='Capture Attenuate Switch (-6dB)' off"
+		cset "name='Capture Mux' MIX_IN"
+		cset "name='Capture ZC Switch' off"
+		cset "name='Capture Switch' On"
+		cset "name='Capture Volume' 15,15"
+	]
+	DisableSequence [
+		cset "name='PCM Playback Volume' 0"
+	]
+	Value {
+	}
+}
+
+SectionDevice."Handset".0 {
+	Comment "Handset"
+
+	ConflictingDevice [
+		"Headset"
+	]
+
+	EnableSequence [
+		cset "name='Mic Mux MUX' Input 2"
+		cset "name='Lineout Playback Switch' on"
+		cset "name='Lineout Playback Volume' 18"
+		cset "name='Mic Volume' 2"
+	]
+	DisableSequence [
+		cset "name='Lineout Playback Switch' off"
+		cset "name='Mic Volume' 0"
+	]
+	Value {
+		PlaybackPriority "1000"
+		PlaybackPCM "hw:${CardId},0"
+		PlaybackVolume "name='Lineout Playback Volume'"
+		PlaybackSwitch "name='Lineout Playback Switch'"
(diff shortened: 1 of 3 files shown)


perexg:
Please, organize the files in this way:

1) move the config to `NXP/iMX8/Librem_5_Devkit/` tree
2) create a symlink from `conf.d/<ALSA_DRIVER_NAME>/Librem 5 Devkit.conf` to `NXP/iMX8/Librem_5_Devkit/Librem 5 Devkit.conf`

`ALSA_DRIVER_NAME` is probably `asoc-simple-card`, but it should be checked, see https://github.com/alsa-project/alsa-ucm-conf/tree/master/ucm2#readme .

Thank you.

agx:
@perexg done. For reference:

```
# cat /proc/asound/cards 
 0 [Devkit         ]: Librem_5_Devkit - Librem 5 Devkit
                      Librem 5 Devkit
 1 [SIM7100        ]: SIMCom_SIM7100 - SIMCom SIM7100
                      SIMCom SIM7100
```

(The SIM7100 is the modem). I've also adjusted the reference to upstream Linux git since the the current configuration got merged in 5.13-rc1.

perexg:
Unfortunately, `Librem_5_Devkit` is a wrong driver name. The information is lost. It does not make sense to have three strings with the samilar information. I would really prefer to keep the name of the driver which creates the ASoC card.

agx:
@perexg the strings come straight from the kernel via tha `simple-audio-card,name` Device-Tree property via the simple-card driver and those are *all* just derived from the card name, see

https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/sound/soc/soc-core.c?h=v5.13-rc1#n1924

We can't just use the codec in device tree since then we can't distinguish from other devices using the same codec however I noticed that asoc-simple-driver doesn't fill in the driver_name so the card name is used there and send a kernel patch:

https://lore.kernel.org/lkml/YNGe3akAntQi8qJD@qwark.sigxcpu.org/

If that works things would move to `ucm2/conf.d/simple-card/Librem 5 Devkit.conf`

agx:
@perexg the kernel change is already on it's way to linux-next (https://lore.kernel.org/alsa-devel/162437355278.1773.13469789646711861169.b4-ty@kernel.org/) so i updated to the patch accordingly.

agx:
The fixed driver name now [landed in linux-next](https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/commit/?id=907f0a3051869a61499905377212500155bd28ec) so we now have

```
# cat /proc/asound/cards 
 0 [Devkit         ]: simple-card - Librem 5 Devkit
                     Librem 5 Devkit
...
```

and the change is hopefullly o.k. to apply.

agx:
@perexg can i do anything else to get this applied?

agx:
card name fix is now in linus tree https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=907f0a3051869a61499905377212500155bd28ec so the names used here match what gets reported from mainline linux.

perexg:
I merged this in 9ae4a7a9f0712db1cb5934eecc905c5c7930260c , but the volume controls do not follow the latest UCM conventions. I added PlaybackMixerElem in 0910aa744e258de6889b2023611fcfa2b630b14e, but some control setup should be moved to the  BootSequence (including the default volume settings) and the volume controls for capture may be handled better.

agx:
@perexg thanks. I'll recheck and see what can be moved.

PR #107, opened 2021-07-23
sof-hda-dsp: Set Capture Switch on in the BootSequence

jason77-wang:
We found an issue that the PA source of Mic2/Headset and Mic2 is
muted by default after newly install an OS, the root cause is the
'Capture Switch' is set to off in the kernel.

Without ucm, the /usr/share/alsa/init/default will set the 'Capture
Switch' to on, similarly we set it to on in the BootSequence of ucm.

### DIFF
--- a/ucm2/sof-hda-dsp/sof-hda-dsp.conf
+++ b/ucm2/sof-hda-dsp/sof-hda-dsp.conf
@@ -57,3 +57,14 @@ If.Dmic0 {
 		cset "name='Dmic0 Capture Volume' 70%"
 	]
 }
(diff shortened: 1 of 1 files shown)


KaiChuan-Hsieh:
@perexg 

hello,
would you please review it? it is kernel driver update to make the capture switch to be default off. We'd like to align previous user experience. Is it possible to merge the MP?

Thanks,

perexg:
Merged in d73a015f25f133879f2106eb8ce2e523043a84ea . Thank you.

PR #109, opened 2021-07-26
sof-soundwire: rt1316 - DAC L and R is replaced by DAC

jason77-wang:
In the latest codec driver rt1316-sdw.c, the DAC L and DAC R are
replaced by DAC, so we change the ucm to add the condition checking,
if 'DAC L' exists, will set those mixers in the old way, otherwise,
set the mixers with new name.

The codec rt1316 is used on a Dell TGL soundwire audio machine.

### DIFF
--- a/ucm2/sof-soundwire/rt1316-1.conf
+++ b/ucm2/sof-soundwire/rt1316-1.conf
@@ -3,17 +3,36 @@
 SectionDevice."Speaker" {
 	Comment	"Speaker"
 
-	EnableSequence [
-		cset "name='rt1316-1 DAC L Switch' 1"
-		cset "name='rt1316-1 DAC R Switch' 1"
-		cset "name='Speaker Switch' on"
-	]
+	If.lrswitch {
+		Condition {
+			Type ControlExists
+			Control "name='rt1316-1 DAC L Switch'"
+		}
+		True {
+			EnableSequence [
+				cset "name='rt1316-1 DAC L Switch' 1"
+				cset "name='rt1316-1 DAC R Switch' 1"
+				cset "name='Speaker Switch' on"
+			]
 
-	DisableSequence [
-		cset "name='rt1316-1 DAC L Switch' 0"
-		cset "name='rt1316-1 DAC R Switch' 0"
-		cset "name='Speaker Switch' off"
-	]
+			DisableSequence [
+				cset "name='rt1316-1 DAC L Switch' 0"
+				cset "name='rt1316-1 DAC R Switch' 0"
+				cset "name='Speaker Switch' off"
+			]
+		}
+		False {
+			EnableSequence [
+				cset "name='rt1316-1 DAC Switch' 1"
+				cset "name='Speaker Switch' on"
+			]
+
+			DisableSequence [
+				cset "name='rt1316-1 DAC Switch' 0"
+				cset "name='Speaker Switch' off"
+			]
+		}
+	}
 
 	Value {
 	      PlaybackPriority 100
--- a/ucm2/sof-soundwire/rt1316-2.conf
+++ b/ucm2/sof-soundwire/rt1316-2.conf
@@ -3,24 +3,46 @@
 SectionDevice."Speaker" {
 	Comment	"Speaker"
 
-	EnableSequence [
-		cset "name='rt1316-1 RX Channel Select' L,L"
-		cset "name='rt1316-2 RX Channel Select' R,R"
+	If.lrswitch {
+		Condition {
+			Type ControlExists
+			Control "name='rt1316-1 DAC L Switch'"
+		}
+		True {
+			EnableSequence [
+				cset "name='rt1316-1 RX Channel Select' L,L"
+				cset "name='rt1316-2 RX Channel Select' R,R"
+				cset "name='rt1316-1 DAC L Switch' 1"
+				cset "name='rt1316-1 DAC R Switch' 1"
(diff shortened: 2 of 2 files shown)


ucm2/sof-soundwire/rt1316-1.conf
```diff
@@ -3,17 +3,36 @@
...
+		}
+		False {
+			EnableSequence [
+				cset "name='rt1316-1 DAC Switch' 1"
+				cset "name='Speaker Switch' on"
+			]
+
+			DisableSequence [
+				cset "name='rt1316-1 DAC Switch' 0"
+				cset "name='Speaker Switch' off"
```
plbossart, line 32: I am not following this.

a) @jason77-wang can you clarify what you meant by 'latest codec driver'? In the upstream code I don't think we ever had this DAC L control? What upstream commit changed this? If this was from an earlier non-upstream version, can we instead just use the newer control names? 

b) @libinyang didn't we test this before?

libinyang: @plbossart Yes, this config should be tested. The config is from https://github.com/thesofproject/alsa-ucm-conf/pull/45. From the comment https://github.com/thesofproject/alsa-ucm-conf/pull/45#issuecomment-706979644, I tested it last year and it passed. And this is only tested on sof kernel, not on the upstream kernel. I'm not sure whether the upstream has some patches before to support DAC L control and whether there are products on the market which is based on SOF kernel. I'm trying to flash back the old kernels to have a check.

libinyang: I have checked and our SOF kernel did use DAC L control before.

plbossart: maybe @libinyang but that was with early drivers. has anyone tested since we merged Mark Brown's for-next branch at the end of February? My point is to try to identify if we have a coverage gap...

libinyang: > maybe @libinyang but that was with early drivers. has anyone tested since we merged Mark Brown's for-next branch at the end of February? My point is to try to identify if we have a coverage gap...

@plbossart I'm afraid not. Our UCM DUT in CI only covers hda mode currently (we had a DUT to cover SDW mode before, but the DUT was returned to the customer) because of lack of machines. And I haven't done the manual test for a long time on the rt1316 platforms. Your question is a good point. I will talk with Mengdong about whether we can find more platforms to cover more UCM configs in CI. Thanks.

plbossart: @libinyang I think if we provide configurations for each device we have, the UCM validator would find cases that don't work. It's not necessarily manual or requiring CI, just making use the controls are registered somewhere so that problems in UCM are identified by @perexg 's tools.

libinyang: > @libinyang I think if we provide configurations for each device we have, the UCM validator would find cases that don't work. It's not necessarily manual or requiring CI, just making sure the controls are registered somewhere so that problems in UCM are identified by @perexg 's tools.

@plbossart Thanks for the suggestion. We have submitted the configurations to https://github.com/alsa-project/alsa-tests/tree/master/python/ucm-validator/configs/sof-soundwire and I checked the configurations uses the new control name. I will check how to make use of the tool.

plbossart: @libinyang  I don't see any TGL-based configurations, only CML ones. I think that's part of the disconnect.

jason77-wang: @plbossart and @libinyang,  the "DAC L control" was in the sof kernel before, and canonical backported the driver to the ubuntu linux-oem-5.10.y kernel. So the issue we face currently is ubuntu linux-oem-5.10 kernel has DAC L/R, ubuntu linux-5.13.0-generic and later kernel has DAC, so I submitted this PR to support both DAC L/R and DAC.

Do I need to remove DAC L/R in this PR?

thx.

jason77-wang: @plbossart and @libinyang,  the "DAC L control" was in the sof kernel before, and canonical backported the driver to the ubuntu linux-oem-5.10.y kernel. So the issue we face currently is ubuntu linux-oem-5.10 kernel has DAC L/R, ubuntu linux-5.13.0-generic and later kernel has DAC, so I submitted this PR to support both DAC L/R and DAC.

Do I need to remove DAC L/R in this PR?

thx.

libinyang: > @plbossart and @libinyang, the "DAC L control" was in the sof kernel before, and canonical backported the driver to the ubuntu linux-oem-5.10.y kernel. So the issue we face currently is ubuntu linux-oem-5.10 kernel has DAC L/R, ubuntu linux-5.13.0-generic and later kernel has DAC, so I submitted this PR to support both DAC L/R and DAC.
> 
> Do I need to remove DAC L/R in this PR?
> 
> thx.

@jason77-wang I preferred to retain the DAC L/R kcontrols. Otherwise, customers may find it not work any more after upgrade alsa-ucm-config.

jason77-wang: @libinyang , Got it. Thanks.

perexg:
Merged in f88cc42b9cc2f8a773d409bdccdccdacbce16ef4 . Thank you.

PR #120, opened 2021-11-15
ucm2: Add UCM support for Dell Desktop

shumingfan:
The Dell Desktop uses two USB audio codecs as front and rear panel.
The front UAC has two outputs of speaker/headphone and one input of headset mic.
The rear UAC has one output of line-out.

Signed-off-by: Shuming Fan <shumingf@realtek.com>


perexg:
I merged this one as an initial version to the repository with 9320885d52757dd311ae5e0a53cbdf92af6f103e, but looking to the config, it may be worth to try to join the two front configs into one. Have you tried it?

Also, please, test again using the latest configs in this repository.

shumingfan:
I updated the commits which joins two configs into one.
I copy the latest ucm2 configs in Ubuntu 20.04. It can't match the right UCM except I modifies the ucm.conf that changes Syntax 4 to Syntax 3.
I also tested on Fedora 35 which uses alsa-lib-1.2.5.1. The system can match the right UCM config without modification.

perexg:
Thanks. Applied to the main repo.

PR #134, opened 2022-01-16
Add UCM for PinePhone

### LINKED ISSUE #124: PinePhone support
It'd be great to have PinePhone support in alsa-ucm-conf. https://gitlab.com/pine64-org/pine64-alsa-ucm/-/issues/3 is the canonical source of PinePhone UCM.

tomfitzhenry:
Fixes https://github.com/alsa-project/alsa-ucm-conf/issues/124.

This PR adds support for the https://wiki.pine64.org/index.php/PinePhone 

This PR contains:
* This UCM config copied from https://gitlab.com/pine64-org/pine64-alsa-ucm/, which was developed by many developers ( https://gitlab.com/pine64-org/pine64-alsa-ucm/-/merge_requests/1 ) and is BSD-3-Clause licensed.
* Fix paths to be relative to alsa-ucm-conf root.
* Some refactoring of the UCM config, incl. removing duplicate directives, and moving volume directives to BootSequence.
* Updating the routing/playback/boost volumes, in response to comments on this PR.

I have tested:

* HiFi
    * Record audio via the headset.
    * Record audio via the internal mic.
    * Play audio on the headphones.
    * Play audio on the speaker.
    * Play audio on the internal speaker.
* Voice Call
    * Speak on the headset.
    * Speak on the internal mic.
    * Listen on the headphones.
    * Listen on the speaker. This induces an echo, and given the audio routing is entirely on chip, and the chip doesn't advertise echo cancellation, I don't see how this is avoidable.
    * Listen on the internal speaker.

See also:
* `alsa-info`: https://alsa-project.org/db/?f=5e331f5ef10bd167b3bd7640bfcf0e879100f80d
* `pa-info`: [pa-info.txt](https://github.com/alsa-project/alsa-ucm-conf/files/8015083/pa-info.txt)
* https://gitlab.com/pine64-org/pine64-alsa-ucm/-/issues/3
* Schematics/code
    * The device-tree configuration has had a unique audio card name since 5.14: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/boot/dts/allwinner/sun50i-a64-pinephone.dtsi?h=v5.14#n436
    * http://files.pine64.org/doc/PinePhone/PinePhone%20v1.2%20Released%20Schematic.pdf
    * https://xnux.eu/devices/feature/audio-pp.html
    * https://github.com/torvalds/linux/blob/master/arch/arm64/boot/dts/allwinner/sun50i-a64-pinephone.dtsi
    * https://github.com/torvalds/linux/blob/master/arch/arm64/boot/dts/allwinner/sun50i-a64.dtsi
    * https://github.com/torvalds/linux/blob/master/sound/soc/sunxi/sun50i-codec-analog.c
    * https://files.pine64.org/doc/datasheet/pine64/Allwinner_A64_User_Manual_V1.0.pdf
    * https://static.chipdip.ru/lib/126/DOC001126567.pdf

### DIFF
--- a/ucm2/Allwinner/A64/PinePhone/HiFi.conf
+++ b/ucm2/Allwinner/A64/PinePhone/HiFi.conf
@@ -3,24 +3,6 @@ SectionVerb {
 		cset "name='AIF2 Digital DAC Playback Switch' off"
 		cset "name='AIF2 ADC Mixer ADC Capture Switch' off"
 	]
-
-	SectionDefaults [
-		# Switch playback off.
-		cset "name='Earpiece Playback Switch' off"
-		cset "name='Headphone Playback Switch' off"
-		cset "name='Line In Playback Switch' off"
-		cset "name='Line Out Playback Switch' off"
-		cset "name='Mic1 Playback Switch' off"
-		cset "name='Mic2 Playback Switch' off"
-
-		# Switch capture off.
-		cset "name='Line In Capture Switch' off"
-		cset "name='Mic1 Capture Switch' off"
-		cset "name='Mic2 Capture Switch' off"
-		cset "name='Mixer Capture Switch' off"
-		cset "name='Mixer Reversed Capture Switch' off"
-	]
-
 }
 
 SectionDevice."Speaker" {
--- a/ucm2/Allwinner/A64/PinePhone/HiFi.conf
+++ b/ucm2/Allwinner/A64/PinePhone/HiFi.conf
@@ -18,7 +18,6 @@ SectionDevice."Speaker" {
 	Value {
 		PlaybackVolume "Line Out Playback Volume"
 		PlaybackSwitch "Line Out Playback Switch"
-		PlaybackChannels 2
 		PlaybackPriority 300
 		PlaybackPCM "hw:${CardId},0"
 	}
@@ -38,7 +37,6 @@ SectionDevice."Earpiece" {
 	Value {
 		PlaybackVolume "Earpiece Playback Volume"
 		PlaybackSwitch "Earpiece Playback Switch"
-		PlaybackChannels 2
 		PlaybackPriority 200
 		PlaybackPCM "hw:${CardId},0"
 	}
@@ -62,7 +60,6 @@ SectionDevice."Mic" {
 	Value {
 		CapturePriority 100
 		CapturePCM "hw:${CardId},0"
-		CaptureChannels 2
 		CaptureVolume "ADC Capture Volume"
 		CaptureSwitch "Mic1 Capture Switch"
 	}
@@ -86,7 +83,6 @@ SectionDevice."Headset" {
 	Value {
 		CapturePriority 500
 		CapturePCM "hw:${CardId},0"
-		CaptureChannels 2
 		CaptureVolume "ADC Capture Volume"
 		CaptureSwitch "Mic2 Capture Switch"
 		JackControl "Headset Microphone Jack"
@@ -107,7 +103,6 @@ SectionDevice."Headphones" {
 	Value {
 		PlaybackVolume "Headphone Playback Volume"
 		PlaybackSwitch "Headphone Playback Switch"
-		PlaybackChannels 2
 		PlaybackPriority 500
 		PlaybackPCM "hw:${CardId},0"
 		JackControl "Headphone Jack"
--- a/ucm2/Allwinner/A64/PinePhone/HiFi.conf
+++ b/ucm2/Allwinner/A64/PinePhone/HiFi.conf
@@ -1,0 +1,101 @@
+SectionVerb {
+	EnableSequence [
+		cset "name='AIF2 Digital DAC Playback Switch' off"
+		cset "name='AIF2 ADC Mixer ADC Capture Switch' off"
+	]
+}
+
+SectionDevice."Speaker" {
+	Comment "Internal speaker"
+	EnableSequence [
+		cset "name='Line Out Playback Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='Line Out Playback Switch' off"
+	]
+
+	Value {
+		PlaybackVolume "Line Out Playback Volume"
+		PlaybackSwitch "Line Out Playback Switch"
+		PlaybackChannels 2
+		PlaybackPriority 300
+		PlaybackPCM "hw:${CardId},0"
+	}
+}
+SectionDevice."Earpiece" {
+	Comment "Internal Earpiece"
+	EnableSequence [
+		cset "name='Earpiece Playback Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='Earpiece Playback Switch' off"
+	]
+
+	Value {
+		PlaybackVolume "Earpiece Playback Volume"
+		PlaybackSwitch "Earpiece Playback Switch"
+		PlaybackChannels 2
+		PlaybackPriority 200
+		PlaybackPCM "hw:${CardId},0"
+	}
+}
+SectionDevice."Mic" {
+	Comment "Internal Microphone"
+	ConflictingDevice [
+		"Headset"
+	]
+	EnableSequence [
+		cset "name='Mic1 Capture Switch' on"
+	]
+	DisableSequence [
+		cset "name='Mic1 Capture Switch' off"
+	]
+	Value {
+		CapturePriority 100
+		CapturePCM "hw:${CardId},0"
+		CaptureChannels 2
+		CaptureVolume "ADC Capture Volume"
+		CaptureSwitch "Mic1 Capture Switch"
+	}
+}
+SectionDevice."Headset" {
+	Comment "Headset Microphone"
+	ConflictingDevice [
+		"Mic"
+	]
+	EnableSequence [
+		cset "name='Mic2 Capture Switch' on"
+	]
+	DisableSequence [
+		cset "name='Mic2 Capture Switch' off"
+	]
+	Value {
+		CapturePriority 500
+		CapturePCM "hw:${CardId},0"
+		CaptureChannels 2
+		CaptureVolume "ADC Capture Volume"
+		CaptureSwitch "Mic2 Capture Switch"
+		JackControl "Headset Microphone Jack"
+	}
+}
+SectionDevice."Headphones" {
+	Comment "Headphones"
+	EnableSequence [
+		cset "name='Headphone Playback Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='Headphone Playback Switch' off"
+	]
+
+	Value {
+		PlaybackVolume "Headphone Playback Volume"
+		PlaybackSwitch "Headphone Playback Switch"
+		PlaybackChannels 2
+		PlaybackPriority 500
+		PlaybackPCM "hw:${CardId},0"
+		JackControl "Headphone Jack"
+	}
+}
--- a/ucm2/Allwinner/A64/PinePhone/HiFi.conf
+++ b/ucm2/Allwinner/A64/PinePhone/HiFi.conf
@@ -34,8 +34,7 @@ SectionDevice."Speaker" {
 	]
 
 	Value {
-		PlaybackVolume "Line Out Playback Volume"
-		PlaybackSwitch "Line Out Playback Switch"
+		PlaybackMixerElem "Line Out"
 		PlaybackPriority 300
 		PlaybackPCM "hw:${CardId},0"
 	}
@@ -53,8 +52,7 @@ SectionDevice."Earpiece" {
 	]
 
 	Value {
-		PlaybackVolume "Earpiece Playback Volume"
-		PlaybackSwitch "Earpiece Playback Switch"
+		PlaybackMixerElem "Earpiece"
 		PlaybackPriority 200
 		PlaybackPCM "hw:${CardId},0"
 	}
@@ -119,8 +117,7 @@ SectionDevice."Headphones" {
 	]
 
 	Value {
-		PlaybackVolume "Headphone Playback Volume"
-		PlaybackSwitch "Headphone Playback Switch"
+		PlaybackMixerElem "Headphone"
 		PlaybackPriority 500
 		PlaybackPCM "hw:${CardId},0"
 		JackControl "Headphone Jack"
--- a/ucm2/Allwinner/A64/PinePhone/HiFi.conf
+++ b/ucm2/Allwinner/A64/PinePhone/HiFi.conf
@@ -3,6 +3,24 @@ SectionVerb {
 		cset "name='AIF2 Digital DAC Playback Switch' off"
 		cset "name='AIF2 ADC Mixer ADC Capture Switch' off"
 	]
+
+	SectionDefaults [
+		# Switch playback off.
+		cset "name='Earpiece Playback Switch' off"
+		cset "name='Headphone Playback Switch' off"
+		cset "name='Line In Playback Switch' off"
+		cset "name='Line Out Playback Switch' off"
+		cset "name='Mic1 Playback Switch' off"
+		cset "name='Mic2 Playback Switch' off"
+
+		# Switch capture off.
+		cset "name='Line In Capture Switch' off"
+		cset "name='Mic1 Capture Switch' off"
+		cset "name='Mic2 Capture Switch' off"
+		cset "name='Mixer Capture Switch' off"
+		cset "name='Mixer Reversed Capture Switch' off"
+	]
+
 }
 
 SectionDevice."Speaker" {
--- a/ucm2/Allwinner/A64/PinePhone/HiFi.conf
+++ b/ucm2/Allwinner/A64/PinePhone/HiFi.conf
@@ -23,8 +23,10 @@ SectionDevice."Speaker" {
 		PlaybackPCM "hw:${CardId},0"
 	}
 }
+
 SectionDevice."Earpiece" {
 	Comment "Internal Earpiece"
+
 	EnableSequence [
 		cset "name='Earpiece Playback Switch' on"
 	]
@@ -41,17 +43,22 @@ SectionDevice."Earpiece" {
 		PlaybackPCM "hw:${CardId},0"
 	}
 }
+
 SectionDevice."Mic" {
 	Comment "Internal Microphone"
+
 	ConflictingDevice [
 		"Headset"
 	]
+
 	EnableSequence [
 		cset "name='Mic1 Capture Switch' on"
 	]
+
 	DisableSequence [
 		cset "name='Mic1 Capture Switch' off"
 	]
+
 	Value {
 		CapturePriority 100
 		CapturePCM "hw:${CardId},0"
@@ -60,17 +67,22 @@ SectionDevice."Mic" {
 		CaptureSwitch "Mic1 Capture Switch"
 	}
 }
+
 SectionDevice."Headset" {
 	Comment "Headset Microphone"
+
 	ConflictingDevice [
 		"Mic"
 	]
+
 	EnableSequence [
 		cset "name='Mic2 Capture Switch' on"
 	]
+
 	DisableSequence [
 		cset "name='Mic2 Capture Switch' off"
 	]
+
 	Value {
 		CapturePriority 500
 		CapturePCM "hw:${CardId},0"
@@ -80,8 +92,10 @@ SectionDevice."Headset" {
 		JackControl "Headset Microphone Jack"
 	}
 }
+
 SectionDevice."Headphones" {
 	Comment "Headphones"
+
 	EnableSequence [
 		cset "name='Headphone Playback Switch' on"
 	]
--- a/ucm2/Allwinner/A64/PinePhone/PinePhone.conf
+++ b/ucm2/Allwinner/A64/PinePhone/PinePhone.conf
@@ -1,0 +1,65 @@
+Syntax 2
+
+# https://wiki.pine64.org/index.php/PinePhone
+# https://files.pine64.org/doc/PinePhone/PinePhone%20v1.2%20Released%20Schematic.pdf
+# https://xnux.eu/devices/feature/audio-pp.html
+
+SectionUseCase."HiFi" {
+	File "/Allwinner/A64/PinePhone/HiFi.conf"
+	Comment "Play HiFi quality music"
+}
+
+SectionUseCase."Voice Call" {
+	File "/Allwinner/A64/PinePhone/VoiceCall.conf"
+	Comment "Make a phone call"
+}
+
+FixedBootSequence [
+	# Switch playback off.
+	cset "name='Earpiece Playback Switch' off"
+	cset "name='Headphone Playback Switch' off"
+	cset "name='Line In Playback Switch' off"
+	cset "name='Line Out Playback Switch' off"
+	cset "name='Mic1 Playback Switch' off"
+	cset "name='Mic2 Playback Switch' off"
+
+	# Switch capture off.
+	cset "name='Line In Capture Switch' off"
+	cset "name='Mic1 Capture Switch' off"
+	cset "name='Mic2 Capture Switch' off"
+	cset "name='Mixer Capture Switch' off"
+	cset "name='Mixer Reversed Capture Switch' off"
+
+	# Routing.
+	cset "name='ADC Digital DAC Playback Switch' off"
+	cset "name='AIF1 DA0 Stereo Playback Route' Stereo"
+	cset "name='AIF1 Data Digital ADC Capture Switch' on"
+	cset "name='AIF1 Slot 0 Digital DAC Playback Switch' on"
+	cset "name='AIF2 DAC Source Playback Route' AIF2"
+	# AIF2 (Modem) is mono.
+	cset "name='AIF2 DAC Stereo Playback Route' Mix Mono"
+	cset "name='AIF3 ADC Source Capture Route' None"
+	cset "name='DAC Playback Switch' on"
+	# Routes DACR->MIXL and DACL->MIXR => MIXL are MIXR are identical mono-mix of the DAC.
+	cset "name='DAC Reversed Playback Switch' on"
+	cset "name='Earpiece Source Playback Route' Left Mixer"
+	cset "name='Headphone Source Playback Route' DAC"
+	# The Pinephone speaker is mono.
+	cset "name='Line Out Source Playback Route' Mono Differential"
+]
+
+BootSequence [
+	# Playback volumes.
+	cset "name='AIF1 DA0 Playback Volume' 160"
+	cset "name='AIF2 DAC Playback Volume' 160"
+	cset "name='DAC Playback Volume' 160"
+	cset "name='Earpiece Playback Volume' 100%"
+	cset "name='Headphone Playback Volume' 70%"
+	cset "name='Line Out Playback Volume' 100%"
+	cset "name='Mic2 Boost Volume' 1"
+
+	# Capture volumes.
+	cset "name='ADC Capture Volume' 160"
+	cset "name='AIF1 AD0 Capture Volume' 160"
+	cset "name='AIF2 ADC Capture Volume' 160"
+]
--- a/ucm2/Allwinner/A64/PinePhone/PinePhone.conf
+++ b/ucm2/Allwinner/A64/PinePhone/PinePhone.conf
@@ -48,3 +48,20 @@ BootSequence [
 	cset "name='AIF1 AD0 Capture Volume' 160"
 	cset "name='AIF2 ADC Capture Volume' 160"
 ]
+
+SectionDefaults [
+	# Switch playback off.
+	cset "name='Earpiece Playback Switch' off"
+	cset "name='Headphone Playback Switch' off"
+	cset "name='Line In Playback Switch' off"
+	cset "name='Line Out Playback Switch' off"
+	cset "name='Mic1 Playback Switch' off"
+	cset "name='Mic2 Playback Switch' off"
+
+	# Switch capture off.
+	cset "name='Line In Capture Switch' off"
+	cset "name='Mic1 Capture Switch' off"
+	cset "name='Mic2 Capture Switch' off"
+	cset "name='Mixer Capture Switch' off"
+	cset "name='Mixer Reversed Capture Switch' off"
+]
--- a/ucm2/Allwinner/A64/PinePhone/PinePhone.conf
+++ b/ucm2/Allwinner/A64/PinePhone/PinePhone.conf
@@ -15,21 +15,6 @@ SectionUseCase."Voice Call" {
 }
 
 FixedBootSequence [
-	# Switch playback off.
-	cset "name='Earpiece Playback Switch' off"
-	cset "name='Headphone Playback Switch' off"
-	cset "name='Line In Playback Switch' off"
-	cset "name='Line Out Playback Switch' off"
-	cset "name='Mic1 Playback Switch' off"
-	cset "name='Mic2 Playback Switch' off"
-
-	# Switch capture off.
-	cset "name='Line In Capture Switch' off"
-	cset "name='Mic1 Capture Switch' off"
-	cset "name='Mic2 Capture Switch' off"
-	cset "name='Mixer Capture Switch' off"
-	cset "name='Mixer Reversed Capture Switch' off"
-
 	# Routing.
 	cset "name='ADC Digital DAC Playback Switch' off"
 	cset "name='AIF1 DA0 Stereo Playback Route' Stereo"
--- a/ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
+++ b/ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
@@ -4,6 +4,23 @@ SectionVerb {
 		cset "name='AIF2 ADC Mixer ADC Capture Switch' on"
 	]
 
+	SectionDefaults [
+		# Switch playback off.
+		cset "name='Earpiece Playback Switch' off"
+		cset "name='Headphone Playback Switch' off"
+		cset "name='Line In Playback Switch' off"
+		cset "name='Line Out Playback Switch' off"
+		cset "name='Mic1 Playback Switch' off"
+		cset "name='Mic2 Playback Switch' off"
+
+		# Switch capture off.
+		cset "name='Line In Capture Switch' off"
+		cset "name='Mic1 Capture Switch' off"
+		cset "name='Mic2 Capture Switch' off"
+		cset "name='Mixer Capture Switch' off"
+		cset "name='Mixer Reversed Capture Switch' off"
+	]
+
 	Value {
 		PlaybackRate 8000
 	}
--- a/ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
+++ b/ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
@@ -1,0 +1,105 @@
+SectionVerb {
+	EnableSequence [
+		cset "name='AIF2 Digital DAC Playback Switch' on"
+		cset "name='AIF2 ADC Mixer ADC Capture Switch' on"
+	]
+
+	Value {
+		PlaybackRate 8000
+	}
+}
+
+SectionDevice."Speaker" {
+	Comment "Internal speaker"
+	EnableSequence [
+		cset "name='Line Out Playback Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='Line Out Playback Switch' off"
+	]
+
+	Value {
+		PlaybackVolume "Line Out Playback Volume"
+		PlaybackSwitch "Line Out Playback Switch"
+		PlaybackChannels 2
+		PlaybackPriority 300
+		PlaybackPCM "hw:${CardId},0"
+	}
+}
+SectionDevice."Earpiece" {
+	Comment "Internal Earpiece"
+	EnableSequence [
+		cset "name='Earpiece Playback Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='Earpiece Playback Switch' off"
+	]
+
+	Value {
+		PlaybackVolume "Earpiece Playback Volume"
+		PlaybackSwitch "Earpiece Playback Switch"
+		PlaybackChannels 2
+		PlaybackPriority 500
+		PlaybackPCM "hw:${CardId},0"
+	}
+}
+SectionDevice."Mic" {
+	Comment "Internal Microphone"
+	ConflictingDevice [
+		"Headset"
+	]
+	EnableSequence [
+		cset "name='Mic1 Capture Switch' on"
+	]
+	DisableSequence [
+		cset "name='Mic1 Capture Switch' off"
+	]
+	Value {
+		CapturePriority 200
+		CapturePCM "hw:${CardId},0"
+		CaptureVolume "ADC Capture Volume"
+		CaptureSwitch "Mic1 Capture Switch"
+		CaptureChannels 2
+	}
+}
+SectionDevice."Headset" {
+	Comment "Headset Microphone"
+	ConflictingDevice [
+		"Mic"
+	]
+	EnableSequence [
+		cset "name='Mic2 Capture Switch' on"
+	]
+	DisableSequence [
+		cset "name='Mic2 Capture Switch' off"
+	]
+	Value {
+		CapturePriority 500
+		CapturePCM "hw:${CardId},0"
+		CaptureChannels 2
+		CaptureVolume "ADC Capture Volume"
+		CaptureSwitch "Mic2 Capture Switch"
+		JackControl "Headset Microphone Jack"
+	}
+}
+SectionDevice."Headphones" {
+	Comment "Headphones"
+	EnableSequence [
+		cset "name='Headphone Playback Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='Headphone Playback Switch' off"
+	]
+
+	Value {
+		PlaybackVolume "Headphone Playback Volume"
+		PlaybackSwitch "Headphone Playback Switch"
+		PlaybackChannels 2
+		PlaybackPriority 500
+		PlaybackPCM "hw:${CardId},0"
+		JackControl "Headphone Jack"
+	}
+}
--- a/ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
+++ b/ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
@@ -38,8 +38,7 @@ SectionDevice."Speaker" {
 	]
 
 	Value {
-		PlaybackVolume "Line Out Playback Volume"
-		PlaybackSwitch "Line Out Playback Switch"
+		PlaybackMixerElem "Line Out"
 		PlaybackPriority 300
 		PlaybackPCM "hw:${CardId},0"
 	}
@@ -57,8 +56,7 @@ SectionDevice."Earpiece" {
 	]
 
 	Value {
-		PlaybackVolume "Earpiece Playback Volume"
-		PlaybackSwitch "Earpiece Playback Switch"
+		PlaybackMixerElem "Earpiece"
 		PlaybackPriority 500
 		PlaybackPCM "hw:${CardId},0"
 	}
@@ -123,8 +121,7 @@ SectionDevice."Headphones" {
 	]
 
 	Value {
-		PlaybackVolume "Headphone Playback Volume"
-		PlaybackSwitch "Headphone Playback Switch"
+		PlaybackMixerElem "Headphone"
 		PlaybackPriority 500
 		PlaybackPCM "hw:${CardId},0"
 		JackControl "Headphone Jack"
--- a/ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
+++ b/ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
@@ -4,23 +4,6 @@ SectionVerb {
 		cset "name='AIF2 ADC Mixer ADC Capture Switch' on"
 	]
 
-	SectionDefaults [
-		# Switch playback off.
-		cset "name='Earpiece Playback Switch' off"
-		cset "name='Headphone Playback Switch' off"
-		cset "name='Line In Playback Switch' off"
-		cset "name='Line Out Playback Switch' off"
-		cset "name='Mic1 Playback Switch' off"
-		cset "name='Mic2 Playback Switch' off"
-
-		# Switch capture off.
-		cset "name='Line In Capture Switch' off"
-		cset "name='Mic1 Capture Switch' off"
-		cset "name='Mic2 Capture Switch' off"
-		cset "name='Mixer Capture Switch' off"
-		cset "name='Mixer Reversed Capture Switch' off"
-	]
-
 	Value {
 		PlaybackRate 8000
 	}
--- a/ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
+++ b/ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
@@ -11,6 +11,7 @@ SectionVerb {
 
 SectionDevice."Speaker" {
 	Comment "Internal speaker"
+
 	EnableSequence [
 		cset "name='Line Out Playback Switch' on"
 	]
@@ -27,8 +28,10 @@ SectionDevice."Speaker" {
 		PlaybackPCM "hw:${CardId},0"
 	}
 }
+
 SectionDevice."Earpiece" {
 	Comment "Internal Earpiece"
+
 	EnableSequence [
 		cset "name='Earpiece Playback Switch' on"
 	]
@@ -45,17 +48,22 @@ SectionDevice."Earpiece" {
 		PlaybackPCM "hw:${CardId},0"
 	}
 }
+
 SectionDevice."Mic" {
 	Comment "Internal Microphone"
+
 	ConflictingDevice [
 		"Headset"
 	]
+
 	EnableSequence [
 		cset "name='Mic1 Capture Switch' on"
 	]
+
 	DisableSequence [
 		cset "name='Mic1 Capture Switch' off"
 	]
+
 	Value {
 		CapturePriority 200
 		CapturePCM "hw:${CardId},0"
@@ -64,17 +72,22 @@ SectionDevice."Mic" {
 		CaptureChannels 2
 	}
 }
+
 SectionDevice."Headset" {
 	Comment "Headset Microphone"
+
 	ConflictingDevice [
 		"Mic"
 	]
+
 	EnableSequence [
 		cset "name='Mic2 Capture Switch' on"
 	]
+
 	DisableSequence [
 		cset "name='Mic2 Capture Switch' off"
 	]
+
 	Value {
 		CapturePriority 500
 		CapturePCM "hw:${CardId},0"
@@ -84,8 +97,10 @@ SectionDevice."Headset" {
 		JackControl "Headset Microphone Jack"
 	}
 }
+
 SectionDevice."Headphones" {
 	Comment "Headphones"
+
 	EnableSequence [
 		cset "name='Headphone Playback Switch' on"
 	]
--- a/ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
+++ b/ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
@@ -23,7 +23,6 @@ SectionDevice."Speaker" {
 	Value {
 		PlaybackVolume "Line Out Playback Volume"
 		PlaybackSwitch "Line Out Playback Switch"
-		PlaybackChannels 2
 		PlaybackPriority 300
 		PlaybackPCM "hw:${CardId},0"
 	}
@@ -43,7 +42,6 @@ SectionDevice."Earpiece" {
 	Value {
 		PlaybackVolume "Earpiece Playback Volume"
 		PlaybackSwitch "Earpiece Playback Switch"
-		PlaybackChannels 2
 		PlaybackPriority 500
 		PlaybackPCM "hw:${CardId},0"
 	}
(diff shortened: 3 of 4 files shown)


ucm2/Allwinner/A64/PinePhone/HiFi.conf
```diff
@@ -0,0 +1,146 @@
...
+		cset "name='AIF1 AD0 Capture Volume' 160"
+		cset "name='AIF1 Data Digital ADC Capture Switch' on"
+		cset "name='AIF2 ADC Mixer ADC Capture Switch' off"
+		cset "name='AIF2 ADC Mixer AIF1 DA0 Capture Switch' off"
+		cset "name='AIF2 ADC Mixer AIF2 DAC Rev Capture Switch' off"
+		cset "name='AIF2 ADC Mixer AIF1 DA0 Capture Switch' off"
+		cset "name='AIF2 ADC Mixer AIF1 DA0 Capture Switch' off"
+	]
+
+        DisableSequence [
```
tomfitzhenry, line 40: This uses spaces, which is inconsistent with the rest of the file (tabs).

tomfitzhenry: Done.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
```diff
@@ -0,0 +1,146 @@
...
+		cset "name='AIF2 Digital DAC Playback Switch' off"
+		cset "name='DAC Reversed Playback Switch' off"
+		cset "name='Earpiece Playback Switch' off"
+		cset "name='Earpiece Source Playback Route' DACL"
+
+		cset "name='Line In Capture Switch' off"
+		cset "name='Mic1 Capture Switch' off"
+		cset "name='Mic1 Boost Volume' 7"
+		cset "name='Mic2 Capture Switch' off"
+		cset "name='Mic2 Boost Volume' 7"
```
a-wai, line 26:
```suggestion
```

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
```diff
@@ -0,0 +1,151 @@
...
+		cset "name='Earpiece Playback Switch' off"
+		cset "name='Earpiece Source Playback Route' DACL"
+
+		cset "name='Line In Capture Switch' off"
+		cset "name='Mic1 Capture Switch' off"
+		cset "name='Mic1 Boost Volume' 0"
+		cset "name='Mic1 Playback Volume' 7"
+		cset "name='Mic2 Capture Switch' off"
+		cset "name='Mic2 Boost Volume' 0"
+		cset "name='Mic2 Playback Volume' 7"
```
a-wai, line 29:
```suggestion
```

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
```diff
@@ -0,0 +1,151 @@
...
+		cset "name='Mixer Capture Switch' off"
+		cset "name='Mixer Reversed Capture Switch' off"
+		cset "name='ADC Capture Volume' 160"
+		cset "name='ADC Gain Capture Volume' 7"
```
a-wai, line 33:
```suggestion
```

a-wai (changes requested):
Thanks for taking care of this!

I've added a few comments (mostly regarding mic boost and ADC gain) based on my testing (`HiFi` only for the latter, couldn't test your version of `VoiceCall` but instead suggested the values I use on a daily basis with Mobian).

tomfitzhenry:
> I've added a few comments (mostly regarding mic boost and ADC gain) based on my testing (HiFi only for the latter, couldn't test your version of VoiceCall but instead suggested the values I use on a daily basis with Mobian).

Thanks for your comprehensive review!

Your comments make sense. For each comment I will search through the commits of https://gitlab.com/pine64-org/pine64-alsa-ucm to see the directives have any justification.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
```diff
@@ -0,0 +1,146 @@
...
+		cset "name='Mixer Capture Switch' off"
+		cset "name='Mixer Reversed Capture Switch' off"
+		cset "name='ADC Capture Volume' 160"
+		cset "name='ADC Gain Capture Volume' 7"
```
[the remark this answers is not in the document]
tomfitzhenry: This was introduced in https://gitlab.com/postmarketOS/pmaports/-/merge_requests/1113/diffs.

From that merge request, it's not clear why this was introduced.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
```diff
@@ -0,0 +1,146 @@
...
+		cset "name='ADC Digital DAC Playback Switch' off"
+		cset "name='AIF1 Slot 0 Digital DAC Playback Switch' on"
+		cset "name='AIF2 Digital DAC Playback Switch' off"
+		cset "name='DAC Reversed Playback Switch' off"
+		cset "name='Earpiece Playback Switch' off"
+		cset "name='Earpiece Source Playback Route' DACL"
+
+		cset "name='Line In Capture Switch' off"
+		cset "name='Mic1 Capture Switch' off"
+		cset "name='Mic1 Boost Volume' 7"
```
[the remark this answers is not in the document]
tomfitzhenry: This was introduced in https://gitlab.com/postmarketOS/pmaports/-/merge_requests/1113/diffs.

From that merge request, it's not clear why this was introduced.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
tomfitzhenry: This was introduced in https://gitlab.com/postmarketOS/pmaports/-/merge_requests/1113/diffs.

From that merge request, it's not clear why this was introduced.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
tomfitzhenry: This was introduced in https://gitlab.com/postmarketOS/pmaports/-/merge_requests/1113/diffs.

From that merge request, it's not clear why this was introduced.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
```diff
@@ -0,0 +1,151 @@
...
+		cset "name='AIF1 AD0 Capture Volume' 160"
+		cset "name='AIF1 Data Digital ADC Capture Switch' on"
+		cset "name='AIF2 ADC Capture Volume' 160"
+		cset "name='AIF2 ADC Mixer ADC Capture Switch' on"
+		cset "name='AIF2 ADC Mixer AIF1 DA0 Capture Switch' off"
+		cset "name='AIF2 ADC Mixer AIF2 DAC Rev Capture Switch' off"
+		cset "name='AIF2 ADC Mixer AIF1 DA0 Capture Switch' off"
+		cset "name='AIF2 ADC Mixer AIF1 DA0 Capture Switch' off"
```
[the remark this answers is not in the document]
tomfitzhenry: This was introduced in https://gitlab.com/postmarketOS/pmaports/-/merge_requests/938/diffs and I think it's safe to remove.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
```diff
@@ -0,0 +1,146 @@
...
+		cset "name='Mixer Reversed Capture Switch' off"
+		cset "name='ADC Capture Volume' 160"
+		cset "name='ADC Gain Capture Volume' 7"
+		cset "name='AIF1 AD0 Capture Volume' 160"
+		cset "name='AIF1 Data Digital ADC Capture Switch' on"
+		cset "name='AIF2 ADC Mixer ADC Capture Switch' off"
+		cset "name='AIF2 ADC Mixer AIF1 DA0 Capture Switch' off"
+		cset "name='AIF2 ADC Mixer AIF2 DAC Rev Capture Switch' off"
+		cset "name='AIF2 ADC Mixer AIF1 DA0 Capture Switch' off"
+		cset "name='AIF2 ADC Mixer AIF1 DA0 Capture Switch' off"
```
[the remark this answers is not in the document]
tomfitzhenry: This was introduced in https://gitlab.com/postmarketOS/pmaports/-/merge_requests/938/diffs and I think it's safe to remove.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
```diff
@@ -0,0 +1,146 @@
...
+		CaptureSwitch "Mic2 Capture Switch"
+		JackControl "Headset Microphone Jack"
+	}
+}
+SectionDevice."Headphones" {
+	Comment "Headphones"
+	EnableSequence [
+		cset "name='AIF1 DA0 Stereo Playback Route' Stereo"
+		cset "name='Headphone Playback Switch' on"
+		cset "name='Headphone Playback Volume' 70%"
```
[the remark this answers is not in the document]
tomfitzhenry: This was introduced in https://gitlab.com/postmarketOS/pmaports/-/merge_requests/2182 and is well discussed. I will test, and discuss with the pmos folk on this.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
```diff
@@ -0,0 +1,151 @@
...
+		cset "name='AIF1 Slot 0 Digital DAC Playback Switch' on"
+		cset "name='AIF2 Digital DAC Playback Switch' on"
+		cset "name='DAC Reversed Playback Switch' off"
+		cset "name='Earpiece Playback Switch' off"
+		cset "name='Earpiece Source Playback Route' DACL"
+
+		cset "name='Line In Capture Switch' off"
+		cset "name='Mic1 Capture Switch' off"
+		cset "name='Mic1 Boost Volume' 0"
+		cset "name='Mic1 Playback Volume' 7"
```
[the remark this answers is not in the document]
tomfitzhenry: This was introduced in https://gitlab.com/postmarketOS/pmaports/-/merge_requests/1652 and is well discussed. I will test, and discuss with the pmos folks on this.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
tomfitzhenry: This was introduced in https://gitlab.com/postmarketOS/pmaports/-/merge_requests/1652 and is well discussed. I will test, and discuss with the pmos folks on this.

tomfitzhenry:
The remaining comments are non-trivial changes.

I will test the Pinephone audio with the comments having applied.

@a-wai, I see Mobian is already using the comments as you've described ( https://gitlab.com/mobian1/packages/alsa-ucm-conf/-/blob/mobian/debian/patches/0001-ucm2-add-PinePhone-configuration.patch ) which gives confidence that they'll work.

For each profile, I will test:
* Headphones
* Internal speaker
* Internal earpiece
* Internal mic
* Mic via audio jack

ucm2/Allwinner/A64/PinePhone/HiFi.conf
```diff
@@ -0,0 +1,144 @@
...
+		cset "name='AIF2 ADC Mixer AIF1 DA0 Capture Switch' off"
+		cset "name='AIF2 ADC Mixer AIF2 DAC Rev Capture Switch' off"
+	]
+
+	DisableSequence [
+	]
+
+	Value {
+	}
+}
```
tomfitzhenry, line 43: Can these empty sections be removed?

ucm2/Allwinner/A64/PinePhone/HiFi.conf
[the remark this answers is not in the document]
smaeul: This control is the gain for the headphone analog power amplifier. I don't think the units here are dB relative to input (it's an amplifier after all), but dB relative to maximum amplification. So this gain should be as small as possible to minimize noise.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
```diff
@@ -0,0 +1,144 @@
...
+		PlaybackSwitch "Line Out Playback Switch"
+		PlaybackChannels 2
+		PlaybackPriority 300
+		PlaybackPCM "hw:${CardId},0"
+	}
+}
+SectionDevice."Earpiece" {
+	Comment "Internal Earpiece"
+	EnableSequence [
+		cset "name='AIF1 DA0 Stereo Playback Route' Mix Mono"
```
smaeul, line 68: The "Speaker" and "Earpiece" sections don't conflict with "Headphones". So enabling both "Speaker" and "Headphones" would give you mix mono audio to the headphones. I think a nice solution here is, at the HiFi profile level:
 * Enable "DAC Playback Switch"
 * Enable "DAC Reversed Playback Switch"
 * Set "AIF1 DA0 Stereo Playback Route" to "Stereo"
 * Set "Headphone Source Playback Route" to "DAC"
 * Set "Earpiece Source Playback Route" to "Left Mixer"

Then, the headphones will always be stereo, because they bypass the analog mixer. At the same time, since both DAC channels will go into both sides of the analog mixer, the earpiece and speaker will get mix mono.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
```diff
@@ -0,0 +1,144 @@
...
+		cset "name='Headphone Playback Switch' off"
+		cset "name='Headphone Source Playback Route' DAC"
+		cset "name='Line In Playback Switch' off"
+		cset "name='Line Out Playback Switch' off"
+		cset "name='Line Out Source Playback Route' Mono Differential"
+		cset "name='Mic1 Playback Switch' off"
+		cset "name='Mic2 Playback Switch' off"
+		cset "name='AIF1 DA0 Playback Volume' 160"
+		cset "name='AIF3 ADC Source Capture Route' None"
+		cset "name='AIF2 DAC Source Playback Route' None"
```
smaeul, line 12: "None" is not a valid value for this control. When not using AIF3, this should be set to "AIF2". See here: https://github.com/torvalds/linux/blob/master/sound/soc/sunxi/sun8i-codec.c#L844-L855

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
```diff
@@ -0,0 +1,149 @@
...
+		cset "name='Mic1 Playback Switch' off"
+		cset "name='Mic2 Playback Switch' off"
+		cset "name='AIF1 DA0 Playback Volume' 160"
+		cset "name='AIF2 DAC Playback Volume' 160"
+		cset "name='AIF3 ADC Source Capture Route' None"
+		cset "name='AIF2 DAC Source Playback Route' None"
+		cset "name='DAC Playback Switch' on"
+		cset "name='DAC Playback Volume' 160"
+		cset "name='ADC Digital DAC Playback Switch' off"
+		cset "name='AIF1 Slot 0 Digital DAC Playback Switch' on"
```
smaeul, line 17: AIF1 Slot 0 is the CPU audio interface. Is it intentional that this is enabled during a call?

tomfitzhenry:
For posterity, here's the resources I'm reading to learn more about ALSA/UCM/Pinephone:
* http://files.pine64.org/doc/PinePhone/PinePhone%20v1.2%20Released%20Schematic.pdf
* https://xnux.eu/devices/feature/audio-pp.html
* https://github.com/torvalds/linux/blob/master/arch/arm64/boot/dts/allwinner/sun50i-a64-pinephone.dtsi
* https://github.com/torvalds/linux/blob/master/arch/arm64/boot/dts/allwinner/sun50i-a64.dtsi
* https://github.com/torvalds/linux/blob/master/sound/soc/sunxi/sun50i-codec-analog.c
* https://files.pine64.org/doc/datasheet/pine64/Allwinner_A64_User_Manual_V1.0.pdf

ucm2/Allwinner/A64/PinePhone/HiFi.conf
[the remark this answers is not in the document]
a-wai: Ah right, my "0dB obsession" fooled me again! So 70% should be fine here, my bad.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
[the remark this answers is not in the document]
tomfitzhenry: I lean towards committing this suggestion, on the basis that we should prefer boosting the analog signal over boosting the digital signal. The change in https://gitlab.com/postmarketOS/pmaports/-/merge_requests/1652 appears to have been an over-correction of https://gitlab.com/postmarketOS/pmaports/-/merge_requests/1113/diffs.

I will test this.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
tomfitzhenry: I lean towards committing this suggestion, on the basis that we should prefer boosting the analog signal over boosting the digital signal. In https://github.com/alsa-project/alsa-ucm-conf/pull/134/files#r800623320, the analog signal boost will occur.

I will test this.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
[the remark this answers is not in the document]
tomfitzhenry: I lean towards committing this suggestion, on the basis that we should prefer boosting the analog signal over boosting the digital signal. In https://github.com/alsa-project/alsa-ucm-conf/pull/134/files#r800623320, the analog signal boost will occur.

I will test this.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
```diff
@@ -0,0 +1,149 @@
...
+		cset "name='Headphone Source Playback Route' DAC"
+		cset "name='Line In Playback Switch' off"
+		cset "name='Line Out Playback Switch' off"
+		cset "name='Line Out Source Playback Route' Mono Differential"
+		cset "name='Mic1 Playback Switch' off"
+		cset "name='Mic2 Playback Switch' off"
+		cset "name='AIF1 DA0 Playback Volume' 160"
+		cset "name='AIF2 DAC Playback Volume' 160"
+		cset "name='AIF3 ADC Source Capture Route' None"
+		cset "name='AIF2 DAC Source Playback Route' None"
```
tomfitzhenry, line 13: `None` is invalid, per https://github.com/alsa-project/alsa-ucm-conf/pull/134/files#r802213404.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
tomfitzhenry: SGTM, per https://github.com/alsa-project/alsa-ucm-conf/pull/134/files#r800623320.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
tomfitzhenry: If "DAC Reversed Playback Switch" is enabled on the HiFi profile, and an XMPP video/audio chat is active, then AFAICT the output would be fed into the input, as if it came from the mic.

AFAICT, an XMPP video/audio chat wouldn't want to use the Voice Call profile, since the Voice Call profile sets "PlaybackRate 8000".

As an alternative, should we just set that "Speaker" and "Earpiece" each conflict with "Headphones". I think this matches user expectations too.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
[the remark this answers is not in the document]
tomfitzhenry: Upon testing ( `arecord -D 'sysdefault:CARD=PinePhone' -c2 -f S16_LE  foo.wav; aplay -D 'sysdefault:CARD=PinePhone' foo.wav`) I observe that 33dB boost is acceptably loud, and 42dB boost induces noise.

I'll commit this suggestion.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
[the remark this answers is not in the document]
tomfitzhenry: Upon testing ( `arecord -D 'sysdefault:CARD=PinePhone' -c2 -f S16_LE  foo.wav; aplay -D 'sysdefault:CARD=PinePhone' foo.wav`) I observe that default 0db gain is acceptably loud, and the extra 6dB boost induces noise.

I'll commit this suggestion.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
tomfitzhenry: Upon testing ( `arecord -D 'sysdefault:CARD=PinePhone' -c2 -f S16_LE  foo.wav; aplay -D 'sysdefault:CARD=PinePhone' foo.wav`) I observe that 33dB boost is acceptably loud, and 42dB boost induces noise.

I'll commit this suggestion.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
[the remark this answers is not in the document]
tomfitzhenry: Upon testing ( `arecord -D 'sysdefault:CARD=PinePhone' -c2 -f S16_LE  foo.wav; aplay -D 'sysdefault:CARD=PinePhone' foo.wav`) I observe that the default are acceptably loud.

I'll commit this suggestion.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
tomfitzhenry: Introduced in https://gitlab.com/postmarketOS/pmaports/-/merge_requests/1614, and looks to pre-date smaeul's work.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
```diff
@@ -0,0 +1,143 @@
...
+		CaptureSwitch "Mic2 Capture Switch"
+		JackControl "Headset Microphone Jack"
+	}
+}
+SectionDevice."Headphones" {
+	Comment "Headphones"
+	EnableSequence [
+		cset "name='AIF1 DA0 Stereo Playback Route' Stereo"
+		cset "name='Headphone Playback Switch' on"
+		cset "name='Headphone Playback Volume' 100%"
```
tomfitzhenry, line 128: Should this be 70%, like as decided in https://github.com/alsa-project/alsa-ucm-conf/pull/134/files#r800616927 ?

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
```diff
@@ -0,0 +1,143 @@
...
+		cset "name='AIF1 AD0 Capture Volume' 160"
+		cset "name='AIF1 Data Digital ADC Capture Switch' on"
+		cset "name='AIF2 ADC Capture Volume' 160"
+		cset "name='AIF2 ADC Mixer ADC Capture Switch' on"
+		cset "name='AIF2 ADC Mixer AIF1 DA0 Capture Switch' off"
+		cset "name='AIF2 ADC Mixer AIF2 DAC Rev Capture Switch' off"
+	]
+
+	Value {
+		PlaybackRate 8000
```
tomfitzhenry, line 40: This is the only directive that meaningfully differs between VoiceCall and HiFi. Perhaps the similarity can be deduplicated (e.g. move to PinePhone.conf somehow?)

ucm2/Allwinner/A64/PinePhone/HiFi.conf
smaeul: "DAC Playback Switch" and "DAC Reversed Playback Switch" control the exact same path from the DACs to the analog Output Mixer, except that the left and right channels are swapped. They are the two pairs of "m"s overlaying the rightward-pointing arrows just to the right of the green "DACL" and "DACR" boxes in this diagram: https://xnux.eu/devices/feature/audio-controls.svg.

The "Reversed" here refers to the stereo channels, not to the direction.

(The only loopback paths in the codec are controlled by the "Mixer (Reversed) Capture Switch"; see the leftward-pointing arrows at the bottom of the diagram. Those are always off in both profiles.)

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
tomfitzhenry: I have tested this with a headset, and observe that with any `Mic2 Boost Volume` that is >0 induces non-trivial noise.

The best I can do is:
* Mic2 Playback Volume at 0 dB
* Mic2 Boost Volume at 0 dB
* ADC Capture Volume at 24 dB (max)
* ADC Gain at 6dB (max)

That comes out loud and clear.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
tomfitzhenry: See https://github.com/alsa-project/alsa-ucm-conf/pull/134/files#r805993961 for discussion.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
tomfitzhenry: Thanks for the explanation, that makes sense.

I like this approach because it means our routing can be only: routes don't change, only switches do. That makes this config easier to understand.

I have updated this PR accordingly: https://github.com/alsa-project/alsa-ucm-conf/pull/134/commits/fb6e94983e69b902c66306214a4cfab3beabb7fa.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
tomfitzhenry: Perhaps folks want to record their calls? Or maybe GNOME Calls uses the VoiceCall even for SIP?

What's the disadvantage of leaving this enabled?

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
tomfitzhenry: I've applied your suggestion.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
tomfitzhenry: I've applied your suggestion.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
smaeul: I suppose there isn't any downside. This won't cause any additional hardware to be powered up, since the capture path must be enabled anyway for recording.

One note though on call recording: you will need to enable "AIF2 Digital ADC Capture Switch" in order to capture audio from the modem. ("AIF1 Data Digital ADC Capture Switch" only gets you audio from the analog codec, i.e. the microphone).

tomfitzhenry: Done.

tomfitzhenry:
@a-wai, @smaeul: Thanks for your reviews, I've addressed your comments.

From here I will:
* do a final test
* add some documentation, and maybe an updated https://xnux.eu/devices/feature/modem-audio-route.svg
* seek review from the ALSA UCM folk, in particular around how we could de-duplicate the config between the two Usecases.

I'd appreciate if you found the time to do a final review and/or test.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
tomfitzhenry: The default Mic2 Boost Volume seems to be 33 dB, not 0 dB. Maybe we will need to set `Mic2 Boost Volume` to 0. Is this better done in the kernel?

tomfitzhenry:
Two issues:
* commit ec8e6ced80711b4824c36d059b9d2cc5682bbebe "PinePhone: HiFi: Match VoiceCall's Mic1 CapturePriority" appears to have broken capture selection when selecting a capture device during a GNOME Calls call. I will remove this commit.
* The headset mic is too noisy. I've unresolved the relevant comment.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
tomfitzhenry: Fixed by explicitly setting `Mic2 Boost Volume` to 0 dB.

tomfitzhenry:
> add some documentation, and maybe an updated https://xnux.eu/devices/feature/modem-audio-route.svg

@megous, do you know which license (if any) this diagram is available under? I'd like to modify it and include it as part of this PR.

philmmanjaro:
If I compare this UCM config to the one by Mobian, I still see some differences: [mobian.diff.txt](https://github.com/alsa-project/alsa-ucm-conf/files/8172438/mobian.diff.txt) Also people reported this: The microphone on phone calls is super sensitive and in the settings I could not lower to any effect. This made calls nearly impossible. Installing pavucontrol (pulse audio volume control) and lowering mic input to about 15% returned calls to normal. If you use Sound Record on Phosh you have some high pitch background noise during the recording.

philmmanjaro:
Also you can't mute the mic in **gnome-calls**. On a pinephone-pro with their UCM config it seems to work. See also here: https://gitlab.com/mobian1/callaudiod/-/merge_requests/10

philmmanjaro:
Seems it also depends if you use **pulseaudio** or **pipewire**. At least **Sound Recorder** may behave differently.

philmmanjaro:
Differences from the Pine64 UCM profile: [p64u.diff.txt](https://github.com/alsa-project/alsa-ucm-conf/files/8172909/p64u.diff.txt)

ucm2/Allwinner/A64/PinePhone/HiFi.conf
```diff
@@ -0,0 +1,135 @@
...
+		cset "name='AIF2 ADC Mixer AIF2 DAC Rev Capture Switch' off"
+		cset "name='AIF2 Digital ADC Capture Switch' on"
+	]
+}
+
+SectionDevice."Speaker" {
+	Comment "Internal speaker"
+	EnableSequence [
+		cset "name='Line Out Playback Switch' on"
+		cset "name='Line Out Playback Volume' 100%"
```
tomfitzhenry, line 43: "It is not recommended to set the fixed values for the volume settings to the Enable / Disable sequences for verbs or devices, if the device exports the hardware volume (MixerElem or Volume/Switch values). The default volume settings should be set in the BootSequence." https://www.alsa-project.org/alsa-doc/alsa-lib/group__ucm__conf.html

ucm2/Allwinner/A64/PinePhone/HiFi.conf
```diff
@@ -0,0 +1,135 @@
...
+		PlaybackChannels 2
+		PlaybackPriority 300
+		PlaybackPCM "hw:${CardId},0"
+	}
+}
+SectionDevice."Earpiece" {
+	Comment "Internal Earpiece"
+	EnableSequence [
+		cset "name='Earpiece Playback Switch' on"
+		cset "name='Earpiece Playback Volume' 100%"
```
tomfitzhenry, line 62: "It is not recommended to set the fixed values for the volume settings to the Enable / Disable sequences for verbs or devices, if the device exports the hardware volume (MixerElem or Volume/Switch values). The default volume settings should be set in the BootSequence." https://www.alsa-project.org/alsa-doc/alsa-lib/group__ucm__conf.html

ucm2/Allwinner/A64/PinePhone/HiFi.conf
```diff
@@ -0,0 +1,135 @@
...
+		CaptureVolume "ADC Capture Volume"
+		CaptureSwitch "Mic2 Capture Switch"
+		JackControl "Headset Microphone Jack"
+	}
+}
+SectionDevice."Headphones" {
+	Comment "Headphones"
+	EnableSequence [
+		cset "name='Headphone Playback Switch' on"
+		cset "name='Headphone Playback Volume' 70%"
```
tomfitzhenry, line 120: "It is not recommended to set the fixed values for the volume settings to the Enable / Disable sequences for verbs or devices, if the device exports the hardware volume (MixerElem or Volume/Switch values). The default volume settings should be set in the BootSequence." https://www.alsa-project.org/alsa-doc/alsa-lib/group__ucm__conf.html

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
```diff
@@ -0,0 +1,141 @@
...
+	}
+}
+
+SectionDevice."Speaker" {
+	Comment "Internal speaker"
+	EnableSequence [
+		cset "name='Line Out Playback Switch' on"
+		cset "name='Line Out Playback Volume' 100%"
```
tomfitzhenry, line 49: "It is not recommended to set the fixed values for the volume settings to the Enable / Disable sequences for verbs or devices, if the device exports the hardware volume (MixerElem or Volume/Switch values). The default volume settings should be set in the BootSequence." https://www.alsa-project.org/alsa-doc/alsa-lib/group__ucm__conf.html

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
```diff
@@ -0,0 +1,141 @@
...
+		PlaybackChannels 2
+		PlaybackPriority 300
+		PlaybackPCM "hw:${CardId},0"
+	}
+}
+SectionDevice."Earpiece" {
+	Comment "Internal Earpiece"
+	EnableSequence [
+		cset "name='Earpiece Playback Switch' on"
+		cset "name='Earpiece Playback Volume' 100%"
```
tomfitzhenry, line 68: "It is not recommended to set the fixed values for the volume settings to the Enable / Disable sequences for verbs or devices, if the device exports the hardware volume (MixerElem or Volume/Switch values). The default volume settings should be set in the BootSequence." https://www.alsa-project.org/alsa-doc/alsa-lib/group__ucm__conf.html

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
```diff
@@ -0,0 +1,141 @@
...
+		CaptureVolume "ADC Capture Volume"
+		CaptureSwitch "Mic2 Capture Switch"
+		JackControl "Headset Microphone Jack"
+	}
+}
+SectionDevice."Headphones" {
+	Comment "Headphones"
+	EnableSequence [
+		cset "name='Headphone Playback Switch' on"
+		cset "name='Headphone Playback Volume' 70%"
```
tomfitzhenry, line 126: "It is not recommended to set the fixed values for the volume settings to the Enable / Disable sequences for verbs or devices, if the device exports the hardware volume (MixerElem or Volume/Switch values). The default volume settings should be set in the BootSequence." https://www.alsa-project.org/alsa-doc/alsa-lib/group__ucm__conf.html

ucm2/Allwinner/A64/PinePhone/HiFi.conf
```diff
@@ -0,0 +1,135 @@
...
+		cset "name='AIF1 Slot 0 Digital DAC Playback Switch' on"
+		cset "name='AIF2 Digital DAC Playback Switch' off"
+		cset "name='DAC Reversed Playback Switch' on"
+		cset "name='Earpiece Playback Switch' off"
+		cset "name='Earpiece Source Playback Route' Left Mixer"
+
+		cset "name='Line In Capture Switch' off"
+		cset "name='Mic1 Capture Switch' off"
+		cset "name='Mic2 Capture Switch' off"
+		cset "name='Mic2 Boost Volume' 0"
```
tomfitzhenry, line 26: I found a headset where Mic2 Boost Volume = 0 results in no audio. (Maybe it always did?)

philmmanjaro:
I normally use pine64-ucm but also explore what upstream or others do. Seems at least we got some useful feedback from users.

bhush9:
Regarding mute,

> Ah, good spot. I wasn't aware of this usecase. I'm trying to prioritise upstreaming, rather than support new usecases. Do you know if this has ever worked for the Pinephone?

@tomfitzhenry I found fix for this, If you add `CaptureMixerElem "ADC"` in the capture nodes, hardware mute seems to work fine, which makes muting audio call work perfectly (at least verified using parecord and there it seems to work fine).

Can you add this in your UCM?

https://gitlab.com/pine64-org/pine64-alsa-ucm/-/merge_requests/7

ucm2/Allwinner/A64/PinePhone/HiFi.conf
tomfitzhenry: Done.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
tomfitzhenry: Done.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
tomfitzhenry: Done.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
tomfitzhenry: Done.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
tomfitzhenry: This was a faulty headset.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
tomfitzhenry: Done.

tomfitzhenry:
> Can you add this in your UCM?
>
> https://gitlab.com/pine64-org/pine64-alsa-ucm/-/merge_requests/7

Per https://gitlab.com/pine64-org/pine64-alsa-ucm/-/merge_requests/7#note_881261109 , I'll hold off including that PR for now.

tomfitzhenry:
I've tested this PR, and it's ready for review. I have updated the PR description with the testing done.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
```diff
@@ -0,0 +1,129 @@
...
+		cset "name='AIF1 DA0 Stereo Playback Route' Stereo"
+		cset "name='AIF3 ADC Source Capture Route' None"
+		cset "name='AIF2 DAC Source Playback Route' AIF2"
+		# AIF2 (Modem) is mono.
+		cset "name='AIF2 DAC Stereo Playback Route' Mix Mono"
+		cset "name='DAC Playback Switch' on"
+		cset "name='ADC Digital DAC Playback Switch' off"
+		cset "name='AIF1 Slot 0 Digital DAC Playback Switch' on"
+		cset "name='AIF2 Digital DAC Playback Switch' off"
+		cset "name='DAC Reversed Playback Switch' on"
```
a-wai, line 19: May I assume this routes `DACR` to `MIXL` (and `DACL` to `MIXR`) so `MIX*` are both identival and effectively a mono mix of the DAC outputs?

a-wai (changes requested):
Sorry it took so long for a new round of review.

I'm still thinking the volume/gain controls should be reworked (i.e. more similar to what distros currently use), but otherwise it looks good, with some nice additions.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
smaeul: Yes, that is correct.

ucm2/Allwinner/A64/PinePhone/PinePhone.conf
```diff
@@ -0,0 +1,28 @@
...
+
+SectionUseCase."Voice Call" {
+	File "/Allwinner/A64/PinePhone/VoiceCall.conf"
+	Comment "Make a phone call"
+}
+
+BootSequence [
+	cset "name='AIF1 DA0 Playback Volume' 100%"
+	cset "name='AIF2 DAC Playback Volume' 100%"
+	cset "name='DAC Playback Volume' 100%"
```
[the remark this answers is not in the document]
tomfitzhenry: My bad, I converted `160` to `100%`, mistakenly believed it was a no-op. Fixed.

ucm2/Allwinner/A64/PinePhone/PinePhone.conf
```diff
@@ -0,0 +1,28 @@
...
+	cset "name='Earpiece Playback Volume' 100%"
+	cset "name='Headphone Playback Volume' 70%"
+	cset "name='Line Out Playback Volume' 100%"
+	cset "name='Mic2 Boost Volume' 1"
+
+	cset "name='ADC Capture Volume' 100%"
+	cset "name='AIF1 AD0 Capture Volume' 100%"
```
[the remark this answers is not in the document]
tomfitzhenry: > Same comment here: those controls are just fine using 160 / 0dB, but make audio unusable when set to 100% (even with Mic1 Boost Volume set to 0)

My bad, I converted `160` to `100%`, mistakenly believed it was a no-op. Fixed.

> Mic1 Boost: not set here, assuming 0 (0dB)

The default is 3. I will confirm this.

Mic2 (headset) appears to need less boost on the headsets I have tried. I will re-confirm.

> As a general rule of thumb, I believe you should start with all volume controls on 0dB (except for the headphones output for the reasons discussed here already) and add small touches only where needed, starting with mic boosts.

Sounds good, this was the intention. The conversion of `160` to `100%` broke this, however. Now fixed.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
tomfitzhenry: I will document this.

ucm2/Allwinner/A64/PinePhone/HiFi.conf
```diff
@@ -0,0 +1,129 @@
...
+		cset "name='Line In Playback Switch' off"
+		cset "name='Line Out Playback Switch' off"
+		cset "name='Line Out Source Playback Route' Mono Differential"
+		cset "name='Mic1 Playback Switch' off"
+		cset "name='Mic2 Playback Switch' off"
+		cset "name='AIF1 DA0 Stereo Playback Route' Stereo"
+		cset "name='AIF3 ADC Source Capture Route' None"
+		cset "name='AIF2 DAC Source Playback Route' AIF2"
+		# AIF2 (Modem) is mono.
+		cset "name='AIF2 DAC Stereo Playback Route' Mix Mono"
```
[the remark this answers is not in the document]
tomfitzhenry: I'm keen to do this. I tried it with FixedBootSequence, but this appeared to occur before the driver had loaded. I will try BootSequence for the switches, and read https://www.alsa-project.org/alsa-doc/alsa-lib/group__ucm__conf.html to understand how this interacts with users who choose to override some volume/switches.

ucm2/Allwinner/A64/PinePhone/PinePhone.conf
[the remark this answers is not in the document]
a-wai: > My bad, I converted `160` to `100%`, mistakenly believed it was a no-op. Fixed.

Oh ok, I understand, those things can be misleading indeed.

> > Mic1 Boost: not set here, assuming 0 (0dB)
> 
> The default is 3. I will confirm this.

I believe you're right, it should default to whatever the hardware sets, which is 33db (`3`) in this case. It should be all good then.

> Mic2 (headset) appears to need less boost on the headsets I have tried. I will re-confirm.

Finding the "right" value here might be tricky as there can be big variations between headsets. Maybe stick with the default? (although I have no strong opinion on that one, and my headset's mic is broken so I can't really test).

> Sounds good, this was the intention. The conversion of `160` to `100%` broke this, however. Now fixed.

Got it, thanks!

ucm2/Allwinner/A64/PinePhone/HiFi.conf
a-wai: Awesome, thanks! (that's a clever use of this output mixer)

ucm2/Allwinner/A64/PinePhone/HiFi.conf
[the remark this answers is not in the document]
a-wai: `FixedBootSequence` would be nice indeed, I wonder why it didn't work when you tested. In this case `BootSequence` might hopefully work just as well, but that's to be thoroughly tested.

ollieparanoid:
Thanks everybody who worked on this!

@tomfitzhenry: you marked this as draft, what's the reason for that? Can we help with getting this finished up?

ollieparanoid:
@tomfitzhenry, again thanks a bunch for all your work on this! I've tested it on postmarketOS edge. The headset mic seems to be routed wrong, everything else works fine. I've tested with a headset with a four-contact (TRRS) plug.

| Test | Result |
| --- | --- |
|  HiFi |  |
| * Record audio via the headset. | NOK (1) |
| * Record audio via the internal mic. | OK |
| * Play audio on the headphones. | OK (2) |
| * Play audio on the speaker. | OK |
| * Play audio on the internal speaker. | OK |
| Voice Call | |
| * Speak on the headset. | NOK (3) |
| * Speak on the internal mic. | OK |
| * Listen on the headphones. | OK |
| * Listen on the speaker. | OK (4) | 
| * Listen on the internal speaker. | OK |

Notes:

1: tried with gnome sound recorder, the line just stays completely flat (wrong routing? see also 3)

2: it's a bit louder than the default we currently have in postmarketOS. We had turned the default down because as it was at max, one could hear noise in a quiet room. If I concentrate, I think I can hear some noise very quietly now, but it's so quiet it's fine in my opinion.

3: routing seems to be wrong here. I tried a call and could hear me talking into the headset on the pinephone itself instead of on the other phone. So for some reason it gets routed to the headset output?

4: I can reproduce the echo you described above.


The config looks fine from a quick glance (though I don't really have experience with reading/writing alsa ucm configs). In my opinion, once the headset mic bug is fixed, this is ready to be merged.

perexg:
The single commit is preferred for the initial work, Thank you.

ollieparanoid:
> > I'm away-from-Pinephone at the moment, so won't be able to re-test the NOK tests for a few weeks.
> 
> I've re-tested these, and I observe that they work for me.

Nice!

> @ollieparanoid, if you could re-test these cases in light of the comments I left in [#134 (comment)](https://github.com/alsa-project/alsa-ucm-conf/pull/134#issuecomment-1220231690) that'd be appreciated.

Yes, sorry for not getting back to that earlier. I should be able to test it over the weekend.

ollieparanoid:
Tested again:

| Test 					| Result |
| --- 					| --- |
|  HiFi 				|  |
| * Record audio via the headset.	| ??? (1) |
| * Record audio via the internal mic.	| OK |
| * Play audio on the headphones.	| OK |
| * Play audio on the speaker.		| OK |
| * Play audio on the internal speaker. | OK |
| Voice Call				| |
| * Speak on the headset.		| ??? (1) |
| * Speak on the internal mic.		| OK |
| * Listen on the headphones.		| OK |
| * Listen on the speaker.		| OK | 
| * Listen on the internal speaker.	| OK |

(1) So everything is fine for me except for the headset mic. I've been testing this for more than an hour now, and in that time I have gotten record audio via the headset working only once and by simply rebooting a lot of times and trying it again. When it worked, I had the headset plugged. But I wasn't able to reproduce it, every other time I've tried it just wasn't able to record anything from the mic again. I also tried a different headset in case this one was broken, but that wasn't the problem.

Also I found that this is not a regression with the config from this PR. When I run the alsa-ucm-conf we currently use in postmarketOS, I have the same problem (didn't realize/test this last time, typically I don't use the headset mic much). I'm not sure why this happens, might be a hardware defect with my particular device or anything else.

> Can you confirm that /var/lib/alsa/asound.state is non-existent at the time of this PR's config having been loaded? Otherwise lingering ALSA commands from previous config might be conflicting with this PR.

Yes, I can confirm that there was no asound.state file during the testing I did now, and probably also not when I tested last time.

> Also, could you try raising the Mic2 Boost Volume to 3 (which can be done via alsamixer)? This is the default, which I'm inclined to have this PR use (as has been suggested by reviewers)? Is the headset mic audible then?

I'm using pulseaudio and alsamixer didn't show the option there. I also didn't find it in pauvcontrol. I guess I would need to change the audio device in alsamixer? (if you want me to test it again, to what?) - but I'm not sure if it would change much given that most of the time the mic doesn't work at all (boost should only increase the volume gain, right?). And the one time it was working, I could hear it clearly, it wasn't very silent.

>> 3: routing seems to be wrong here. I tried a call and could hear me talking into the headset on the pinephone itself instead of on the other phone. So for some reason it gets routed to the headset output?
>
> Could this be sound captured from your other phone's mic? If I perform this test with my other phone close by, I also hear my voice on the headset, but if I mute my other phone, I no longer hear my voice on the headset.

Oh yes that was the reason! When I muted the other phone I couldn't hear anything while speaking to the headset mic.

So now that we know that the failures I had earlier are not regressions from changing to this config, and everything else works fine, I would recommend that this gets merged.

ollieparanoid (approved):

a-wai:
Finally took some time to test, sorry about the delay.

In my case everything works as expected (including headset mic, but we're not using the same kernel), although I had to manually reset the earpiece volume to 100%; this is likely due to me having set it to 0 a while ago (this is my test device, so I do all sorts of fiddling on that one) and therefore not caused by this PR.

I therefore grant this PR my "Approved" stamp ;)

a-wai (approved):

ucm2/Allwinner/A64/PinePhone/PinePhone.conf
```diff
@@ -0,0 +1,65 @@
...
+	Comment "Play HiFi quality music"
+}
+
+SectionUseCase."Voice Call" {
+	File "/Allwinner/A64/PinePhone/VoiceCall.conf"
+	Comment "Make a phone call"
+}
+
+FixedBootSequence [
+	# Switch playback off.
```
perexg, line 18: I don't think that this initialization scheme is quite correct. The verb EnableSequence or global SectionDefaults sequence should disable the controlled devices  - https://www.alsa-project.org/alsa-doc/alsa-lib/group__ucm__conf.html - to put the hardware to a good initial state before any device from the verb is used. Move the "turn mixer paths off" commands to SectionDefaults for multiple verbs. It will be executed before any verb is activated.

The other controls may be moved to the standard BootSequence, too. But if they are not supposed (or dangerous) to be modified by user later, you can keep them here. Note that alsactl should restore the previous state of all ALSA driver controls on boot.

ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
```diff
@@ -0,0 +1,105 @@
...
+
+	Value {
+		PlaybackVolume "Line Out Playback Volume"
+		PlaybackSwitch "Line Out Playback Switch"
+		PlaybackChannels 2
+		PlaybackPriority 300
+		PlaybackPCM "hw:${CardId},0"
+	}
+}
+SectionDevice."Earpiece" {
```
perexg, line 30: Missing new-line.

perexg (changes requested):

perexg (changes requested):
SectionDefaults should be in PinePhone.conf (like BootSequence). It is shared by all profiles (and used when profile changes).

philmmanjaro:
For whatever reason with this new config there is no sound at all on the Pinephone when using Manjaro.

philmmanjaro:
Hmm, interesting @ollieparanoid and @tomfitzhenry. There is something missing when used on bleeding edge distros. Issue with [PMOS](https://gitlab.com/postmarketOS/pmaports/-/issues/2115) and with [Manjaro](https://gitlab.manjaro.org/manjaro-arm/issues/pinephone/phosh/-/issues/282). What is the best way to debug this?

nullr0ute:
> Hmm, interesting @ollieparanoid and @tomfitzhenry. There is something missing when used on bleeding edge distros. Issue with [PMOS](https://gitlab.com/postmarketOS/pmaports/-/issues/2115) and with [Manjaro](https://gitlab.manjaro.org/manjaro-arm/issues/pinephone/phosh/-/issues/282). What is the best way to debug this?

I wonder if there's a difference between their kernel forks and pure upstream linus kernel.

philmmanjaro:
Dowstream distros starting to revert these changes to known working ones. Example: https://raw.githubusercontent.com/dreemurrs-embedded/Pine64-Arch/master/PKGBUILDS/pine64/alsa-ucm-conf/0001-Revert-to-Mobian-UCM.patch

a-wai:
Just tested on a fresh Mobian install with both PulseAudio (v16.1) and PipeWire (v0.3.71) + WirePlumber (v0.4.14) and everything seems alright, except for the earpiece: anything playing on the right channel will be dropped, only the left channel actually plays back audio; I suspect that might cause issues with the modem audio, so I've created #324 to address this.

tomfitzhenry:
> @tomfitzhenry over a year overdue, my deepest apologies for not replying earlier (and letting the notification mail slip to the bottom of my inbox...)
> 
> Anyway, yes, the plan is to drop any obsolete downstream patches... I have yet to work with the ALSA team to refresh mobile-related patches, but it's definitely on my TODO list!

No worries!

Thank you for looking into this, and confirming things are working.

I looked more into why this PR initially had reports of issues, and have found that there is now advice to wipe the ALSA state upon discovering any issues: https://github.com/alsa-project/alsa-ucm-conf/commit/c1920e95d476c33f4b43484f14c36149829e514 

This is needed, because some UCM directives are only executed once, and then persisted across reboots (in the ALSA state). Thus, when there's a change in UCM config, the state can become inconsistent with the config, E.g. not setting now-required settings.

This also explains why my testing showed success, but some users reported issues. Pulse audio vs pipewire was likely a red herring.

So the tl;dr for distro maintainers is to get users to wipe their ALSA state if they report issues, especially if there's a large change to the relevant UCM.

PR #143, opened 2022-03-29
USB-Audio: Add profile for MSI MPG X570S Carbon Max Wifi.

lordhoto:
This patch depends on Linux 5.17.

This board features an Realtek ALC4080 chip. The various inputs and
outputs are distributed over multiple hardware devices:
* Input
 - hw:$card,0 loop back from speaker (called "Analog In")
 - hw:$card,1 line in at back panel
 - hw:$card,2 microphone from front AND back panel
* Output
 - hw:$card,0 speaker output at back panel (up to 7.1)
 - hw:$card,1 headerphone output at front panel
 - hw:$card,2 could not figure this one out
 - hw:$card,3 guessed S/PDIF output (sadly no way to test)

By default Mic and Line in are disabled. This confuses applications,
thus I made the profile enable them by default.

Without an UCM profile PulseAudio is not able to create output on
the front panel nor record any microphone input.

Signed-off-by: Johannes Schickel <lordhoto@gmail.com>

### DIFF
--- a/ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
+++ b/ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
@@ -1,3 +1,14 @@
+Define.SPDIFIndex "3"
+
+If.asus-rog-usb {
+	Condition {
+		Type RegexMatch
+		String "${CardComponents}"
+		Regex "USB(0b05:1996)"
+	}
+	True.Define.SPDIFIndex "2"
+}
+
 SectionDevice."Speaker" {
 	Comment "Speakers"
 	Value {
@@ -23,7 +34,7 @@ SectionDevice."SPDIF" {
 	Comment "S/PDIF Out"
 	Value {
 		PlaybackPriority 100
-		PlaybackPCM "hw:${CardId},3"
+		PlaybackPCM "hw:${CardId},${var:SPDIFIndex}"
 		PlaybackMixerElem "IEC958"
 	}
 }
@@ -47,10 +58,19 @@ SectionDevice."Mic" {
 		cset "name='Mic Capture Switch' on"
 	]
 
-	Comment "Microphone"
+	Comment "Front Microphone"
 	Value {
 		CapturePriority 200
 		CapturePCM "hw:${CardId},2"
+		JackControl "Mic - Input Jack"
+		CaptureMixerElem "Mic"
+	}
+
+	# On ASUS ROG Maximus XIII and others, back microphone
+	Comment "Microphone"
+	Value {
+		CapturePriority 300
+		CapturePCM "hw:${CardId}"
 		JackControl "Mic - Input Jack"
 		CaptureMixerElem "Mic"
 	}
--- a/ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
+++ b/ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
@@ -1,4 +1,5 @@
 Define.SPDIFIndex "3"
+Define.Mic1Name "Microphone"
 
 If.asus-rog-usb {
 	Condition {
@@ -6,7 +7,21 @@ If.asus-rog-usb {
 		String "${CardComponents}"
 		Regex "USB(0b05:1996)"
 	}
-	True.Define.SPDIFIndex "2"
+	True {
+		Define.SPDIFIndex "2"
+		Define.Mic1Name "Front Microphone"
+
+		SectionDevice."Mic2" {
+			# On ASUS ROG Maximus XIII and others, back microphone
+			Comment "Microphone"
+			Value {
+				CapturePriority 300
+				CapturePCM "hw:${CardId}"
+				JackControl "Mic - Input Jack"
+				CaptureMixerElem "Mic"
+			}
+		}
+	}
 }
 
 SectionDevice."Speaker" {
@@ -53,24 +68,15 @@ SectionDevice."Line" {
 	}
 }
 
-SectionDevice."Mic" {
+SectionDevice."Mic1" {
 	EnableSequence [
 		cset "name='Mic Capture Switch' on"
 	]
 
-	Comment "Front Microphone"
+	Comment "${var:Mic1Name}"
 	Value {
 		CapturePriority 200
 		CapturePCM "hw:${CardId},2"
-		JackControl "Mic - Input Jack"
-		CaptureMixerElem "Mic"
-	}
-
-	# On ASUS ROG Maximus XIII and others, back microphone
-	Comment "Microphone"
-	Value {
-		CapturePriority 300
-		CapturePCM "hw:${CardId}"
 		JackControl "Mic - Input Jack"
 		CaptureMixerElem "Mic"
 	}
--- a/ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
+++ b/ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
@@ -1,0 +1,57 @@
+SectionDevice."Speaker" {
+	Comment "Speakers"
+	Value {
+		PlaybackChannels 8
+		PlaybackPriority 200
+		PlaybackPCM "hw:${CardId}"
+		JackControl "Speaker - Output Jack"
+		PlaybackMixerElem "Speaker"
+	}
+}
+
+SectionDevice."Headphones" {
+	Comment "Front Headphones"
+	Value {
+		PlaybackPriority 300
+		PlaybackPCM "hw:${CardId},1"
+		JackControl "Headphone - Output Jack"
+		PlaybackMixerElem "Front Headphone"
+	}
+}
+
+SectionDevice."SPDIF" {
+	Comment "S/PDIF Out"
+	Value {
+		PlaybackPriority 100
+		PlaybackPCM "hw:${CardId},3"
+		PlaybackMixerElem "IEC958"
+	}
+}
+
+SectionDevice."Line" {
+	EnableSequence [
+		cset "name='Line Capture Switch' on"
+	]
+
+	Comment "Line In"
+	Value {
+		CapturePriority 100
+		CapturePCM "hw:${CardId},1"
+		JackControl "Line - Input Jack"
+		CaptureMixerElem "Line"
+	}
+}
+
+SectionDevice."Mic" {
+	EnableSequence [
+		cset "name='Mic Capture Switch' on"
+	]
+
+	Comment "Microphone"
+	Value {
+		CapturePriority 200
+		CapturePCM "hw:${CardId},2"
+		JackControl "Mic - Input Jack"
+		CaptureMixerElem "Mic"
+	}
+}
--- a/ucm2/USB-Audio/Realtek/ALC4080.conf
+++ b/ucm2/USB-Audio/Realtek/ALC4080.conf
@@ -1,0 +1,5 @@
+Comment "USB-audio on Realtek ALC4080"
+SectionUseCase."HiFi" {
+	File "/USB-Audio/Realtek/ALC4080-HiFi.conf"
+	Comment "Default Alsa Profile"
+}
--- a/ucm2/USB-Audio/Realtek/ALC4080.conf
+++ b/ucm2/USB-Audio/Realtek/ALC4080.conf
@@ -1,5 +1,5 @@
 Comment "USB-audio on Realtek ALC4080"
 SectionUseCase."HiFi" {
 	File "/USB-Audio/Realtek/ALC4080-HiFi.conf"
-	Comment "Default Alsa Profile"
+	Comment "Play HiFi quality Music"
 }
--- a/ucm2/USB-Audio/USB-Audio.conf
+++ b/ucm2/USB-Audio/USB-Audio.conf
@@ -28,6 +28,17 @@ If.realtek-alc1220-vb {
 	True.Define.ProfileName "Realtek/ALC1220-VB-Desktop"
 }
 
+If.realtek-alc4080 {
+
+	Condition {
+		Type RegexMatch
+		String "${CardComponents}"
+		# 0db0:419c MSI MPG X570S Carbon Max Wifi
+		Regex "USB(0db0:419c)"
+	}
+	True.Define.ProfileName "Realtek/ALC4080"
+}
+
 If.gigabyte-aorus-main {
 	Condition {
 		Type String
--- a/ucm2/USB-Audio/USB-Audio.conf
+++ b/ucm2/USB-Audio/USB-Audio.conf
@@ -33,8 +33,10 @@ If.realtek-alc4080 {
 	Condition {
 		Type RegexMatch
 		String "${CardComponents}"
+		# 0b05:1996 ASUS on multiple boards (including ASUS ROG Maximus XIII)
 		# 0db0:419c MSI MPG X570S Carbon Max Wifi
-		Regex "USB(0db0:419c)"
+		# 0db0:a073 MSI MAG X570S Torpedo Max
+		Regex "USB((0b05:1996)|(0db0:419c)|(0db0:a073))"
 	}
 	True.Define.ProfileName "Realtek/ALC4080"
 }
--- a/ucm2/USB-Audio/USB-Audio.conf
+++ b/ucm2/USB-Audio/USB-Audio.conf
@@ -34,9 +34,10 @@ If.realtek-alc4080 {
 		Type RegexMatch
 		String "${CardComponents}"
 		# 0b05:1996 ASUS on multiple boards (including ASUS ROG Maximus XIII)
+		# 0db0:1feb MSI Edge Wifi Z690
 		# 0db0:419c MSI MPG X570S Carbon Max Wifi
 		# 0db0:a073 MSI MAG X570S Torpedo Max
-		Regex "USB((0b05:1996)|(0db0:419c)|(0db0:a073))"
+		Regex "USB((0b05:1996)|(0db0:1feb)|(0db0:419c)|(0db0:a073))"
 	}
 	True.Define.ProfileName "Realtek/ALC4080"
 }


ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
```diff
@@ -0,0 +1,57 @@
...
+		PlaybackMixerElem "Front Headphone"
+	}
+}
+
+SectionDevice."SPDIF" {
+	Comment "S/PDIF Out"
+	Value {
+		PlaybackPriority 100
+		PlaybackPCM "hw:${CardId},3"
+		PlaybackMixerElem "IEC958"
```
Tatsh, line 27: This should be `hw:${CardId},2`.

ucm2/USB-Audio/USB-Audio.conf
```diff
@@ -28,6 +28,17 @@ If.realtek-alc1220-vb {
 	True.Define.ProfileName "Realtek/ALC1220-VB-Desktop"
 }
 
+If.realtek-alc4080 {
+
+	Condition {
+		Type RegexMatch
+		String "${CardComponents}"
+		# 0db0:419c MSI MPG X570S Carbon Max Wifi
+		Regex "USB(0db0:419c)"
```
Tatsh, line 37: Please make it `USB((0db0:419c)|(0b05:1996))` to match ASUS ROG Maximus XIII

ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
```diff
@@ -0,0 +1,57 @@
...
+
+SectionDevice."Line" {
+	EnableSequence [
+		cset "name='Line Capture Switch' on"
+	]
+
+	Comment "Line In"
+	Value {
+		CapturePriority 100
+		CapturePCM "hw:${CardId},1"
```
Tatsh, line 39: Detects plug in, but have not yet got any sound in. Will try with a better device soon.

ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
```diff
@@ -0,0 +1,57 @@
...
+
+SectionDevice."Mic" {
+	EnableSequence [
+		cset "name='Mic Capture Switch' on"
+	]
+
+	Comment "Microphone"
+	Value {
+		CapturePriority 200
+		CapturePCM "hw:${CardId},2"
```
Tatsh, line 53: This works but only on the front connector. Back connector gets no input. Also it seems (with a typical lavalier microphone) to be low volume. Talking at normal recording volume reaches about -30 dBm and I am expecting to get closer to -12. Maybe not a UCM issue.

ucm2/USB-Audio/Realtek/ALC4080.conf
```diff
@@ -0,0 +1,5 @@
+Comment "USB-audio on Realtek ALC4080"
+SectionUseCase."HiFi" {
+	File "/USB-Audio/Realtek/ALC4080-HiFi.conf"
+	Comment "Default Alsa Profile"
```
Tatsh, line 4: Maybe should write *Play HiFi quality music* like the other profile.

ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
```diff
@@ -0,0 +1,57 @@
...
+		PlaybackChannels 8
+		PlaybackPriority 200
+		PlaybackPCM "hw:${CardId}"
+		JackControl "Speaker - Output Jack"
+		PlaybackMixerElem "Speaker"
+	}
+}
+
+SectionDevice."Headphones" {
+	Comment "Front Headphones"
```
Tatsh, line 13: Need to test.

ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
mavoga: On MSI MAG X570S Torpedo Max: `hw:${CardId},3` outputs to S/PDIF (tested)

ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
mavoga: Works on MSI MAG X570S Torpedo Max (tested)

ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
mavoga: Front Mic works on MSI MAG X570S Torpedo Max (tested), rear connector not tested; level somewhat low

mavoga (commented):
Fixes most issues on MSI MAG X570S Torpedo Max

ucm2/USB-Audio/USB-Audio.conf
pinkflames: ALC4080 with the ID `0b05:1996` is used by other ASUS motherboards as well, so it's not specific to just Maximus XIII.

pinkflames:
Tried the profile with my ASUS motherboard and for some reason the only profile that I could select in plasma-pa wsa the Off profile i.e. no audio and not even the option to use Pro Audio. I'll try to get help for that but I do not even know where that would be.

Tatsh:
> Tried the profile with my ASUS motherboard and for some reason the only profile that I could select in plasma-pa wsa the Off profile i.e. no audio and not even the option to use Pro Audio. I'll try to get help for that but I do not even know where that would be.

Try plugging in headphones or mic. It seems jack sense is messing with keeping this profile enabled. If I unplug one like the mic even if others remain plugged in, sometimes it falls back to 'Dummy audio'. Not sure how to fix this yet.

pinkflames:
I have headphones plugged into the front jack at all times. Not sure if that's related but frankly I don't care very much because I'd rather have my remapping done in PW/WP and avoid having to deal with the surprising behaviors of UCM (e.g. I tried to create my own profile based on the related ALC1220 sound controller and with it headphone output was only available for Stereo profile despite being independent subdevice). 

Additionally I want to use the 2.1 profile for which a virtual sink is required at all times due to the underlying hardware only enabling LFE output if FC is also enabled which perhaps is something that could be fixed in UCM but I already resolved it at PW level in much less time despite encountering two upstream issues.

perexg:
If I may politely ask you, please, report all issues with UCM to be resolved.

ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
lordhoto: I'm not sure what to do here. Maybe the ASUS board has a different layout? Which board did you try this on @Tatsh?

ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
lordhoto: Level is indeed a bit low. The volume levels on Windows are noticeably higher. I guess Windows seems to apply some boost (even without enabling it in the settings).

Back microphone connector definitely works on hw:${CardId},2 on my board. Front and back levels are between -42dBm to -36dBm during normal talking for me.

ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
Tatsh: It's definitely 2 on ASUS ROG Maximus XIII. What you could do is have this be conditional based on `RegexMatch` (not tested below).

```
Define SPDIFIndex "3"

If.asus-rog-maximus-xii {
    Condition {
      Type RegexMatch
       String "${CardComponents}"
      Regex "USB(0b05:1996)"
    }
    True.SPDIFIndex "2"
}
```

Then refer to `${SPDIFIndex}` in `PlaybackPCM`.

ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
Tatsh: I think you can use `cset` to set the level like here: https://github.com/alsa-project/alsa-ucm-conf/blob/master/ucm2/Qualcomm/apq8096/HiFi.conf#L18= If this value is dBm, then maybe it should be about 20?

ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
Tatsh: Fixed syntax:

```
Define.SPDIFIndex "3"

If.asus-rog-usb {
        Condition {
                Type RegexMatch
                String "${CardComponents}"
                Regex "USB(0b05:1996)"
        }
        True.Define.SPDIFIndex "2"
}

...

SectionDevice."SPDIF" {
        Comment "S/PDIF Out"
        Value {
                PlaybackPriority 100
                PlaybackPCM "hw:${CardId},${var:SPDIFIndex}"
                PlaybackMixerElem "IEC958"
        }
}
```

ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
Tatsh: I tried setting the volume this way and it made no difference.

ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
Tatsh: Tested and it works.

Tatsh:
@lordhoto see https://github.com/lordhoto/alsa-ucm-conf/pull/1

ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
Tatsh: Line-in works well enough. I tested with my phone connected though I had to turn up the volume a bit.

It may be a good thing that there is not a boost like Windows. This allows more headroom. OBS can boost the source live (Audacity cannot). When I use a lavalier mic in normal position and boost with OBS to 20 dBm, it sounds good enough and there is not much noise.

ucm2/USB-Audio/USB-Audio.conf
tekstryder: On the MSI Edge Wifi Z690 motherboard the ALC4080 is: `0db0:1feb`

lordhoto: @mavoga @tekstryder btw. for proper Control Names I supplied a patch for my board to the Linux kernel: https://github.com/torvalds/linux/commit/5762f980ca10dcfe5eead7c40d1c34cae61f409b

Does Speaker/Front Headphone Volume control work for you? Maybe you will need to use the same map for the IDs of your boards.

mavoga: @lordhoto I patched `sound/usb/mixer_maps.c` to apply the same map to the USB id of my card too, therefore the volume controls work perfectly (currently trying to figure out how to correctly send a kernel patch since I never did it before...).

lordhoto:
@Tatsh I merged your changes and fixed the microphone on my MSI board again afterwards. I hope it will still work for your ASUS board.

I installed a sound firmware update for my board. Now, for some reason, the back panel microphone doesn't work at all on Linux. All the other features seem to still work fine though.

XDmToter:
I have an "ASUS ROG STRIX Z590-I GAMING WIFI" with the same issue.  Microphone input does not work unless I add a custom config to Pipewire.  Sometimes I have to use "hw:0,2" while other times I have to use "hw:2,2".

I wanted to post this here since I didn't see anyone else mention `0b05:1998` in the proposed RegEx.

`lsusb`:
```
Bus 001 Device 003: ID 0b05:1998 ASUSTek Computer, Inc. USB Audio
```

`arecord -l`:
```
**** List of CAPTURE Hardware Devices ****
card 2: Audio [USB Audio], device 0: USB Audio [USB Audio]
  Subdevices: 0/1
  Subdevice #0: subdevice #0
card 2: Audio [USB Audio], device 1: USB Audio [USB Audio #1]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
card 2: Audio [USB Audio], device 2: USB Audio [USB Audio #2]
  Subdevices: 1/1
  Subdevice #0: subdevice #0
```

Tatsh:
What is the command or way to add a config to Pipewire to make it work?

pinkflames:
The method I used with PipeWire is to pick the `Pro Audio` profile. Afterwards it depends on what you want. The front headphone jack should be available as the `USB Audio Pro 1` while the digital outputs will be available as the device with the number 2 or 3. There will also be 3 input devices but I have not verified what they connect to or if they function as intended.

The only drawback is the main 7.1 output which will work but lacks a proper channel map in `Pro Audio` profile (as it always is by design for `Pro Audio`), so if you need more than stereo out of it, you need to add a virtual sink following upstream wiki on adding virtual sinks via the loopback module and then you can define the desired channel map for virtual sink.

Tatsh:
I think we need to have other profiles besides just the Play HiFi. We should break it down to Analog Stereo Duplex, Analog Stereo Output, IEC958 Input, etc. The reason is because if we want to support IEC958 bitstream (AC3, etc) pass-through, we need to remove Pulse/Pipewire's control of the device. Unless someone knows how to make it work with Pulse/Pipewire working. Right now in `pavucontrol`, there is no option to pass-through AC3 or others on the S/PDIF device (unlike HDMI on my GPU).

lordhoto:
I am not sure we even can define different profiles for Stereo Duplex etc.. According to the documentation the uses cases one can define are:
```
#define SND_USE_CASE_VERB_INACTIVE              "Inactive"              /**< Inactive Verb */
#define SND_USE_CASE_VERB_HIFI                  "HiFi"                  /**< HiFi Verb */
#define SND_USE_CASE_VERB_HIFI_LOW_POWER        "HiFi Low Power"        /**< HiFi Low Power Verb */
#define SND_USE_CASE_VERB_VOICE                 "Voice"                 /**< Voice Verb */
#define SND_USE_CASE_VERB_VOICE_LOW_POWER       "Voice Low Power"       /**< Voice Low Power Verb */
#define SND_USE_CASE_VERB_VOICECALL             "Voice Call"            /**< Voice Call Verb */
#define SND_USE_CASE_VERB_IP_VOICECALL          "Voice Call IP"         /**< Voice Call IP Verb */
#define SND_USE_CASE_VERB_ANALOG_RADIO          "FM Analog Radio"       /**< FM Analog Radio Verb */
#define SND_USE_CASE_VERB_DIGITAL_RADIO         "FM Digital Radio"      /**< FM Digital Radio Verb */
```
Taken from: https://git.alsa-project.org/?p=alsa-lib.git;a=blob;f=include/use-case.h#l100

These speaker configurations don't really seem related to the profiles. If at all, they would probably be defined inside a profile. But I couldn't find any way to do that. Most likely ALSA UCM is not designed for defining such profiles.

Tatsh:
Yes I have no idea how it is done to get those other profiles. I just know they existed on my previous HDA-based computer. I presume much of it has to be done in kernel. Out of scope for here.

pinkflames:
The profiles for your old device were almost certainly generated by the ACP rather than UCM2 system. Unfortunately I have no idea how ACP is supposed to map on to UCM2 but apparently UCM2 is the future, so one would expect it to be possible to re-create ACP configuration in UCM2.

Tatsh:
Any idea why it says *(unavailable)*?

image

ucm2/USB-Audio/USB-Audio.conf
AWJoki: Hi. For msi mag Z690 TomaHawk wifi ddr4 `0db0:b202`

AWJoki:
Hi. Everybody.. I have tried and Read all here up and down but one thing i dont understand how do i know if the profile Config is loaded or not.. and A other Question do PipeWire map after this??

nathanfranke (approved):
Copied the changes to `/usr/share/alsa` and my rear and front audio works as expected.

- Does the comment have to be "Play HiFi quality Music"? Why not just "Realtek ALC4080"?
- For some reason, the microphone stops working if I first plug it in rear, then front, then back into rear. It fixes itself after a restart.

My motherboard is MSI `MAG-X570S-TOMAHAWK-MAX-WIFI`

Tatsh:
> Does the comment have to be "Play HiFi quality Music"? Why not just "Realtek ALC4080"?

A lot of other profiles use this name. We should stick with what the project is doing. Better chance of this getting merged.

nathanfranke:
Should there be an option to only use Stereo? My headphones are only stereo but since the profile is set to surround some applications are trying to play audio through different speakers that aren't plugged in.

lordhoto:
> Should there be an option to only use Stereo? My headphones are only stereo but since the profile is set to surround some applications are trying to play audio through different speakers that aren't plugged in.

I guess there should. I could not find any information on how to define a profile to be able to select various speaker configurations though.  The VB1220 profile also does not offer this option.

> For some reason, the microphone stops working if I first plug it in rear, then front, then back into rear. It fixes itself after a restart.

Might be a firmware bug. After upgrading the firmware I also had strange issues with back microphone.

> Any idea why it says (unavailable)?

I have never seen that to be honest. I can't reproduce it myself.

perexg:
> > Should there be an option to only use Stereo? My headphones are only stereo but since the profile is set to surround some applications are trying to play audio through different speakers that aren't plugged in.
> 
> I guess there should. I could not find any information on how to define a profile to be able to select various speaker configurations though. The VB1220 profile also does not offer this option.

I just added the `Variant` block to the latest alsa-lib: https://github.com/alsa-project/alsa-lib/commit/c3a5d32ac37b0a09c77e98cd3b1be888bb0141e4 . It will be available since 1.2.7 release (soon).

perexg:
I applied all patches to the main repo. Thank you for your work.

Tatsh:
> Any idea why it says _(unavailable)_?
> 
> image

I figured out why but I am not sure how to fix it if it involves the profile. The code in pavucontrol reads the `available` field from Pulse:

```c++
        if (!profileIt->available)
            desc += _(" (unavailable)");
```

According to Pulse:

```c
typedef struct pa_card_profile_info2 {
// ...
    int available;
    /**< Is this profile available? If this is zero, meaning "unavailable",
     * then it makes no sense to try to activate this profile. If this is
     * non-zero, it's still not a guarantee that activating the profile will
     * result in anything useful, it just means that the server isn't aware of
     * any reason why the profile would definitely be useless. \since 5.0 */
} pa_card_profile_info2;
```

Is something in the profile able to change this value to 1?

Tatsh:
This flag gets set to non-zero when I plug in a mic. I think the profile name should be updated too: "Play and record HiFi quality Music". There should be a profile for only playback to avoid my issue with unavailable so I don't have to keep a mic plugged in.

ucm2/USB-Audio/USB-Audio.conf
yhojann-cl: Hi, for MAG X570S TOMAHAWK MAX WIFI is  `0db0:151f`

ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf
tekstryder: @lordhoto SPDIFIndex is 2 on the MSI MPG Z690 EDGE(0db0:1feb) as well.

Do I need to open a new issue for this change?

lordhoto: @tekstryder sorry for the late reply: yes, you'll need to open a new issue/pull request for this change. There is pull request has been closed already.

ucm2/USB-Audio/USB-Audio.conf
AndreyYashkin: > btw. for proper Control Names I supplied a patch for my board to the Linux kernel:

@lordhoto I am trying to patch the profile for MSI MPG Z690 FORCE WIFI, but simply adding USB Audio ID from `lsusb` output does not seem to enable front headphone and mic. How the kernel patch is related to this patch? Do I understand correctly that without applying the similar kernel patch for each motherboard listed in "If.realtek-alc4080" the config is not expected to work?

yhojann-cl:
Please, add `151f` id for https://github.com/alsa-project/alsa-ucm-conf/issues/172

yhojann-cl:
This changes does not works on MPG X570S EDGE MAX WIFI.

schmorp:
As an addition, on the MSI MPG Z690 Force/Carbon WIFI, the USB ID for output is 0db0:005a and SPDIF is on index 2.

Chais:
The MSI MPG B650I Edge Wifi also uses the ALC4080. ID `0db0:36e7`. Currently stereo output works fine, microphone input is detected as plugged in, but doesn't actually work as an input. Seems to work as an output instead.

PR #144, opened 2022-03-29
Fix support for sof-essx8336

mchehab:
There are a couple of issues at the support for sof-essx8336, causing it to fail working.

This PR fix them. Tested on Huawei Matebook D15. Before the fix, the device was appearing as "sysdefault:0" instead of "hw:0", and there were several other problems.

 After the fix:
- the device is now identified as hw:0;
- calling alsamixer without `-Dsysdefault:0` (or `-Dhw:0`) now shows a mixer (instead of showing nothing);
- The HDMI volume controls now appear;
- Using `alsaucm load` now works fine.

There are still a couple of things to be fixed, but I suspect that at least some of them would require firmware (or driver?) changes:
- the internal microphone is not working yet;
- the headphone/headset doesn't work.

### DIFF
--- a/ucm2/Intel/sof-essx8336/HiFi.conf
+++ b/ucm2/Intel/sof-essx8336/HiFi.conf
@@ -21,18 +21,21 @@ If.amic {
 			"Headset"
 		]
 
-		Value {
-			CapturePriority 100
-			CapturePCM "hw:${CardId}"
-		}
-
 		EnableSequence [
+			cset "name='Differential Mux' lin1-rin1"
 			cset "name='Internal Mic Switch' on"
 		]
 
 		DisableSequence [
 			cset "name='Internal Mic Switch' off"
 		]
+
+		Value {
+			CapturePriority 100
+			CapturePCM "hw:${CardId}"
+			CaptureMixerElem "ADC PGA Gain"
+			CaptureMasterElem "ADC"
+		}
 	}
 }
 
@@ -65,7 +68,7 @@ If.dmic {
 }
 
 SectionDevice."Speaker" {
-	Comment "Speaker"
+	Comment "Speakers"
 
 	ConflictingDevice [
 		"Headphones"
@@ -81,10 +84,10 @@ SectionDevice."Speaker" {
 	Value {
 		PlaybackPriority 100
 		PlaybackPCM "hw:${CardId}"
-		PlaybackMixerElem "Speaker"
-		PlaybackMasterElem "Master"
-		PlaybackVolume "Speaker Volume"
-		PlaybackSwitch "Speaker Switch"
+		# The es8316 only has a HP-amp which is muxed to the speaker
+		# or to the headpones output
+		PlaybackMixerElem "Headphone Mixer"
+		PlaybackMasterElem "DAC"
 	}
 }
 
@@ -95,21 +98,11 @@ SectionDevice."Headphones" {
 		"Speaker"
 	]
 
-	EnableSequence [
-		cset "name='Headphone Switch' on"
-	]
-
-	DisableSequence [
-		cset "name='Headphone Switch' off"
-	]
-
 	Value {
 		PlaybackPriority 300
 		PlaybackPCM "hw:${CardId}"
-		PlaybackMixerElem "Headphone"
-		PlaybackMasterElem "Master"
-		PlaybackVolume "Headphone Volume"
-		PlaybackSwitch "Headphone Switch"
+		PlaybackMixerElem "Headphone Mixer"
+		PlaybackMasterElem "DAC"
 		JackControl "Headphone Jack"
 		JackHWMute "Speaker"
 	}
@@ -129,16 +122,19 @@ SectionDevice."Headset" {
 	}
 
 	EnableSequence [
-		cset "name='Headset Mic Switch' on"
+		cset "name='Headset Switch' on"
+		cset "name='Digital Mic Mux' 'dmic disable'"
 	]
 
 	DisableSequence [
-		cset "name='Headset Mic Switch' off"
+		cset "name='Headset Switch' on"
 	]
 
 	Value {
 		CapturePriority 300
 		CapturePCM "hw:${CardId}"
+		CaptureMixerElem "ADC PGA Gain"
+		CaptureMasterElem "ADC"
 		JackControl "Headset Mic Jack"
 	}
 }
--- a/ucm2/Intel/sof-essx8336/HiFi.conf
+++ b/ucm2/Intel/sof-essx8336/HiFi.conf
@@ -1,4 +1,13 @@
 SectionVerb {
+	EnableSequence [
+		# Disable all inputs / outputs
+		cset "name='Speaker Switch' off"
+		cset "name='Headphone Switch' off"
+		cset "name='Headset Mic Switch' off"
+		cset "name='Internal Mic Switch' off"
+		cset "name='DAC Mono Mix Switch' off"
+	]
+}
 
 If.amic {
 	Condition {
--- a/ucm2/Intel/sof-essx8336/sof-essx8336.conf
+++ b/ucm2/Intel/sof-essx8336/sof-essx8336.conf
@@ -1,5 +1,29 @@
 Syntax 4
 
+BootSequence [
+	# Setup muxes / switches
+	cset "name='Left Headphone Mixer Left DAC Switch' on"
+	cset "name='Right Headphone Mixer Right DAC Switch' on"
+	# Set digital mix mux to "dmic disable"
+	# That doesn't affect dmic, but other values mute headset mic
+	cset "name='Digital Mic Mux' 0"
+
+	# Set HP vol to 0 dB
+	cset "name='Headphone Playback Volume' 100%"
+	cset "name='Headphone Mixer Volume' 100%"
+
+	# Set DAC vol
+	cset "name='DAC Playback Volume' 60%"
+	# LDATA TO LDAC, RDATA TO RDAC
+	cset "name='DAC Source Mux' 0"
+
+	# Disable Auto Level Control
+	cset "name='ALC Capture Switch' off"
+
+	# Set capture vol
+	cset "name='ADC Capture Volume' 60%"
+]
+
 Define.DeviceDmic ""
 
 If.devdmic {
--- a/ucm2/Intel/sof-essx8336/sof-essx8336.conf
+++ b/ucm2/Intel/sof-essx8336/sof-essx8336.conf
@@ -12,7 +12,7 @@ If.devdmic {
 }
 
 SectionUseCase."HiFi" {
-	File "/Intel/sof-es8336/HiFi.conf"
+	File "/Intel/sof-essx8336/HiFi.conf"
 	Comment "Play and record HiFi quality Music"
 }
 
--- a/ucm2/Intel/sof-essx8336/sof-essx8336.conf
+++ b/ucm2/Intel/sof-essx8336/sof-essx8336.conf
@@ -39,28 +39,6 @@ SectionUseCase."HiFi" {
 	File "/Intel/sof-essx8336/HiFi.conf"
 	Comment "Play and record HiFi quality Music"
 }
-
-# the kcontrols initial values, which will be set by `alsactl init`
-If.speaker {
-	Condition {
-		Type ControlExists
-		Control "DAC Playback Volume"
-	}
-	True.BootSequence [
-		cset "name='DAC Playback Volume' 60%"
-	]
-}
-
-If.headphone {
-	Condition {
-		Type ControlExists
-		Control "name='Headphone Playback Volume'"
-	}
-	True.BootSequence [
-		cset "name='Headphone Playback Volume' 60%"
-	]
-}
-
 If.dmic {
 	Condition {
 		Type String
@@ -77,12 +55,3 @@ If.dmic {
 	}
 }
 
-If.Capture {
-	Condition {
-		Type ControlExists
-		Control "name='ADC Capture Volume'"
-	}
-	True.BootSequence [
-		cset "name='ADC Capture Volume' 60%"
-	]
-}


ucm2/Intel/sof-essx8336/sof-essx8336.conf
```diff
@@ -20,10 +20,11 @@ SectionUseCase."HiFi" {
 If.speaker {
 	Condition {
 		Type ControlExists
-		Control "DAC Playback Volume"
+		Control "name='DAC'"
 	}
 	True.BootSequence [
-		cset "name='DAC Playback Volume' 60%"
+		cset "name='DAC' 60%"
+		cset "name='DAC Source Mux' 0"
```
plbossart, line 27: @mchehab are you sure about this one?

in https://github.com/thesofproject/linux/issues/3248#issuecomment-1060133008 @yangxiaohua2009 suggested this control name?

mchehab: No. Please see the updated version. The name of the control is correct, and setting it on higher values than 60% may cause level saturation.
So, basically, what I did at the newest version is to place this setting at SectionVerb.EnableSequence at patch 1 
, and, since the same volumes are there for both headphone and speaker, I dropped those if's at patch 3, and fixed the control names on patch 4.

ucm2/Intel/sof-essx8336/sof-essx8336.conf
```diff
@@ -15,50 +15,3 @@ SectionUseCase."HiFi" {
...
-	False.If.Dmic0 {
-		Condition {
-			Type ControlExists
-			Control "name='Dmic0 Capture Volume'"
-		}
-		True.BootSequence [
-			cset "name='Dmic0 Capture Volume' 70%"
-		]
-	}
-}
```
plbossart, line 54: this is not related to ES8336 but Intel DMIC support. not sure why you removed it?

ucm2/Intel/sof-essx8336/HiFi.conf
```diff
@@ -1,4 +1,35 @@
...
+		# Set DAC vol
+		cset "name='DAC Playback Volume' 60%"
+		# LDATA TO LDAC, RDATA TO RDAC
+		cset "name='DAC Source Mux' 0"
+
+		# Disable Auto Level Control
+		cset "name='ALC Capture Switch' off"
+
+		# Set capture vol
+		cset "name='ADC Capture Volume' 60%"
```
plbossart, line 30: the settings between line17 and 30 should probably be part of the boot sequence so that they are set even when UCM is not used to select a device?

plbossart (commented):
minor improvements suggested below, thanks @mchehab

ucm2/Intel/sof-essx8336/sof-essx8336.conf
mchehab: no particular reason. I'll drop this hunk.

ucm2/Intel/sof-essx8336/HiFi.conf
mchehab: Ok, I'll move them to the boot sequence.

yangxiaohua2009:
There is current a bug in the codec driver that turn off the Internal Mic will mute the headset mic, whether the headset mic is on or off.
For switching between headset mic and Internal Mic, use `cset "name='Differential Mux' 'lin1-rin1'"` and
`cset "name='Differential Mux' 'lin2-rin2'"` 
The headset mic and Internal Mic are just lin1 and lin2, however we don't know the actual hardware relationship. So there is a 50% chance that the ucm works for you doesn't work for others.

mchehab:
> There is current a bug in the codec driver that turn off the Internal Mic will mute the headset mic, whether the headset mic is on or off. For switching between headset mic and Internal Mic, use `cset "name='Differential Mux' 'lin1-rin1'"` and `cset "name='Differential Mux' 'lin2-rin2'"` The headset mic and Internal Mic are just lin1 and lin2, however we don't know the actual hardware relationship. So there is a 50% chance that the ucm works for you doesn't work for others.

Hmm... I guess we need then to have a quirk to select between different maps, e. g.:

```
static const struct snd_soc_dapm_route sof_es8316_intmic_in1_map[] = {
	{"MIC1", NULL, "Internal Mic"},
	{"MIC2", NULL, "Headset Mic"},
};

static const struct snd_soc_dapm_route sof_es8316_intmic_in1_map_rev[] = {
	{"MIC2", NULL, "Internal Mic"},
	{"MIC1", NULL, "Headset Mic"},
};
```

And then change `sof_es8316_widgets` mic entries to something like:

```
	SND_SOC_DAPM_MIC("Headset Mic", sof_es8316_headset_mic_event),
	SND_SOC_DAPM_MIC("Internal Mic", sof_es8316_internal_mic_event),
```

Placing the logic to change the differential mux inside the headset/internal events.

I'll try to write a patch like that.

ucm2/Intel/sof-essx8336/HiFi.conf
```diff
@@ -1,4 +1,37 @@
...
+		# Disable all inputs / outputs
+		cset "name='Speaker Switch' off"
+		cset "name='Headphone Switch' off"
+		cset "name='Headset Mic Switch' off"
+		cset "name='Internal Mic Switch' off"
+		cset "name='DAC Mono Mix Switch' off"
+	]
+}
+
+BootSequence [
```
perexg, line 12: The boot sequence belongs to the upper file (sof-essx8336.conf).

yangxiaohua2009:
> > There is current a bug in the codec driver that turn off the Internal Mic will mute the headset mic, whether the headset mic is on or off. For switching between headset mic and Internal Mic, use `cset "name='Differential Mux' 'lin1-rin1'"` and `cset "name='Differential Mux' 'lin2-rin2'"` The headset mic and Internal Mic are just lin1 and lin2, however we don't know the actual hardware relationship. So there is a 50% chance that the ucm works for you doesn't work for others.
> 
> Hmm... I guess we need then to have a quirk to select between different maps, e. g.:
> 
> ```
> static const struct snd_soc_dapm_route sof_es8316_intmic_in1_map[] = {
> 	{"MIC1", NULL, "Internal Mic"},
> 	{"MIC2", NULL, "Headset Mic"},
> };
> 
> static const struct snd_soc_dapm_route sof_es8316_intmic_in1_map_rev[] = {
> 	{"MIC2", NULL, "Internal Mic"},
> 	{"MIC1", NULL, "Headset Mic"},
> };
> ```
> 
> And then change `sof_es8316_widgets` mic entries to something like:
> 
> ```
> 	SND_SOC_DAPM_MIC("Headset Mic", sof_es8316_headset_mic_event),
> 	SND_SOC_DAPM_MIC("Internal Mic", sof_es8316_internal_mic_event),
> ```
> 
> Placing the logic to change the differential mux inside the headset/internal events.
> 
> I'll try to write a patch like that.

Please refer to the code in
https://elixir.bootlin.com/linux/v5.18-rc1/source/sound/soc/intel/boards/bytcht_es8316.c

mchehab:
> > Placing the logic to change the differential mux inside the headset/internal events.
> > I'll try to write a patch like that.
> 
> Please refer to the code in https://elixir.bootlin.com/linux/v5.18-rc1/source/sound/soc/intel/boards/bytcht_es8316.c
I ended writing a code somewhat similar:

```
diff --git a/sound/soc/intel/boards/sof_es8336.c b/sound/soc/intel/boards/sof_es8336.c
index 2965c4c1cbea..e9bf516ae4c9 100644
--- a/sound/soc/intel/boards/sof_es8336.c
+++ b/sound/soc/intel/boards/sof_es8336.c
@@ -31,6 +31,7 @@
 #define SOF_ES8336_ENABLE_DMIC			BIT(5)
 #define SOF_ES8336_JD_INVERTED			BIT(6)
 #define SOF_ES8336_HEADPHONE_GPIO		BIT(7)
+#define SOC_ES8336_HEADSET_MIC1			BIT(8)
 
 static unsigned long quirk;
 
@@ -90,6 +91,8 @@ static void log_quirks(struct device *dev)
 		dev_info(dev, "quirk headphone GPIO enabled\n");
 	if (quirk & SOF_ES8336_JD_INVERTED)
 		dev_info(dev, "quirk JD inverted enabled\n");
+	if (quirk & SOC_ES8336_HEADSET_MIC1)
+		dev_info(dev, "quirk headset at mic1 port enabled\n");
 }
 
 static int sof_es8316_speaker_power_event(struct snd_soc_dapm_widget *w,
@@ -145,13 +148,23 @@ static const struct snd_soc_dapm_route sof_es8316_audio_map[] = {
 	{"Speaker", NULL, "HPOL"},
 	{"Speaker", NULL, "HPOR"},
 	{"Speaker", NULL, "Speaker Power"},
+
+	{"Differential Mux", "lin1-rin1", "MIC1"},
+	{"Differential Mux", "lin2-rin2", "MIC2"},
+	{"Differential Mux", "lin1-rin1 with 20db Boost", "MIC1"},
+	{"Differential Mux", "lin2-rin2 with 20db Boost", "MIC2"},
 };
 
-static const struct snd_soc_dapm_route sof_es8316_intmic_in1_map[] = {
+static const struct snd_soc_dapm_route sof_es8316_headset_mic2_map[] = {
 	{"MIC1", NULL, "Internal Mic"},
 	{"MIC2", NULL, "Headset Mic"},
 };
 
+static const struct snd_soc_dapm_route sof_es8316_headset_mic1_map[] = {
+	{"MIC2", NULL, "Internal Mic"},
+	{"MIC1", NULL, "Headset Mic"},
+};
+
 static const struct snd_soc_dapm_route dmic_map[] = {
 	/* digital mics */
 	{"DMic", NULL, "SoC DMIC"},
@@ -225,8 +238,13 @@ static int sof_es8316_init(struct snd_soc_pcm_runtime *runtime)
 
 	card->dapm.idle_bias_off = true;
 
-	custom_map = sof_es8316_intmic_in1_map;
-	num_routes = ARRAY_SIZE(sof_es8316_intmic_in1_map);
+	if (quirk & SOC_ES8336_HEADSET_MIC1) {
+		custom_map = sof_es8316_headset_mic1_map;
+		num_routes = ARRAY_SIZE(sof_es8316_headset_mic1_map);
+	} else {
+		custom_map = sof_es8316_headset_mic2_map;
+		num_routes = ARRAY_SIZE(sof_es8316_headset_mic2_map);
+	}
 
 	ret = snd_soc_dapm_add_routes(&card->dapm, custom_map, num_routes);
 	if (ret)
@@ -297,7 +315,8 @@ static const struct dmi_system_id sof_es8336_quirk_table[] = {
 			DMI_MATCH(DMI_SYS_VENDOR, "HUAWEI"),
 			DMI_MATCH(DMI_BOARD_NAME, "BOHB-WAX9-PCB-B2"),
 		},
-		.driver_data = (void *)(SOF_ES8336_HEADPHONE_GPIO)
+		.driver_data = (void *)(SOF_ES8336_HEADPHONE_GPIO |
+					SOC_ES8336_HEADSET_MIC1)
 	},
 	{}
 };

```

One thing that seems to be missing at the bytcht_es8316.c is the route map for the Differential Dmux.

ucm2/Intel/sof-essx8336/HiFi.conf
mchehab: Ok. moved it to sof-essx8336.conf.

plbossart (commented):
LGTM, thanks @mchehab

Neo-29:
I'm here  because like you I'm stuck with my teclast F15 plus 2.. With no sound but with hardware detected.
I copied the files in alsa folder...
No success.
I see however now the microphone in the gnome config but doesn't work ( as you mentionned)

I use last 5.17 kernel.
I tried with 5. 18rc 1... It's worse....

Dummy for output
Nothing in input

How can I help improve the work on that?

mchehab:
Added some changes at the headset patch, in order to properly use the <headset switch> that should now be working properly after the latest patches that I submitted today to alsa ML, which is also at:

https://github.com/mchehab/linux/tree/sound-es8336

mchehab:
> I'm here because like you I'm stuck with my teclast F15 plus 2.. With no sound but with hardware detected. I copied the files in alsa folder... No success. I see however now the microphone in the gnome config but doesn't work ( as you mentionned)
> 
> I use last 5.17 kernel. I tried with 5. 18rc 1... It's worse....
> 
> Dummy for output Nothing in input
> 
> How can I help improve the work on that?

There are some settings that the driver can't detect from the hardware itself, but requires a patch adding some quirks like this:

https://github.com/mchehab/linux/commit/d678aef9fc9a07aced611aa7cb8c9b800c649e5a

Those could be tried by creating a file like `/etc/modprobe.d/sound.conf` with:

```
options snd_soc_sof_essx8336 dyndbg=+p quirk=<value>
```
The relevant bits to play with the `<value>`  above is defined at `sound/soc/intel/boards/sof_es8336.c` as:
```
SOF_ES8336_SPEAKERS_EN_GPIO1_QUIRK	BIT(4)   // 0x010  (or 16) - use alternative pin to enable speaker
SOF_ES8336_JD_INVERTED			BIT(6)   // 0x040  (or 64) - inverts headphone detection logic
SOF_ES8336_HEADPHONE_GPIO		BIT(7)   // 0x080 (or 128) - use a separate pin to enable headphone
SOC_ES8336_HEADSET_MIC1			BIT(8)   // 0x100 (or 256) - change the input for the headset microphone
```


So, basically, for some device where speaker uses GPIO1 and the headset uses MIC1, one should use 0x110 or 16 + 256 = 272, e. g.

```
options snd_soc_sof_essx8336 dyndbg=+p quirk=0x110
```

or 

```
options snd_soc_sof_essx8336 dyndbg=+p quirk=272
```
Should enable both quirks. Once you get this right, a patch can be written to auto-detect your board.

Neo-29:
Tks for the answer
Tested both parameters ....
no sucesss

mchehab:
> Tks for the answer Tested both parameters .... no sucesss

did you test all possible combinations of 0x10, 0x40 and 0x80 flags (e. g. 0x10, 0x50, 0x90, 0xc0, 0xd0)?

Also, you could open `alsamixer -D hw:0` and try to mute/unmute switches and increase volumes there, as sometimes some of the mixers that affect the speaker are just on a too low position ;-)

plbossart:
> Tks for the answer Tested both parameters .... no sucesss

Just making sure you did reboot between your tests, the parameters are only taken on initial boot?

Neo-29:
Yes.. Reboot.... Nothing.
With the script I found somewhere here...
I just have a click.... No sound..
So it's a dead end here

Neo-29:
I didn't test all the combinations.. Just the ones you gave.
Too long to test all... This is really a nightmare

mchehab:
> Yes.. Reboot.... Nothing. With the script I found somewhere here... I just have a click.... No sound.. So it's a dead end here

It is a good sign if you heard a click. Maybe it is just the volumes that are too low. You may use `alsamixer -D hw:0` to see all the volumes exposed by the hardware. Try to increase volumes and unmute switches.

> I didn't test all the combinations.. Just the ones you gave. Too long to test all... This is really a nightmare

Yeah, it takes some time, but there aren't that many combinations.

fcsm1922:
Here's what I got.

1. When running the script in the /usr/share/alsa directory, after copying the ucm and ucm2 folders there
```
/usr/bin/install: 'ucm2/Intel/sof-essx8336/Hdmi.conf' and '/usr/share/alsa/ucm2/Intel/sof-essx8336/Hdmi.conf' are the same file
/usr/bin/install: 'ucm2/Intel/sof-essx8336/HiFi.conf' and '/usr/share/alsa/ucm2/Intel/sof-essx8336/HiFi.conf' are the same file
/usr/bin/install: 'ucm2/Intel/sof-essx8336/sof-essx8336.conf' and '/usr/share/alsa/ucm2/Intel/sof-essx8336/sof-essx8336.conf' are the same file
```

2. When adding the options line to the sound.conf file, which is located along the path /etc/modprobe.d/sound.conf
snd_soc_sof_essx8336 dyndbg=+p quirk=0x110 and reboot didn't change anything. there was no sound (((

3. Checked different versions of the flag combination (options snd_soc_sof_essx8336 dyndbg=+p quirk= 0x01,0x40,x0x80, 0x50, 0x90,
0xc0, 0xd0) After each new combination, the system rebooted. All options in lsamixer -D hw:0 are set
to the maximum.

4. My asla-info output
http://alsa-project.org/db/?f=e05f665fe4c6a9581eeb65fcb144079468cb37c3

5. no sound..

quintafeira:
Tried these files with the following quirk values (I don't know if all of them are valid, I just tried all combinations possible. There's also a possibility I missed some value, as there's only 15 in my list. I think it should be 16.):

0x010
0x040
0x080
0x100
0x050
0x090
0x110
0x0d0
0x150
0x1d0
0x0c0
0x140
0x1c0
0x180
0x190

With every value, if I boot the system with headphones connected, the headphones are recognized, but there's no sound. Then, I disconnect the headphones and connect again. I got sound, but the computer became very slow, and the keyboard stopped working until I rebooted again. Also, after copying these files to /usr/share/alsa/, various HDMI outputs appeared in pavucontrol. Before, I had only speakers and heaphones.

Here's the alsa-info of the last value I tried:
[alsa-info.txt](https://github.com/alsa-project/alsa-ucm-conf/files/8498521/alsa-info.txt)

Then, while reviewing this comment before posting it, I looked in the alsa-info.txt and noticed the module is actually called `snd_soc_sof_es8336` and not `snd_soc_sof_essx8336`. Then I tried all quirks again and all I got was dummy output, but the microphone is working.

Here's the alsa-info after trying everything again:
[alsa-info.txt](https://github.com/alsa-project/alsa-ucm-conf/files/8498835/alsa-info.txt)

mchehab:
> Tried these files with the following quirk values (I don't know if all of them are valid, I just tried all combinations possible. There's also a possibility I missed some value, as there's only 15 in my list. I think it should be 16.):

The missing one is to remove the quirks line, letting the driver to use the default quirks ;-)

> 
> 0x010 0x040 0x080 0x100 0x050 0x090 0x110 0x0d0 0x150 0x1d0 0x0c0 0x140 0x1c0 0x180 0x190
> 
> With every value, if I boot the system with headphones connected, the headphones are recognized, but there's no sound. Then, I disconnect the headphones and connect again. I got sound, but the computer became very slow, and the keyboard stopped working until I rebooted again. Also, after copying these files to /usr/share/alsa/, various HDMI outputs appeared in pavucontrol. Before, I had only speakers and heaphones.
> 
> Here's the alsa-info of the last value I tried: [alsa-info.txt](https://github.com/alsa-project/alsa-ucm-conf/files/8498521/alsa-info.txt)
> 
> Then, while reviewing this comment before posting it, I looked in the alsa-info.txt and noticed the module is actually called `snd_soc_sof_es8336` and not `snd_soc_sof_essx8336`. Then I tried all quirks again and all I got was dummy output, but the microphone is working.
> 
> Here's the alsa-info after trying everything again: [alsa-info.txt](https://github.com/alsa-project/alsa-ucm-conf/files/8498835/alsa-info.txt)
> 
> Before trying these files, sound was working, but with the bug described in [thesofproject/linux#3590](https://github.com/thesofproject/linux/issues/3590)

From your comments at:
    https://github.com/thesofproject/linux/issues/3248#issuecomment-1094064700

It seems that your device is not a Comet Lake, but, instead, Gemini Lake... Better to ask @plbossart if those bindings are supposed to work also with Gemini Lake.

plbossart:
@quintafeira when you add quirks, the first rule is to check if you have a card created with visible devices. I have no idea e.g. how the microphones could be working if you have no card...

You also only want to change the quirk fields that correspond to GPIOs and presence of microphones. If  you modify the 'SSP' related fields, then you also need to change the topologies - that's only for power users and no longer needed with the latest code in 5.18-rcX.

The bindings will work the same in GeminiLake, the only thing that is not supported on GeminiLake is the automatic selection of MCLK1 instead MCLK0. I'll have to add this when I have time, and more importantly when someone reports that they actually need that capability and can test it - I don't have any hardware.

fcsm1922:
Hooray! a sound appeared with the following parameters alsamixer item
https://imgur.com/a/eCOrEe0
BUT the sound comes with wheezing (clicks pass) and interference (the volume is reduced)

fcsm1922:
@plbossart, @mchehab
Elimination of clicks and wheezes from the speaker can be incredible with the right selection of quirks (
Options SND_SOC_SOF_SESSX8336 QUIRK)? Or a matter of something else?

perexg:
Applied all commits. Thank you.

PR #145, opened 2022-04-03
Add Use Case for Arturia Minifuse 2

nodens:
Arturia Minifuse 2
(https://www.arturia.com/products/audio/minifuse/minifuse2)
is an USB audio interface with 2 mono inputs and 2 channel outputs.
It uses 1 subinterface with 4 capture and 4 playback channels:
Capture channels:
- Mic/Line/Inst 1 (Mono)
- Mic/Line/Inst 2 (Mono)
- Loopback (Stereo, L+R)

Playback:
- Main (Stereo, L+R)
- Loopback (L+R)

The Loopback capture and playback are intended to ease recording sound
from the computer, e.g if you're doing a podcast and want to record
something that plays from the computer. Of course us linux user don't
need that, since we have many tools able to do that.

However, without a dedicated use case, alsa treats the channels as
surround 4.0, which make the card difficult to use.

At this stage, no controls are available.

TODO/FIXMEs:
- add a channel map override on the pcm device ?

PR #128, opened 2021-12-12
Add profile for Behringer UMC204HD

brndd:
Adds a profile for the Behringer UMC204HD to split a 4-channel output into two outputs, and a 2-channel input into two inputs, as per the physical layout of the device. I have very little idea what I'm doing, so let me know if something looks incorrect (there's one thing that looks incorrect even to me which I shall point out in a comment below). The config does at least work, though.

[alsa-info.txt](https://github.com/alsa-project/alsa-ucm-conf/files/7698277/alsa-info.txt)

-----

UMC204HD_P0BK0_Front_XL
UMC204HD_P0BK0_Rear_XL

PR #153, opened 2022-05-14
Add "Pro Audio" profile for UMC204HD

brndd:
Follow-up to #128 

I figured that some users might prefer to keep the channel mapping untouched from what the card defaults to when unconfigured. Providing an alternative profile like this (implementation adapted from #145) for these users seems like a good idea.

I believe PipeWire autogenerates profiles like these for unconfigured devices, so that's where the "Pro Audio" nomenclature comes from. Likely a good idea to match that since PipeWire is increasingly the default on most distributions.

### DIFF
--- a/ucm2/USB-Audio/Behringer/UMC204HD-ProAudio.conf
+++ b/ucm2/USB-Audio/Behringer/UMC204HD-ProAudio.conf
@@ -1,0 +1,36 @@
+LibraryConfig.generic.Config {
+	ctl.hw {
+		@args [ CARD ]
+		@args.CARD.type string
+		type hw
+		card $CARD
+	}
+
+	pcm.hw {
+		@args [ CARD ]
+		@args {
+			CARD.type string
+		}
+		type hw
+		card $CARD
+		device 0
+	}
+}
+
+SectionDevice."Output" {
+	Comment "UMC204HD 192k Multichannel Output"
+	Value {
+		PlaybackPriority 300
+		PlaybackChannels 4
+		PlaybackPCM "hw:${CardId}"
+	}
+}
+
+SectionDevice."Input" {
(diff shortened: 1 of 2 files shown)


perexg:
Could you test a bit reshuffled version in https://github.com/perexg/alsa-ucm-conf/commit/73e6a327e40cd6fa2f8d6c82cd29c31c4b12e9fc (https://github.com/perexg/alsa-ucm-conf/commit/73e6a327e40cd6fa2f8d6c82cd29c31c4b12e9fc.patch) ?

I prefer to name the verb / device as `Direct` rather than `ProAudio`. Some multichannel cards does not have pro capabilities.

The configuration is generic now and I expect (according the waiting pull requests) that many similar cards will reuse it.

brndd:
Sure. Seems like a good idea, as this is indeed a fairly common problem.

I had to do a minor fix for it to work: `UMC204HD.conf` line 11 should be `Include.dhw.File "/common/direct.conf"`. After that it works perfectly.

The human-readable names (Comment) of the devices look like this. I think it might be good if the name indicated whether it is an input or an output, and the card number might not be useful information. But this is a matter of taste.

2022-05-14T20:30:27

perexg:
> Sure. Seems like a good idea, as this is indeed a fairly common problem.
> 
> I had to do a minor fix for it to work: `UMC204HD.conf` line 11 should be `Include.dhw.File "/common/direct.conf"`. After that it works perfectly.

Oops. Thank for the info.

> The human-readable names (Comment) of the devices look like this. I think it might be good if the name indicated whether it is an input or an output, and the card number might not be useful information. But this is a matter of taste.

The direction information is already in the device attributes. It seems that the frontend only shows an icon. I think it's upon UI to represent this information to users.

I removed the card number, but the purpose was to distinguish somehow multiple instances of the same sound card.

PR #156, opened 2022-05-25
Update mapping for the new Gigabyte Aorus Pro WiFi rev 1.2

13r0ck:
Requires kernel patch from:
https://bugzilla.kernel.org/attachment.cgi?id=301036&action=diff

This doesn't fix the front mic for some reason (any ideas please?)~~, as well I am not sure if this will cause a regression for the 1.1 board~~


13r0ck:
The original commit caused regressions in the 1.1 revision of the board. This now creates a separate map for the rev 1.2, please let me know if there is an easier way :smile:

PR #160, opened 2022-05-28
ucm2 profile for Behringer Flow8

smangels:
- kudos to Pekka Oinas for supporting me finding the right settings
- provides an UCM2 profile for Behringer's digital mixer Flow8
  configured into "streaming" mode
- 2 stereo outputs, 1 stereo input, when vendor ID product ID are `1397:050d`
- prepare filenames for Flow8's recording mode `1397:050c` (10 chan input, 4 chan output)

### DIFF
--- a/ucm2/USB-Audio/Behringer/Flow8-Recording-Hifi.conf
+++ b/ucm2/USB-Audio/Behringer/Flow8-Recording-Hifi.conf
@@ -153,11 +153,92 @@ SectionDevice."Mic4" {
 	}
 }
 
-SectionDevice."Line1" {
-	Comment "Line-56 L/R"
+SectionDevice."Line5" {
+	Comment "Line/Inst 5 (L)"
+
+	ConflictingDevice [
+		"Line56"
+	]
 
 	Value {
 		CapturePriority 104
+	}
+	Macro.pcm_split.SplitPCMDevice {
+		Name "flow8_mono_in"
+		Direction Capture
+		HWChannels 10
+		Channels 1
+		Channel0 4
+		ChannelPos0 MONO
+	}
+}
+
+SectionDevice."Line6" {
+	Comment "Line/Inst(HiZ) 6 (R)"
+
+	ConflictingDevice [
+		"Line56"
+	]
+
+	Value {
+		CapturePriority 105
+	}
+	Macro.pcm_split.SplitPCMDevice {
+		Name "flow8_mono_in"
+		Direction Capture
+		HWChannels 10
+		Channels 1
+		Channel0 5
+		ChannelPos0 MONO
+	}
+}
+
+SectionDevice."Line7" {
+	Comment "Line/Inst 7 (L)"
+
+	ConflictingDevice [
+		"Line78"
+	]
+
+	Value {
+		CapturePriority 106
+	}
+	Macro.pcm_split.SplitPCMDevice {
+		Name "flow8_mono_in"
+		Direction Capture
+		HWChannels 10
+		Channels 1
+		Channel0 6
+		ChannelPos0 MONO
+	}
+}
+
+SectionDevice."Line8" {
+	Comment "Line/Inst(HiZ) 8 (R)"
+
+	ConflictingDevice [
+		"Line78"
+	]
+
+	Value {
+		CapturePriority 107
+	}
+	Macro.pcm_split.SplitPCMDevice {
+		Name "flow8_mono_in"
+		Direction Capture
+		HWChannels 10
+		Channels 1
+		Channel0 7
+		ChannelPos0 MONO
+	}
+}
+
+
+SectionDevice."Line56" {
+	Comment "Line-56 L/R"
+
+	Value {
+		CapturePriority 108
 	}
 	Macro.pcm_split.SplitPCMDevice {
 		Name "flow8_stereo_in"
@@ -165,17 +246,17 @@ SectionDevice."Line1" {
 		HWChannels 10
 		Channels 2
 		Channel0 4
-		Channel0 5
+		Channel1 5
 		ChannelPos0 FL
 		ChannelPos1 FR
 	}
 }
 
-SectionDevice."Line2" {
+SectionDevice."Line78" {
 	Comment "Line-78 L/R"
 
 	Value {
-		CapturePriority 105
+		CapturePriority 109
 	}
 	Macro.pcm_split.SplitPCMDevice {
 		Name "flow8_stereo_in"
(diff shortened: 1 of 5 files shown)


perexg:
Could you give a look to the latest changes are use the SplitPCM macro in this configuration ? Thank you.

You can use `USB-Audio/GoXLR/GoXLR-HiFi.conf` and/or `USB-Audio/Arturia/Minifuse-2-HiFi.conf` as reference.

smangels:
How could I test that setup? Is there a way to easily switch between standard ALSA library and the one I've build form latest source code? Could that be done using symbolic linking? What is the recommended setup @perexg ?

perexg:
> How could I test that setup? Is there a way to easily switch between standard ALSA library and the one I've build form latest source code? Could that be done using symbolic linking? What is the recommended setup @perexg ?

It's not so easy. You can use LD_PRELOAD and run the audio server (binary) manually with this environment variable (example for alsaucm: `LD_PRELOAD="$HOME/alsa/alsa-lib/src/.libs/libasound.so" alsaucm -c hw:1 dump text`).

smangels:
Ok, I'll wait with testing until the release has reached ArchLinux. The code changes to use the new MACRO syntax have been pushed into the PR.

smangels:
just for documentation purposes, here is the link to the block diagram for that mixer https://mediadl.musictribe.com/download/software/behringer/FLOW/FLOW8/BE-P0DNM-FLOW8-Block-Diagram.pdf

smangels:
I've tested the configuration file together with release of alsa-ucm-conf 1.2.7 on an ArchLinux machine in both _Streaming_ as well as _Recording_ mode. Everything works as expected. I really like the new MACRO based interface. @perexg do you need any other test which proofs that the configuration is working
Screenshot from 2022-06-12 01-00-50
Screenshot from 2022-06-12 01-00-58
Screenshot from 2022-06-12 01-01-01
?

smangels:
And here is the recording mode detected by UCM2 after switching the mode in the Behringer mixer Flow8, all mono and stereo channels are correctly mapped.

Screenshot from 2022-06-12 01-01-21
Screenshot from 2022-06-12 01-01-30
.

smangels:
@perexg the only issue I've detected is that there is a noticeable delay between video and audio. I guess that we need some kind of buffer size description? Is that something that we would like to configure in UCM2 as well?

perexg:
I merged your changes (in one commit). I kept the direct profiles with small extension in 5dcd693973aa4883b2cf5b5504e8f3e646850e22 . Thank you for your contribution.

perexg:
> @perexg the only issue I've detected is that there is a noticeable delay between video and audio. I guess that we need some kind of buffer size description? Is that something that we would like to configure in UCM2 as well?

The delay synchronization should be handled by the application and the driver. The special buffer settings should not be required. It appears like an issue in the driver. You may report this in the kernel bugzilla: https://bugzilla.kernel.org (Audio component).

smangels:
It is very interesting that this doesn't happen in DIRECT mode but when running my own profile applied using UCM2. Why isn't that visible in DIRECT mode? Strange

perexg:
> It is very interesting that this doesn't happen in DIRECT mode but when running my own profile applied using UCM2. Why isn't that visible in DIRECT mode? Strange

It may be an issue in the alsa-lib's dshare plugin then. I'll give a look.

smangels:
Appreciate it, let me know if you need anything from `lsusb` or `alsa-info.sh`

smangels:
Hi, any news regarding the delay issue? It is still audible even in the latest release of ALSA pushed to ArchLinux.

perexg:
I did some tests with another USB card to verify, if the alsa-lib plugins used for the splitting may cause it. I cannot confirm this. The USB driver adds a little delay (few ms), but it should not be noticable. What is your kernel version?

Could you check the stream settings in hardware like:

```
$ cat /proc/asound/card5/pcm0p/sub0/*
access: MMAP_INTERLEAVED
format: S16_LE
subformat: STD
channels: 2
rate: 48000 (48000/1)
period_size: 6000
buffer_size: 18000
...
```

smangels:
I'll give it a try as soon I'm back home again. We're celebrating midsommar today here in Sweden.
/Sebastian 


BR,

Re: 
> 
> 
> I did some tests with another USB card to verify, if the alsa-lib plugins used for the splitting may cause it. I cannot confirm this. The USB driver adds a little delay (few ms), but it should not be noticable. What is your kernel version?
> 
> Could you check the stream settings in hardware like:
> 
> `$ cat /proc/asound/card5/pcm0p/sub0/*
> access: MMAP_INTERLEAVED
> format: S16_LE
> subformat: STD
> channels: 2
> rate: 48000 (48000/1)
> period_size: 6000
> buffer_size: 18000
> ...
`

smangels:
Hi @perexg is there anything that I could do to support you finding the root cause?

perexg:
Could you try this configuration change for a test ?

```
diff --git a/ucm2/common/pcm/split.conf b/ucm2/common/pcm/split.conf
index 95a65b5..2a6bb46 100644
--- a/ucm2/common/pcm/split.conf
+++ b/ucm2/common/pcm/split.conf
@@ -40,6 +40,8 @@ DefineMacro.SplitPCM.If.0 {
                                        chmap [ "${var:__HWChannelPos0}" ]
                                }
                                channels "${evali:$__HWChannels}"
+                               period_size 960
+                               buffer_size 19200
                        }
                        bindings.0 $CHN0
                }
```

smangels:
Screenshot from 2022-07-04 23-07-41

No audible change, still the same delay between video and audio when watching music videos on Youtube.

smangels:
Interestingly, things change when cutting the values in half (9600, 960) was almost there, when going lower than this, the CPU load went up to almost 20%

bfySebman:
On a second machine running Fedora 35, Alsa-Lib 1.2.6 and UCM version 1.2.6.1 there is no delay in audio. Kernel is 5.18.5-100
```
$ cat /proc/asound/card2/pcm0p/sub0/*
access: MMAP_INTERLEAVED
format: S32_LE
subformat: STD
channels: 4
rate: 48000 (48000/1)
period_size: 64
buffer_size: 512
card: 2
device: 0
subdevice: 0
stream: PLAYBACK
id: USB Audio
name: USB Audio
subname: subdevice #0
class: 0
subclass: 0
subdevices_count: 1
subdevices_avail: 0
state: RUNNING
owner_pid   : 3179
trigger_time: 670185.170727818
tstamp      : 670478.442432953
delay       : -14078948
avail       : 14079724
avail_max   : 14079724
-----
hw_ptr      : 14079212
appl_ptr    : 0
tstamp_mode: ENABLE
period_step: 1
avail_min: 64
start_threshold: 1
stop_threshold: 4611686018427387904
silence_threshold: 0
silence_size: 4611686018427387904
boundary: 4611686018427387904

```
Interestingly that the period_size and buffer_size are much smaller.

perexg:
I did some tests and I can confirm that pulseaudio does use the whole buffer for those devices. I think that it's a pulseaudio issue, because pipewire does not have any problems with the similar settings. Nothing forces applications to use the whole buffer (partial queue is also fine). I lowered the period size in commit fb69d0f0274b03e1ead0a7feeaceac608760dc72 which may improve latency (for pipewire).

smangels:
oh great, I wonder why the second system I've described in the comments above doesn't show that behavior. Do you think that a change in pulseaudio has caused that behavior using the complete buffer? I'll give it a try once it has been released and arrive in ArchLinux repositories. 
Thanks. 
/Sebastian 


BR,

Re:

Issue #165, opened 2022-06-04
ASUS ROG Maximus Z690 Hero (ALC4082) - No audio from microphone, no audio on the front panel


Lastique:
I'm having a problem with getting audio from the microphone connected to the rear panel jack. Also, if I connect the headset to the front panel headphone/mic jacks, I get neither audio out or in. My main concern is the rear panel jack though. The motherboard is ASUS ROG Maximus Z690 Hero, ALC4082 USB Audio chipset.

I have tried this:

1. Installed linux kernel 5.18.0.
2. Copied `ucm2` from alsa-ucm-conf master (rev ef3b6513b6581705d36432428a7f6a4ae23b1276 which is after https://github.com/alsa-project/alsa-ucm-conf/pull/143 was merged) to `/usr/share/alsa`.
3. In `/usr/share/alsa/ucm2/USB-Audio/USB-Audio.conf` added hardware id of my USB Audio device (`0b05:1a27`) to the `realtek-alc4080` rule. Also added the hardware id in `/usr/share/alsa/ucm2/USB-Audio/Realtek/ALC4080-HiFi.conf`, in the `asus-rog-usb` block condition.
4. Rebooted.
5. The headset is connected to the rear panel (Front out and Mic in jacks; also tried Line in).

There are three devices for the USB Audio card, and there are three volume controls in alsamixer: "Line", "Mic" and "Analog In". The problem is that there is no audio coming from the first two devices (which are presumably "Line" and "Mic"). I can record audio from the third device ("Analog In") when the microphone is connected to Mic jack.

Unfortunately, "Analog In" cannot be selected in PulseAudio for some reason (I created a [bug](https://gitlab.freedesktop.org/pulseaudio/pulseaudio/-/issues/1364) for PA), which means no audio can be recorded.

I also tried to connect the headset to the front panel connectors, but got no audio output nor capture. This is not my intended configuration though, I normally connect the headset to the rear panel.

`arecord` output:
[arecord.txt](https://github.com/alsa-project/alsa-ucm-conf/files/8838239/arecord.txt)
`lsusb` output:
[lsusb.txt](https://github.com/alsa-project/alsa-ucm-conf/files/8838250/lsusb.txt)
`alsa-info.sh` output:
[alsa-info.txt](https://github.com/alsa-project/alsa-ucm-conf/files/8846120/alsa-info.txt)

This is on Kubuntu 22.04.

Lastique:
[...]

PR #166, opened 2022-06-06
Add basic support for ALC4082 on ASUS ROG Maximus Z690 Hero

Lastique:
This enables Microphone In in the rear jack and Headphones Out in the front panel jack on ASUS ROG Maximus Z690 Hero with Realtek ALC4082.

Things that still don't work:

- Microphone In on the front panel
- Side channel output on the Line In jack in 7.1 mode
- Rear channels and Center/Sub are swapped

Line In on the rear panel detects microphone plugged in. Actual input not tested (no line out device available; no input from microphone). S/PDIF not tested.

Related to https://github.com/alsa-project/alsa-ucm-conf/issues/165.

### DIFF
--- a/ucm2/USB-Audio/USB-Audio.conf
+++ b/ucm2/USB-Audio/USB-Audio.conf
@@ -35,10 +35,11 @@ If.realtek-alc4080 {
 		Type RegexMatch
 		String "${CardComponents}"
 		# 0b05:1996 ASUS on multiple boards (including ASUS ROG Maximus XIII)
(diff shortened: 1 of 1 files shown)


Lastique:
Also, I should note that in the KDE System Settings that I use to test channel output, the output device always shows 7.1 configuration, regardless of what is connected. I'm not sure if this is an ALSA or PulseAudio issue. There is no way to choose, say, stereo mode in pavucontrol if only stereo headphones are connected. Is this something that needs to be done in PulseAudio or ALSA?

PR #167, opened 2022-06-07
ucm2: add the ucm support for the mt8195-demo platform

fchiby:


fchiby:
This use alsa-ucm-conf package to detect audio hardware automatically for supporting all AIoT platforms in one single image. 
This pull request:
    - Add the initial support of alsa ucm for the mt8195-demo platform and enable JACK, I2S and PDM devices. 
    - Add HDMI and DP audio for i1200-demo

ucm2/MediaTek/mt8195_demo/HiFi.conf
```diff
@@ -0,0 +1,131 @@
+SectionVerb {
+	EnableSequence [
+		cset "name='Headset Volume' 0"
```
perexg, line 3: The card initialization should be in the BootSequence section.

fchiby: Thanks for the review. I've just pushed a new version which fixing this.

perexg:
Applied with two changes:

- fixed the symlink from conf.d
- removed duplicated Dp device

PR #173, opened 2022-06-19
tegra: max98090: add speakers and internal mic

jenneron:


TuxThePenguin0:
I've tested this UCM config file on google-nyan-big, it fixes the speaker output and makes automatic headphone and microphone switching function (microphone switching also needs a kernel patch because of a misconfigured gpio). As far as I can tell audio is fully functional.
I haven't tested any other devices.

ucm2/Tegra/max98090/HiFi.conf
```diff
@@ -37,10 +37,38 @@ SectionVerb {
...
+	ConflictingDevice [
+		"Headphones"
+	]
+	Value {
+		PlaybackPriority 100
+		PlaybackPCM "hw:${CardId}"
+		PlaybackVolume "Speaker Volume"
+	}
+	EnableSequence [
+		cset "name='Speakers Switch' on"
```
perexg, line 52: Could you rewrite the configuration so:

1) all switches are off in the Verb.EnableSequence
2) manipulate only with the necessary controls in the Device sequences

Also, the default values should be moved from Verb.EnableSequence to the main file - BootSequence.

jenneron: Done, but i'm not sure i understand this step correctly:
> all switches are off in the Verb.EnableSequence

Issue #190, opened 2022-07-17
USB-Audio: Motu M2: Audio interface unable to be loaded in 1.2.7.2


KevHg:
I upgraded from 1.2.6.1 to 1.2.7.2 and found that pulseaudio no longer detects my Motu M2 interface.

Main error log seems to be:
`module-alsa-card.c: Failed to find a working profile.`
`failed to load module module-alsa-card`

I'm new to debugging Linux issues, let me know what other info I can provide to help out on this topic.

perexg:
It looks like that Motu M2 has similar USB ID 07fd:000b like Motu M4: https://github.com/alsa-project/alsa-ucm-conf/pull/158/ .

Could someone attach output from `alsa-info.sh --no-upload` ?

sreich:
Here's mine. This is with the working 1.2.7.1, lemme know if you actually want the broken version's output instead

[alsa-info.txt](https://github.com/alsa-project/alsa-ucm-conf/files/9128478/alsa-info.txt)

atomGit:
> Downgrading alsa-lib, alsa-ucm-conf, and lib32-alsa-lib to 1.2.7.1-1 ...

...worked for me also on Manjaro - i used the downgrade package from the AUR to get around the "Downgrading from A.L.A. is disabled on the stable branch." issue

no reboot necessary, i just power-cycled the DAC

kiosion:
Running into this as well, thankfully downgrading worked for me as well. I've always had intermittent issues with the M2 but it seems totally borked on the latest update.

domsch1988:
Same on my Fresh Fedora 36 Install. alsa-lib 1.2.7.2-1
Downgrade on my fresh install doesn't seem to be so easy, as there are some file collisions between the versions. If i can provide anything to help this issue, let me know.

atomGit:
did you try downgrading only alsa-ucm-conf? that alone worked for me (as i later figured out)

domsch1988:
Yeah i was able to downgrade alsa-ucm to 1.2.7.1-1 through a direct link rather than getting it from the repo. This worked fine. Just need to watch my updates a bit until this is getting fixed.

sreich:
@domsch1988 been a while since i used fedora but in your /etc/dnf/dnf.conf you can add exclude=packagename

i did the same for arch so i don't have to bother with it each time until this gets fixed.

[...]

PR #158, opened 2022-05-26
ucm2 profile for MOTU M4 interface

rjahanbakhshi:
This is the initial ucm2 profile for MOTU M4 interface. This creates a default profile with 2 separate line-outs and 4 line-ins/mic. The profile follows the device IO design and covers the most default use cases. For the rest, direct.conf is included so the pro profile in pipewire will be accessible for more complex use-cases.

MOTU M4 is a 4×4 interface. It has 4 simultaneous inputs and outputs.  For input, there are 2 combo jacks on the front that provide a preamp with XLR connectivity for microphones, as well as 1/4″ Hi-Z compatibility. Rounding out the inputs are a pair of Line Ins on the back of the interface.

For outputs, it has two pairs of DC-coupled 1/4″ balanced outs; one is labeled Monitor and the other Line Out. Line Out can be sent to another pair of monitors, PA speakers, or any destination. Each pair of outputs is independent.

Additionally, there are two pairs of mirrored RCA outputs. Mirrored, in this case, means that they receive the same output that the Monitor outs and the Line Outs get. You can route those to more speakers, or another audio source with RCA inputs. MIDI in and out expands this interface’s abilities even further.

PR #191, opened 2022-07-17
ucm2 profile for MOTU M2 

tw1618:
With Release v1.2.7.2 my Audio Interface Motu M2 stopped working as discussed in #190. It looks like that the Motu M4 config #158 is loaded as they use the same (?) ID

Those changes work for me.

[alsa-info.txt](https://github.com/alsa-project/alsa-ucm-conf/files/9128719/alsa-info.txt)

### DIFF
--- a/ucm2/USB-Audio/USB-Audio.conf
+++ b/ucm2/USB-Audio/USB-Audio.conf
@@ -67,11 +67,20 @@ If.steinberg-ur44 {
 
 If.M4 {
 	Condition {
-		Type String
-		Haystack "${CardComponents}"
-		Needle "USB07fd:000b"
+		Type RegexMatch
+		String "${CardName}"
+		Regex "M4"
 	}
 	True.Define.ProfileName "MOTU/M4"
+}
+
+If.M2 {
+	Condition {
+		Type RegexMatch 
+		String "${CardName}"
+		Regex "M2"
+	}
+	True.Define.ProfileName "MOTU/M2"
 }
 
 If.dell-wd15 {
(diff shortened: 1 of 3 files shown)


ember-rose:
Confirmed this works with 1.2.7.2 on my machine. Thanks!

ucm2/USB-Audio/USB-Audio.conf
```diff
@@ -67,13 +67,22 @@ If.steinberg-ur44 {
 
 If.M4 {
 	Condition {
-		Type String
-		Haystack "${CardComponents}"
-		Needle "USB07fd:000b"
+		Type RegexMatch
```
perexg, line 70: This is very bad condition. It will match any USB cards with `M4` in the `CardName`. Keep the `USB07fd:000b` check and extend the `True` block like:

```
If.Motu {
  Condition {
    Type String
    Haystack "${CardComponents}"
    Needle "USB07fd:000b"
  }
  True.If.M4 {
    Condition {
      Type String
      Haystack "${CardLongName}"
      Needle "MOTU M4"
    }
    True.Define.ProfileName "MOTU/M4"
    False.Define.ProfileName "MOTU/M2"
  }
}
```

perexg (changes requested):

tw1618:
Thank you for your review! All changes should be addressed now.

perexg:
Merged in 3a6acaa5090a0c5a0d29c8a4906b6a60378774e2 with some cosmetic fixes. Thanks.

PR #193, opened 2022-07-26
rt715-sdca: use sensible capture gain value

plbossart:
The value of 124 used for FU02 is way too much, with saturation even in low-volume cases. The 0dB value (47) is a much better initialization value.

BugLink: https://github.com/thesofproject/linux/issues/3766
Signed-off-by: Pierre-Louis Bossart <pierre-louis.bossart@linux.intel.com>


fredoh9 (approved):
I tested this with Dell Andrew laptop.
47 (0 db) create nice and moderately good waveform. 124 is way too high.

PR #196, opened 2022-08-04
USB-Audio: Add profile for MSI MEG Z690I Unify

13r0ck:
This is tested and works, though I am not a fan of my `Needle` and `Haystack` solution here, but before also making a PR to alsa-utils I wanted direction. Would it be better to add a `Type Bool` condition, or a `skip` modifier to `SectionDevice`? I prefer the later, because it skips the `If {} Condition {} true {}` lines, but I would like an alsa maintainer's opinion.


13r0ck:
I'm not sure what I am missing, but I would love to help get this merged :smiley:

perexg:
Merged differently through b51437390a12dbfb881574b7712af71e61f6f48e . It would be nice, if you can test the latest ucm configuration tree.

PR #197, opened 2022-08-08
ASUS ROG Zenith II Extreme UCM2 profile

nabajour:
Hi, 
I'm trying to get sound working on my ASUS ROG Zenith II Extreme motherboard. There is a [kernel patch](https://bugzilla.kernel.org/show_bug.cgi?id=211005) that I'm using for a while, that seems to get most ALSA stuff working. It kinda also works with the default ProAudio ACP profile (minus some case where it seems to loose synch after a while. But recording works).

Now, I'm trying to get it to work correctly with my Debian unstable desktop, using PipeWire with the PA tools (mostly pavucontrol).

Mostly, the routing of the inputs for capture seem to not work. It's always routing the input from `hw:0,0`, which is `Line In`.

I tried making a UCM2 profile, and got some basic stuff working. but there are some things where I'm stuck.

What I do is edit the UCM files, and then reload PW and WP with `systemctl --user restart pipewire pipewire-pulse wireplumber`. Then, I check with `pavucontrol` what devices appear. If there is a syntax error, it falls back to `Off` profile, or `ProAudio` profile, which was an experiment from me to see if I could just have the same pass-through behaviour as the ACP pass-through. I don't know if there is a better way to debug UCM2 profile loading, this approach leaves me pretty blind on what failed. 

Here are the issues I currently have:

1. surround input channels are mixed up compared to alsa channels (testing with `speaker-test` output, sound goes to wrong outputs with pw.
2. Can't get SPDIF to work. When I uncomment the SPDIF block, it doesn't load the profile. 
3. Mic 1 (the rear mic) appears and reacts correctly to jack plug/unplug. But recording seems to take the input from `hw:0,0` instead of `hw:0,1`
4. uncommenting the Line In or the Mic2 (front mic)  block fails. It doesn't load the profile.

I don't know if the issue is still on the ALSA driver side or on the PW/PA config. I'm not a kernel programmer and don't know much about ALSA and PW, so any pointer here is welcome.


nabajour:
For the Mic input issue, here is something interesting:

``` 
$ wpctl status                                                                  
PipeWire 'pipewire-0' [0.3.56, nab@halo, cookie:3707031228]
 └─ Clients:
        31. xdg-desktop-portal                  [0.3.56, nab@halo, pid:3046]
        32. WirePlumber                         [0.3.56, nab@halo, pid:44330]
        33. WirePlumber [export]                [0.3.56, nab@halo, pid:44330]
        36. pipewire                            [0.3.56, nab@halo, pid:44331]
        62. PulseAudio systray                  [0.3.56, nab@halo, pid:3095]
        63. wpctl                               [0.3.56, nab@halo, pid:47690]
        97. qpwgraph                            [0.3.56, nab@halo, pid:3111]

Audio
 ├─ Devices:
 │      41. USB Audio                           [alsa]
 │
 ├─ Sinks:
 │      46. USB Audio Front Headphones          [vol: 0.97]
 │  *   47. USB Audio Speakers                  [vol: 0.59]
 │
 ├─ Sink endpoints:
 │
 ├─ Sources:
 │  *   48. USB Audio Microphone                [vol: 1.53]
 │
 ├─ Source endpoints:
 │
 └─ Streams:

Video
 ├─ Devices:
 │
 ├─ Sinks:
 │
 ├─ Sink endpoints:
 │
 ├─ Sources:
 │
 ├─ Source endpoints:
 │
 └─ Streams:

Settings
 └─ Default Configured Node Names:
         0. Audio/Sink    alsa_output.usb-Generic_USB_Audio-00.HiFi__hw_Audio__sink
         1. Audio/Source  alsa_input.usb-Generic_USB_Audio-00.HiFi__hw_Audio_1__source
```

So my mic device is 48. 

```
$ wpctl inspect 48                                                           
id 48, type PipeWire:Interface:Node
    alsa.card = "0"
    alsa.card_name = "Zenith II Main Audio"
    alsa.class = "generic"
    alsa.device = "0"
    alsa.driver_name = "snd_usb_audio"
    alsa.id = "USB Audio"
    alsa.long_card_name = "Zenith-II-Main-Audio"
    alsa.mixer_device = "_ucm0001.hw:Audio"
    alsa.name = "USB Audio"
    alsa.resolution_bits = "16"
    alsa.subclass = "generic-mix"
    alsa.subdevice = "0"
    alsa.subdevice_name = "subdevice #0"
    api.alsa.card.longname = "Zenith-II-Main-Audio"
    api.alsa.card.name = "Zenith II Main Audio"
    api.alsa.open.ucm = "true"
    api.alsa.path = "hw:Audio,1"
    api.alsa.pcm.card = "0"
    api.alsa.pcm.stream = "capture"
    audio.adapt.follower = ""
    audio.channels = "2"
    audio.position = "FL,FR"
    card.profile.device = "4"
  * client.id = "33"
    clock.quantum-limit = "8192"
    device.api = "alsa"
    device.class = "sound"
  * device.id = "41"
    device.profile.description = "Microphone"
    device.profile.name = "HiFi: hw:Audio,1: source"
    device.routes = "1"
  * factory.id = "18"
    factory.mode = "split"
    factory.name = "api.alsa.pcm.source"
    library.name = "audioconvert/libspa-audioconvert"
  * media.class = "Audio/Source"
  * node.description = "USB Audio Microphone"
    node.driver = "true"
    node.max-latency = "16384/48000"
  * node.name = "alsa_input.usb-Generic_USB_Audio-00.HiFi__hw_Audio_1__source"
  * node.nick = "Zenith II Main Audio"
    node.pause-on-idle = "false"
  * object.path = "alsa:pcm:0:hw:Audio,1:capture"
  * object.serial = "49"
  * priority.driver = "2000"
  * priority.session = "2000"
```

If I read this correctly, it finds the alsa mic as `hw:Audio,1`, (Audio being Card 0), here: `api.alsa.path = "hw:Audio,1"`. Which is what I expect, Card 0, Device 1. But it shows also this:

```
    alsa.card = "0"
    alsa.card_name = "Zenith II Main Audio"
    alsa.class = "generic"
    alsa.device = "0"
```

Is it jumping back to device 0 ?

nabajour:
Progress! I tried this:

```
$ systemctl stop pipewire.socket ; systemctl stop pipewire-pulse.socket
$ spa-acp-tool -vvv -c 0
```
As the output was very long, I commented out my `SectionUseCase."ProAudio"`, which was a test anyway. 

As I didn't see anything weird in there, I added `SPDIF`, `Line In`, `Mic2`. Nothing weird in the output either. Playing around and looking at the `list-verbose` output, it looked like the devices where correct.

Then I tried 
```
$ systemctl stop pipewire.socket ; systemctl stop pipewire-pulse.socket
```
And lo and behold, my devices appear in pavucontrol. After a reboot, it still seems to be working. If I change the code to reactivate the `SectionUseCase."ProAudio"`, it fails again, with the same pattern (meaning that it fails to create the profile with everything activated, but if I disable the SPIDIF, Mic2 and Line In, it loads the profile again.) 

I don't know how those SectionUseCase interfere, but it looks like I should just get rid of the ProAudio SectionUseCase test. 

So looks like it's a step forward. I'll test further and play around to tweak the config.

nabajour:
So here is what I'm still missing now:
- the mapping of 5.1 channels on analog output through hw:0,0 with aplay and pw-play is different. The channels seem to be swapped (tested with `speaker-test`)
- no 5.1 sound output on digital output, only 2.1. Does it need some compression config somewhere, like alsa needs the a52 plugin?
- digital outputs hw:0,1 and hw:0,3 work when added separately in the profile. But I can't add both to the profile. I don't know what the difference is between them.

PR #200, opened 2022-08-19
ucm2: Add intial acpes8336 support

yangxiaohua2009:
This add support for es8336 on AMD  Jadeite platform.

ASLAINFO http://alsa-project.org/db/?f=b5aae41350b63c40c7743ba43ff8a7178ee93229


perexg:
I'm afraid, the file contains really old syntax (extra cset lines with hardcoded card names, no hardware volume support).

Files are also placed to the wrong directory (it should be in AMD/acp/es8336 subdirectory or so).

I suggest to look to the current configuration files and modify the configuration accordingly.

PR #202, opened 2022-08-21
bug: Fix verbose logging in GitHub Workflow

PatrickLang:
I noticed the GitHub Workflow `ucm_validate` was setting `LOGLEVEL` instead of `V` which is used in the `Makefile` to set the log level. The tests in this repo are still failing, but fixing the logging first should help.


ChrisLane:
@perexg is this PR not wanted?

PR #207, opened 2022-08-28
ucm2: HDA - add support for Internal Mic

kasper93:
Simple patch to support also "Internal Mic".

Everything seems to be working fine after this change, except that I don't have "Internal Mic Jack", I'm lying there a little to set it as such. Everything switches nicely when connecting/disconnecting headphones mic. But there are two input devices to select and selection doesn't have effect. When headphone mic is connected the internal one should be hidden (it is handled somehow when using HDA driver), but there is no Jack and I don't know how to handle it properly. (It is marked as conflicting device with the other one though) Except this minor annoyance it is fine.  

For reference: [alsa-info.txt](https://github.com/alsa-project/alsa-ucm-conf/files/9439697/alsa-info.txt)

### DIFF
--- a/ucm2/HDA/HiFi-analog.conf
+++ b/ucm2/HDA/HiFi-analog.conf
@@ -1,5 +1,6 @@
 # Generic HDA devices for analog I/O
 
+Define.FrontMicName ""
 Define.DeviceFrontMic ""
 Define.LineDevice ""
 Define.hpvol "Headphone"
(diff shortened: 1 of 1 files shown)


perexg:
Merged in 869ef171d3e7d6c87c283c7fcf5dd6ce79a6da78 with the empty Jack condition + new variable. The legacy code uses the Phantom Jack switch, but for UCM, the conflicting device should be enough. It would be better to fix the UCM code in pulseaudio/pipewire to handle this issue correctly.

PR #211, opened 2022-09-01
mt8195-demo: redefine soundcard initialization

fchiby:
The previous initialization in Bootsequence set all Switch to off after boot. Whithout alsactl, alsa-ucm-conf does't work.
So remove the setting off of the Switch in the BootSequence and define Enable/DisableSequence in each device section for a proper setting.

Signed-off-by: Fadwa Chiby <fchiby@baylibre.com>

### DIFF
--- a/ucm2/MediaTek/mt8195_demo/mt8195_demo.conf
+++ b/ucm2/MediaTek/mt8195_demo/mt8195_demo.conf
@@ -1,4 +1,4 @@
-Syntax 3
+Syntax 2
 
 SectionUseCase."HiFi" {
 	File "/MediaTek/mt8195_demo/HiFi.conf"
@@ -16,12 +16,12 @@ BootSequence [
 	cset "name='PGA_L_Mux' AIN1"
 	cset "name='HDMI_OUT_MUX' Connect"
 	cset "name='DPTX_OUT_MUX' Connect"
+	# we need to enable all devices before starting PA.
+	# In our driver we use PCM, which means that we have
+	# to route Front End to a BackEnd and then only we can
+	# open a device. Without linking to BE it will fail.
 	cset "name='O176 I070 Switch' on"
 	cset "name='O177 I071 Switch' on"
-	cset "name='O034 I168 Switch' on"
-	cset "name='O035 I169 Switch' on"
-	cset "name='O036 I012 Switch' on"
-	cset "name='O037 I013 Switch' on"
 	cset "name='O072 I022 Switch' on"
 	cset "name='O073 I023 Switch' on"
 	cset "name='O074 I024 Switch' on"
@@ -30,6 +30,10 @@ BootSequence [
 	cset "name='O077 I027 Switch' on"
 	cset "name='O078 I028 Switch' on"
 	cset "name='O079 I029 Switch' on"
+	cset "name='O034 I168 Switch' on"
+	cset "name='O035 I169 Switch' on"
+	cset "name='O036 I012 Switch' on"
+	cset "name='O037 I013 Switch' on"
 	cset "name='O002 I004 Switch' on"
 	cset "name='O003 I005 Switch' on"
 	cset "name='O004 I006 Switch' on"
@@ -38,26 +42,4 @@ BootSequence [
 	cset "name='O007 I009 Switch' on"
 	cset "name='O008 I010 Switch' on"
 	cset "name='O009 I011 Switch' on"
-	cset "name='O176 I070 Switch' off"
-	cset "name='O177 I071 Switch' off"
-	cset "name='O034 I168 Switch' off"
-	cset "name='O035 I169 Switch' off"
-	cset "name='O036 I012 Switch' off"
-	cset "name='O037 I013 Switch' off"
-	cset "name='O072 I022 Switch' off"
-	cset "name='O073 I023 Switch' off"
-	cset "name='O074 I024 Switch' off"
-	cset "name='O075 I025 Switch' off"
-	cset "name='O076 I026 Switch' off"
-	cset "name='O077 I027 Switch' off"
-	cset "name='O078 I028 Switch' off"
-	cset "name='O079 I029 Switch' off"
-	cset "name='O002 I004 Switch' off"
-	cset "name='O003 I005 Switch' off"
-	cset "name='O004 I006 Switch' off"
-	cset "name='O005 I007 Switch' off"
-	cset "name='O006 I008 Switch' off"
-	cset "name='O007 I009 Switch' off"
-	cset "name='O008 I010 Switch' off"
-	cset "name='O009 I011 Switch' off"
 ]
--- a/ucm2/MediaTek/mt8195_demo/HiFi.conf
+++ b/ucm2/MediaTek/mt8195_demo/HiFi.conf
@@ -1,3 +1,61 @@
+SectionVerb {
+
+       EnableSequence [
+		cset "name='HDMI_OUT_MUX' Connect"
+		cset "name='DPTX_OUT_MUX' Connect"
+		cset "name='O176 I070 Switch' on"
+		cset "name='O177 I071 Switch' on"
+		cset "name='O034 I168 Switch' on"
+		cset "name='O035 I169 Switch' on"
+		cset "name='O036 I012 Switch' on"
+		cset "name='O037 I013 Switch' on"
+		cset "name='O072 I022 Switch' on"
+		cset "name='O073 I023 Switch' on"
+		cset "name='O074 I024 Switch' on"
+		cset "name='O075 I025 Switch' on"
+		cset "name='O076 I026 Switch' on"
+		cset "name='O077 I027 Switch' on"
+		cset "name='O078 I028 Switch' on"
+		cset "name='O079 I029 Switch' on"
+		cset "name='O002 I004 Switch' on"
+		cset "name='O003 I005 Switch' on"
+		cset "name='O004 I006 Switch' on"
+		cset "name='O005 I007 Switch' on"
+		cset "name='O006 I008 Switch' on"
+		cset "name='O007 I009 Switch' on"
+		cset "name='O008 I010 Switch' on"
+		cset "name='O009 I011 Switch' on"
+       ]
+       
+      DisableSequence [
+		cset "name='HDMI_OUT_MUX' Disconnect"
+		cset "name='DPTX_OUT_MUX' Disconnect"
+		cset "name='O176 I070 Switch' off"
+		cset "name='O177 I071 Switch' off"
+		cset "name='O034 I168 Switch' off"
+		cset "name='O035 I169 Switch' off"
+		cset "name='O036 I012 Switch' off"
+		cset "name='O037 I013 Switch' off"
+		cset "name='O072 I022 Switch' off"
+		cset "name='O073 I023 Switch' off"
+		cset "name='O074 I024 Switch' off"
+		cset "name='O075 I025 Switch' off"
+		cset "name='O076 I026 Switch' off"
+		cset "name='O077 I027 Switch' off"
+		cset "name='O078 I028 Switch' off"
+		cset "name='O079 I029 Switch' off"
+		cset "name='O002 I004 Switch' off"
+		cset "name='O003 I005 Switch' off"
+		cset "name='O004 I006 Switch' off"
+		cset "name='O005 I007 Switch' off"
+		cset "name='O006 I008 Switch' off"
+		cset "name='O007 I009 Switch' off"
+		cset "name='O008 I010 Switch' off"
+		cset "name='O009 I011 Switch' off"
+      ]
+	
+}
+
 SectionDevice."HDMI" {
 	Comment "Hdmi/DP output"
 


chhsieh0810:
Test: pass (Can hear audio from headset after changing the default sink)
```
$ pactl list cards
Card #0
        Name: alsa_card.platform-mt8195-sound
        Driver: module-alsa-card.c
        Owner Module: 7
        Properties:
                alsa.card = "0"
                alsa.card_name = "mt8195_demo"
                alsa.long_card_name = "mt8195_demo"
                device.bus_path = "platform-mt8195-sound"
                sysfs.path = "/devices/platform/mt8195-sound/sound/card0"
                device.form_factor = "internal"
                device.string = "0"
                device.description = "Built-in Audio"
                module-udev-detect.discovered = "1"
                device.icon_name = "audio-card"

$ pactl list short sinks
0 alsa_output.platform-mt8195-sound.HiFi__hw_mt8195demo_6__sink module-alsa-card.c s16le 8ch 44100Hz SUSPENDED
1 alsa_output.platform-mt8195-sound.HiFi__hw_mt8195demo_0__sink module-alsa-card.c s16le 2ch 44100Hz SUSPENDED
2 alsa_output.platform-mt8195-sound.HiFi__hw_mt8195demo_5__sink module-alsa-card.c s16le 2ch 48000Hz SUSPENDED

$ pactl set-default-sink 1
```

fchiby:
> Test: pass (Can hear audio from headset after changing the default sink)
Thank you @chhsieh0810

perexg:
I am a bit lost with the *BootSequence* change. If alsactl is not used, is not safer to keep settings to off and let users to enable the paths manually ? Also the trigger (on/off) is probably not necessary.

Also note that the Verb/EnableSequence should disable all devices (put controls to a safe initial state). The new command *disdevall ""* may be used for this purpose.

fchiby:
> I am a bit lost with the _BootSequence_ change. If alsactl is not used, is not safer to keep settings to off and let users to enable the paths manually ? Also the trigger (on/off) is probably not necessary.
> 
> Also note that the Verb/EnableSequence should disable all devices (put controls to a safe initial state). The new command _disdevall ""_ may be used for this purpose.

Thank you @perexg  for review !

I kept only settings to on in _BootSequence_ so that all devices are enabled and audio works without any manual settings. Can you please explain why the settings to off is safer? 

And can you please clarify why we should disable all devices in Verb/EnableSequence? you mentioned that to **put controls to a safe initial state** , but if the initial state is **all off,** the audio will not work.

perexg:
> kept only settings to on in _BootSequence_ so that all devices are enabled and audio works without any manual settings. Can you please explain why the settings to off is safer?

For normal users, it's expected that pulseaudio or pipewire will be used (thus those servers will enable the specific UCM device/path) on boot later. Eventually, users can enable the specific devices using `alsaucm -c hw:0 set _verb HiFi _enadev Speaker` etc. command. The settings is saved and restored using alsactl state systemd units.
 
> And can you please clarify why we should disable all devices in Verb/EnableSequence? you mentioned that to **put controls to a safe initial state** , but if the initial state is **all off,** the audio will not work.

You should ensure to not duplicate the output/input to/from multiple devices (unless the pulseaudio/pipewire asks for it). Note that PA/PW nor UCM alsa-lib's code do not disable all devices when the verb is enabled. You should do it in the verb's EnableSequence to ensure that the application starts from a deterministic state.

Perhaps, for your hardware, the mixer paths are separate (hardware elements are not shared), but the second argument is power consumption. I expect that the enabled audio path costs something.

fchiby:
Thank you @perexg  for all these explanations, it is more clear for me now. 

However, the client doesn't have **alsactl** and doesn't want to enable devices manually. So I set to on Switch in _Bootsequence_, because otherwise Pulseaudio failed at the initialization. (even with _EnableSequence_ in each device).

perexg:
> Thank you @perexg for all these explanations, it is more clear for me now.
> 
> However, the client doesn't have **alsactl** and doesn't want to enable devices manually. So I set to on Switch in _Bootsequence_, because otherwise Pulseaudio failed at the initialization. (even with _EnableSequence_ in each device).

I guess that there's a driver issue then. The EnableSequence should be called and the controls set for the active device. You may check the state of controls using `amixer contents` for example.

fchiby:
>I guess that there's a driver issue then. The EnableSequence should be called and the controls set for the active device. You may check the state of controls using amixer contents for example.

Hello @perexg, 

 It seems that we need to enable all devices before starting PA. In our driver we use PCM, which means that we have to route Front End to a BackEnd and then only We can open a device. Without linking to BE it will fail. 
     
here are the commands I used to call _EnableSequence_ in each device :
     
     ```
     alsaucm  -c hw:0  set _verb HiFi set _enadev HDMI
     alsaucm  -c hw:0  set _verb HiFi set _enadev Headphones
     alsaucm  -c hw:0  set _verb HiFi set _enadev Mic1
     alsaucm  -c hw:0  set _verb HiFi set _enadev Line1
     alsaucm  -c hw:0  set _verb HiFi set _enadev Line2
     alsaucm  -c hw:0  set _verb HiFi set _enadev Mic2 

```  
     
     However, this only works if I don't add "disdevall" in **HiFi.conf**. Otherwise, I got this error: 
 ```

 i1200-demo:~$  alsaucm  -c hw:0  set _verb HiFi set _enadev Headphones
ALSA lib ../../../alsa-lib-1.2.6.1/src/ucm/parser.c:937:(parse_sequence) error: sequence command 'disdevall' is ignored
```

perexg:
> > I guess that there's a driver issue then. The EnableSequence should be called and the controls set for the active device. You may check the state of controls using amixer contents for example.
> 
> Hello @perexg,
> 
> It seems that we need to enable all devices before starting PA. In our driver we use PCM, which means that we have to route Front End to a BackEnd and then only We can open a device. Without linking to BE it will fail.

Then it explains the on/off calls in the original BootSequence. Could you keep them and add a comment, why they are required (to let open the PCM devices by PA)?

> here are the commands I used to call _EnableSequence_ in each device :
> 
> ```
>      alsaucm  -c hw:0  set _verb HiFi set _enadev HDMI
>      alsaucm  -c hw:0  set _verb HiFi set _enadev Headphones
>      alsaucm  -c hw:0  set _verb HiFi set _enadev Mic1
>      alsaucm  -c hw:0  set _verb HiFi set _enadev Line1
>      alsaucm  -c hw:0  set _verb HiFi set _enadev Line2
>      alsaucm  -c hw:0  set _verb HiFi set _enadev Mic2
> ```
> 
>     * However, this only works if I don't add "disdevall" in **HiFi.conf**. Otherwise, I got this error:
> 
> 
> ```
> i1200-demo:~$  alsaucm  -c hw:0  set _verb HiFi set _enadev Headphones
> ALSA lib ../../../alsa-lib-1.2.6.1/src/ucm/parser.c:937:(parse_sequence) error: sequence command 'disdevall' is ignored
> ```

It's supported in alsa-lib 1.2.7+. For older alsa-lib, this command has no effect.

fchiby:
Hello @perexg, 

Can you please help with merge if everything seems to you good. 
Thanks a bunch!

fchiby:
Hello @perexg, 

Sorry for disturbing you. The client really need to use these MR, can you please help with merge as soon as you can :). 
Thanks a lot !

berolinux:
Revised patch LGTM

ucm2/MediaTek/mt8195_demo/mt8195_demo.conf
```diff
@@ -16,12 +16,12 @@ BootSequence [
 	cset "name='PGA_L_Mux' AIN1"
 	cset "name='HDMI_OUT_MUX' Connect"
 	cset "name='DPTX_OUT_MUX' Connect"
+	# we need to enable all devices before starting PA.
+	# In our driver we use PCM, which means that we have
+	# to route Front End to a BackEnd and then only we can
+	# open a device. Without linking to BE it will fail.
```
perexg, line 22: If you turn a specific UCM device off, can it be re-opened (is the route between frontend / backend preserved)? I mean if the 'off' in DisableSequence for devices are ok or not.

fchiby: If I turn a specific UCM device off in` mt8195_demo.conf ` I can re-opened the device using     ` alsaucm  -c hw:0  set _verb HiFi set _enadev ` .
(This only works if I don't add "disdevall" in **HiFi.conf** )

perexg: It makes me think that the enable/disable logic is completely broken. It should be allowed:

`set _verb HiFi`
`set _enadev <some_dev>`
`set _enadev <other_dev>`
... the associated PCM devices are availabe for both UCM devices ..
`set _disdev <some_dev>`
... the associated PCM devices are availabe for <other_dev> UCM device ..
`set _ena_dev <some_dev>`
`set _verb Inactive`    # disable verb
`set _verb HiFi` ...

perexg: I don't follow you. The sequences should be executed. Just run the suggested tests and check the PCM availability (using any ALSA native tool like `aplay/arecord`). The ALSA control elements may be monitored using `amixer events` or `alsactl monitor` commands.

fchiby: Here my tests results: 
```

i1200-demo:~$ set _verb HiFi
i1200-demo:~$ set _enadev Headphones
i1200-demo:~$ set _enadev HDMI      
i1200-demo:~$ set _disdev Headphones
i1200-demo:~$ set _enadev Headphones
i1200-demo:~$ set _verb Inactive
i1200-demo:~$ set _verb HiFi
i1200-demo:~$ arecord -f S16_LE test.wav
Recording WAVE 'test.wav' : Signed 16 bit Little Endian, Rate 8000 Hz, Mono
^CAborted by signal Interrupt...
i1200-demo:~$ aplay test.wav 
Playing WAVE 'test.wav' : Signed 16 bit Little Endian, Rate 8000 Hz, Mono

```

I recorded the sound successfully and was able to play it and listening to my recording successfully using my Headphones.

perexg:
1) boot state - the sound card usually boots to a state where the basic I/O may be enabled when no previous state exists
2) ucm verb activation - all output paths (devices) should be off, but PCM devices are ready for the use - PA requires this
3) the one or more ucm device may be activated or deactivated and corresponding mixer paths activated/deactivated
4) ucm verb deactivation - all active devices are moved to the disabled state (DisableSequence is executed)

Note that the PCM device availability is bound to the verb activation. If you modify the ALSA controls which modifies the ALSA PCM device availability (usability), you should do it only in the verb's enable/disable sequences.

ucm2/MediaTek/mt8195_demo/mt8195_demo.conf
perexg: You check pulseaudio PCM/mixer not ALSA PCM/mixer. The `amixer events` should be run simultaneously (all time) on another terminal - it's just for a confirmation that ucm set the right controls. Use '-D hw:0,X' device for arecord/aplay (like it's defined in the UCM config). The aplay/arecord tests should be executed after each 'ucm set' step to check the usability of the PCM devices.

perexg: Also note that UCM expects to use those PCM devices:

```
alsa-ucm-conf]$ grep PCM ucm2/MediaTek/mt8195_demo/*.conf
ucm2/MediaTek/mt8195_demo/HiFi.conf:		PlaybackPCM "hw:${CardId},5"
ucm2/MediaTek/mt8195_demo/HiFi.conf:		PlaybackPCM "hw:${CardId},0"
ucm2/MediaTek/mt8195_demo/HiFi.conf:		CapturePCM "hw:${CardId},10"
ucm2/MediaTek/mt8195_demo/HiFi.conf:		CapturePCM "hw:${CardId},11"
ucm2/MediaTek/mt8195_demo/HiFi.conf:		PlaybackPCM "hw:${CardId},6"
ucm2/MediaTek/mt8195_demo/HiFi.conf:		CapturePCM "hw:${CardId},14"
```

fchiby: I checked PCM devices (exept hw:0,5 because I don't have a monitor)  after each ucm set (please check the attachment). 
[check PCM devices.txt](https://github.com/alsa-project/alsa-ucm-conf/files/10022976/check.PCM.devices.txt)

fchiby: Hello @perexg, 
I rechecked the usability  of all PCM  devices  and it looks good for me. Can you please confirm? I put the results in the attachment bellow. 
[PCM devices  usability.txt](https://github.com/alsa-project/alsa-ucm-conf/files/10032715/PCM.devices.usability.txt)

Thanks for your support !

perexg: Which controls makes the PCM devices unavailable ?

fchiby: Hello @perexg,


>  Which controls make the PCM devices unavailable ?
Controls which set to off in DisableSequence.  If we take as an exemple Headphones, **'O176 I070 Switch' and 'O177 I071 Switch'** makes PCM device **hw:0,0** unavailable if they are setting to OFF. 
  
FYI, I set the Switch to OFF using amixer ( amixer sset -c0 <switch-id> off), because when calling DisableSequence the stat of the Switch does not change (it's like ` alsaucm  -c hw:0  set _verb HiFi set _disdev <device_name>`  does nothing).

perexg: DisableSequence is set only for the active (previously enabled devices). So you should use _enadev and _disdev...

perexg: > Controls which set to off in DisableSequence. If we take as an exemple Headphones, 'O176 I070 Switch' and 'O177 I071 Switch' makes PCM device hw:0,0 unavailable if they are setting to OFF.

So with this change, the UCM application won't be able to use the PCM device next time. Basically, because there are no ConflictingDevices in your UCM configuration, PA opens all devices together, so the problem is not visible.  I think that the controls which affects the PCM device availability should be set only in verb's Enable/Disable sequences to make sure that they are live for the whole time when the UCM verb is used.

perexg: So my recommendation is:

1) remove off from the BootSequence
2) add all PCM related controls also to the verb sequences (Enable/Disable)
3) do not use PCM related controls in the device configuration blocks

fchiby: Thanks @perexg !

PR #218, opened 2022-09-23
ucm2: mt8195-sof: Fix wrong JackControl for headphone

nfraprado:
The value set for the headphone's JackControl doesn't match any jack control available. Change it to the right one.

Signed-off-by: Nícolas F. R. A. Prado <nfraprado@collabora.com>

This jack control isn't actually exposed by the driver yet, but the patch doing that is on the mailing list for review [1].

[1] https://lore.kernel.org/all/20220922235951.252532-3-nfraprado@collabora.com/


kholk:
With the kernel part picked, I agree with this change, so you get my:

`Tested-by: AngeloGioacchino Del Regno <angelogioacchino.delregno@collabora.com>`

Cheers!

Issue #220, opened 2022-09-30
i have some technical issues with the Arturia minifuse 1 sound card.  i would like help with converting the minifuse 2 UCM config to work with minifuse 1


Ali32bit:
greetings.        so i have a sound card with multiple outputs and some loop back channels.  it is USB complaint but those extra bells and whistles  cause ALSA to think its a surround sound device when its actually just stereo.  and this actually impacts the sound when i play games on it cause sounds from behind are basically muted.   and thats the case with some music too but less obvious
here is what alsa_info generated for me too : 

http://alsa-project.org/db/?f=b6db7723ffa65f32b93297109bf1f8e4d853cf4e

there should already be a minifuse 2 setup so if possible please adapt it to minifuse 1 or help me convert and install it myself.     the hardware is mostly the same but it has 1 input instead of 2  and no extra midi ports in the back.

nodens:
Please try my branch from the PR above, which is my attempt at adding such support.
The idea here is that from UCM point of view, it's exactly the same as Minifuse 2 with one less input (and so no physical stereo input, but there is still a stereo loopback).

nodens:
Note: I've been in touch with @Ali32bit on Reddit, and they seem to have issues testing my fix, so:

Some more info to make it easier for you to test, also because it might help other people getting here after a web search:
- get my branch from https://github.com/nodens/alsa-ucm-conf/tree/add_arturia_minifuse_1_use_case
- copy the `ucm2` tree to your system (probably `/usr/share/alsa/ucm2`, and of course it's probably better to make a backup beforehand, but it's easy to get back to your system version by removing the ucm2 folder on your system and reinstall the package containing it: on Debian/ubuntu-like systems it's named `alsa-ucm-conf`). See also information on the `README.md` file.
- unplug / replug your minifuse
- check with pavucontrol if you see the correct profiles in the configuration tab (or equivalent).

If you don't see any change, you can try to restart your sound daemons (eg pulseaudio or pipewire + pipewire-pulse + wireplumber), or even reboot your system to be sure, and we'll check from there…  It's totally possible I forgot something :)

nodens:
Either my PR is not working, or maybe pulseaudio/pipewire config is interfering… 

What do you see in pavucontrol in the configuration tab ? For each device, you should be able to set a profile, what are they for the minifuse?

To see what's going on, other helpful stuff:

- run `alsaucm reload` and check for errors (including with `journalctl` / `journalctl --user`)
[...]

PR #226, opened 2022-10-06
Draft: Add Arturia Minifuse 1 Support

nodens:
Arturia Minifuse 1 is a stripped-down version of Minifuse 2:
- 1 Input
- 2 Outputs
- Loopback interface to ease recording of the computer output

(attempt at fixing #220)


nodens:
This should work now (only tested with Minifuse 2 hardware though).

PR #228, opened 2022-10-11
If system has Dmic, variable DeviceDmic will be empty

xanthein:
change condition from False to True to configure Dmic0 correctly


xanthein:
@perexg Could you help take a look of this pull request

PR #203, opened 2022-08-23
Adding Focusrite Scarlett 2i4 gen2

PatrickLang:
This cleans up the Focusrite Scarlett 2i4 Gen2 device a bit. This device is split into two stereo pairs for output, and has 2 mic/line/instrument inputs. There are no toggles or mixer controls exposed to Alsa.

It has hardware switches for:
- Headphones - can monitor line outputs 1+2 or 3+4
- Headphone input/playback mix
- Monitor out (Line 1+2)
- Each input has a level pot, and switches for pad and Z (line/inst)
- There is 1 physical +48v switch

Please let me know if any more info would be helpful. I have captured some more notes showing before/after state at https://gist.github.com/PatrickLang/79a3090eea6f2d640179457b5d428134#scarlett-2i4-2nd-gen---initial-state

I'm planning to write a UCM2 profile for the more complicated Focusrite Scarlett 18i8 Gen2 next which has more inputs, but also multiple mix busses controllable through Alsa

PR #234, opened 2022-10-27
Add Focusrite Scarlett 2i2 gen2

superseed:
This is basically #203 but for the 2i2.

I don't really know what I'm doing, but with these changes, I get two separate mono inputs in Pipewire, which is what I was looking for.


perexg:
Recoded in 6dee56f11fbd48dd412179722a445416fcd779ef . Thanks for you PR.

PR #235, opened 2022-10-31
feat: add support for ASUS ROG Strix Z790-E Gaming Wifi

DASPRiD:
I've verified this profile on my motherboard, and both back- and front jacks seem to work fine, including SPDIF.


DASPRiD:
Just wondering, when is the next release planned? The last one was October 2022. I had to lock the ucm package on my distribution and manually replace the config files, because distribution updates would constantly override them.

PR #242, opened 2022-11-16
ucm2: sof-soundwire: add basic settings for RT1318 SDCA device

shumingfan:
Add support for rt1318 amplifier

Signed-off-by: Shuming Fan <shumingf@realtek.com>

### DIFF
--- a/ucm2/sof-soundwire/rt1318-1.conf
+++ b/ucm2/sof-soundwire/rt1318-1.conf
@@ -1,0 +1,20 @@
+# Use case Configuration for sof-soundwire card
+
+SectionDevice."Speaker" {
+	Comment	"Speaker"
+
+	EnableSequence [
+		cset "name='rt1318-1 DAC Switch' 1"
+		cset "name='Speaker Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='rt1318-1 DAC Switch' 0"
+		cset "name='Speaker Switch' off"
+	]
+
+	Value {
+	      PlaybackPriority 100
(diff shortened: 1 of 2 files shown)


perexg:
It seems fine, but I would like to know, if all of those amplifier chips are really missing the analog volume control in the hardware. If there's analog volume control, it should be exposed in the kernel driver to the user space and described in the UCM configuration.

plbossart:
> It seems fine, but I would like to know, if all of those amplifier chips are really missing the analog volume control in the hardware. If there's analog volume control, it should be exposed in the kernel driver to the user space and described in the UCM configuration.

Even if they did, we have a plumbing issue @perexg. In the case where there are two amplifiers, we would need to have two volumes applied at the same time in two different hardware devices. This is not possible with older SoundWire devices, but with newer SDCA-based devices there are means to synchronize volume changes across related devices. It'll require a lot more work to implement this I am afraid, for now only the PCM is synchronized.

perexg:
I just feel that we miss something which was a "hard" standard before (analog volume control). It may affect also the power consumption of the amplifier. Having more components is just an implementation detail. We can do a remap for those devices in the user space, but the drivers should expose this functionality (if hw is capable, of course).

plbossart:
oh I completely agree with you @perexg that this is something we need to work on. 

But as you said we need to create a 'virtual' control that relays the user choices to separate devices and when possible we have to synchronize the control .put sequences. Our Windows friends are having the same issue btw, when the 'amplifier function' is implemented with different devices new plumbing is required across the stack.

PR #243, opened 2022-11-17
Revert "ucm2: add profile for the Librem 5"

craftyguy:
This reverts commit 998849d9510754960d808e20dad4c440adaf87ef.

The oem (Purism) is working on some big improvements to this config, which I knew of when I submitted the patch this reverts, however I didn't realize until now that their changes would break config.

It's probably best to just revert this so it doesn't make it into a release and wait for the oem to submit their improved ucm upstream.


perexg:
We can update the configuration when we have a better one. I cannot do any arbitration here.

PR #245, opened 2022-11-30
USB-Audio: Update quirk for Aorus Pro WiFi Rev 1.2

13r0ck:
An improvement for https://github.com/alsa-project/alsa-ucm-conf/pull/157

The line in was using the wrong device.

### DIFF
--- a/ucm2/USB-Audio/Realtek/ALC1220-VB-Desktop-HiFi.conf
+++ b/ucm2/USB-Audio/Realtek/ALC1220-VB-Desktop-HiFi.conf
@@ -1,5 +1,6 @@
 Define.Mic1Dev 1
 Define.Mic2Dev 2
+Define.LineDev 0
 
 If.trx4-aorus-wifi-rev-1-2 {
 	Condition {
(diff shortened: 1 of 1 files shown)


perexg:
This change is not aligned to Mic1Dev and Mic2Dev. Also, device 1 is used for Mic2. You should reshuffle all clashing devices.

13r0ck:
Sorry I'm not sure I understand. Are you saying that the quirk in the kernel needs to be reshuffled?

perexg:
With your change Mic2 UCM device and Line UCM device uses the PCM device number 1 . Is this an thoughtful change ?  Could you comment ? Also, use the number for LineDev like for other variables. The empty string is equal to device number zero.

perexg:
> Well I sure thought it was, but I tried the patch without those added and it still works. I went ahead an removed those from the commit.

It does not resolve things. The defaults are used (on top of the file) plus your change will result with:

```
Define.Mic1Dev 1
Define.Mic2Dev 2
Define.LineDev 1
```

So clash is moved from Mic2 to Mic1. Have you tested the microphone inputs?

13r0ck:
Ah I understand now. Thanks for the clarification. I tested mic and line with the new commit, and they both are detected and record now. Hot plug when switching jacks sometimes doesn't detect sound, but I don't believe that is related to this.

13r0ck:
@perexg thank you for reviewing my other PR's, any update on this?

13r0ck:
Thank you!

PR #250, opened 2022-12-14
ucm2: add profile for the Samsung Galaxy Note II (Midas)

FrostI7Alex:
Might need few changes in the future, but for now - it works.


Newbytee (commented):
It's nice that you do this, but how does this work given that Galaxy S III uses the same "Midas" name? Is it able to avoid picking this one for S III somehow?

jenneron:
@Newbytee 

> how does this work given that Galaxy S III uses the same "Midas" name? 

The same way. It will match "Midas" and use these configs

> Is it able to avoid picking this one for S III somehow?

Yes, but they have to use different name

It can be overwritten in dts https://github.com/torvalds/linux/blob/master/arch/arm/boot/dts/exynos4412-midas.dtsi#L293

PR #254, opened 2022-12-20
USB-Audio: alc4080 - add MSI MAG B650I Edge WiFi (ID 0db0:36e7)

13r0ck:


liayn:
It's actually called "MSI **MPG** B650I Edge WIFI"
Not sure if that matters.

PR #258, opened 2022-12-24
ucm2: sof-essx8336: don't use disdevall for this device

mchehab:
disdevall causes undesired side effects on this device, as it breaks main microphone's detection, and makes hdmi outputs invisible at pavucontrol (tested with pipewire-pulse).

So, rely on cset to disable what's not needed at the enable sequence.

Fixes: 15a12508003c ("ucm2: put disdevall to HDA/Soundwire configs")
Signed-off-by: Mauro Carvalho Chehab <mchehab@kernel.org>

### DIFF
--- a/ucm2/Intel/sof-essx8336/HiFi.conf
+++ b/ucm2/Intel/sof-essx8336/HiFi.conf
@@ -1,8 +1,7 @@
 SectionVerb {
 	EnableSequence [
-		disdevall ""
-		# Disable all inputs / outputs
-		#   (may be duplicated with disdevall)
+		# Using disdevall cause troubles here, as it
+		# breaks microphone detection.
 		cset "name='Headphone Switch' off"
 		cset "name='Headset Mic Switch' off"
 		cset "name='Internal Mic Switch' off"


perexg:
I think that the problem may be somewhere else. When I track the commands in the DisableSequence for devices:

```
Analog Mic:               cset "name='Internal Mic Switch' off"
Speaker:                  cset "name='Speaker Switch' off"
Headset:                  cset "name='Headset Switch' on"
HDMI devices:             cset "name='IEC958 Playback Switch',index=${eval:($__Number-1)} off"
```

The headset sequence looks suspicious and 'Speaker Switch' initialization is not in the verb's Enable sequence.

For HDMI - those switches should not affect the PCM device accessibility. Could you gather any logs from wireplumber to see more details, why those devices are not detected ?

ChronosX88:
Is there a way to enable internal microphone? Speaker works, but micro is not, unfortunately :(

perexg:
May be fixed with PR #328 . Closing, reopen if the issue persists.

PR #259, opened 2022-12-26
sof-hda-dsp: Set Dmic0 Capture Switch on

KaiChuan-Hsieh:
Internal microphone default is off after fresh installation. Add operation to set the control on to align with other sound architecture.


KaiChuan-Hsieh:
@perexg 

Hello,

Could you help to check if it is a valid change?

Thanks,

Issue #251, opened 2022-12-18
Dell XPS 9320: Mic-Mute LED not working


heftig:
The mic-mute LED on my XPS 9320 does not work. Reading `/sys/class/sound/ctl-led/mic/card0/list` outputs `52 18`.

Control 18 is the expected `rt714 FU02 Capture Switch` set in `ucm2/codecs/rt715-sdca/init.conf`. Control 52 is:

```
numid=52,iface=MIXER,name='PGA5.0 5 Master Capture Switch'
  ; type=BOOLEAN,access=rw------,values=2
  : values=on,on
```

In this state, the LED does turn on when both switches are off, but muting via PipeWire only toggles control 18.

Detaching the LED from `52` (so that only `18` remains) makes it work properly.

I've tried finding out how 52 gets attached, but couldn't find anything. Is this done in-kernel?

Linux 6.1
ALSA 1.2.8
Arch Linux

perexg:
Already fixed in 79a8ec44d3dcf097f4a4492c506cbcf338324175 .

heftig:
Is there a release scheduled which contains the fix?

PR #261, opened 2022-12-28
ucm2: sof-soundwire: Update Mic LED settings

andychi117:
Test on XPS 9320 with Ubuntu Jammy 22.04.
[kernel]
6.0.0-1009-oem

After apply this patch, mic mute led works.

Signed-off-by: Andy Chi <andy.chi@canonical.com>

### DIFF
--- a/ucm2/sof-soundwire/sof-soundwire.conf
+++ b/ucm2/sof-soundwire/sof-soundwire.conf
@@ -70,5 +70,6 @@ If.mic_init_rt715 {
 	}
 	True.BootSequence [
 		cset "name='PGA5.0 5 Master Capture Switch' 1"
+		sysw "-/class/sound/ctl-led/mic/card${CardNumber}/detach:PGA5.0 5 Master Capture Switch"
 	]
 }


andychi117:
Tried this solution on my XPS 9320, but it doesn't work.
https://github.com/alsa-project/alsa-ucm-conf/issues/251

andychi117:
Hello @perexg , can you help to take a look on this MR? Thanks.

perexg:
Could you show me the list  in `/sys/class/sound/ctl-led/mic/card0/list` ?  Also, please, analyze/show the control names (`amixer -c 0 controls` command) - numid is the identifier.

andychi117:
> Could you show me the list in `/sys/class/sound/ctl-led/mic/card0/list` ? Also, please, analyze/show the control names (`amixer -c 0 controls` command) - numid is the identifier.

Yes, below are the related information.
```
$ cat /sys/class/sound/ctl-led/mic/card0/list
18 52
```

```
numid=52,iface=MIXER,name='PGA5.0 5 Master Capture Switch'
numid=18,iface=MIXER,name='rt714 FU02 Capture Switch'
```

plbossart:
@perexg I am not following or understanding what makes you think we set the LEDs in the SOF topology file. That may have been true for the Intel DMIC case, but in the case of the SoundWire devices the topology file has no knowledge of what codec is connected and whether or not an LED is used.

ucm2/sof-soundwire/sof-soundwire.conf
```diff
@@ -70,5 +70,6 @@ If.mic_init_rt715 {
 	}
 	True.BootSequence [
 		cset "name='PGA5.0 5 Master Capture Switch' 1"
+		sysw "-/class/sound/ctl-led/mic/card${CardNumber}/detach:PGA5.0 5 Master Capture Switch"
```
plbossart, line 73: this part of UCM seems to be wrong, I don't understand why we would use the SOF firmware controls for muting. This should be using the RT715 controls.

The ONLY case where SOF firmware controls should be used is for PCH-attached DMIC - because there is no codec involved. All other cases should be handled by the codecs.

perexg:
@plbossart : We are talking about mic flag for `PGA5.0 5 Master Capture Switch` . I believe that this is handled through the topology files. Also, the topology filenames contains rt714 components (like `sof-adl-rt711-l0-rt1316-l12-rt714-l3`), so I assume that topology files have some knowledge what's behind DSP.

In the current ALSA code:

```
/core/control_led.c:		ledtrig_audio_set(led->trigger_type, route ? LED_OFF : LED_ON);

/pci/hda/hda_generic.c:		knew->access |= SNDRV_CTL_ELEM_ACCESS_MIC_LED;

/soc/sof/topology.c:		mask = scontrol->led_ctl.direction ? SNDRV_CTL_ELEM_ACCESS_MIC_LED :
```

So it appears that the sof/topology.c is the only place where both mic flags are set.

plbossart:
Ah yes, there's something wrong here. I see this in the conf

````
SectionVendorTuples."CONTROLMIXER5.5_tuples_w" {
	tokens "sof_led_tokens"
	tuples."word" {
		SOF_TKN_MUTE_LED_USE		"1"
		SOF_TKN_MUTE_LED_DIRECTION	"1"
	}
}
````

this is just wrong and likely a combination of m4 being so difficult to read and copy/paste.

@ranj063, do you have any idea why those tokens are set in SoundWire topologies?

plbossart:
> @plbossart : We are talking about mic flag for `PGA5.0 5 Master Capture Switch` . I believe that this is handled through the topology files. 

That part is correct but it's a mistake.

> Also, the topology filenames contains rt714 components (like `sof-adl-rt711-l0-rt1316-l12-rt714-l3`), so I assume that topology files have some knowledge what's behind DSP.

That part is not correct. In fact we use the same topology definitions for all amplifiers and on the capture side we have no idea if the mic codec is rt715 or rt714-sdca. What is defined is a generic set of pipelines on capture that start on the 'ALH' hardware block. 

In other words, the topology is just DSP centric and CANNOT define what controls to use for mute or volume on the codec side, this is a platform-level information that's way beyond what the topology was designed for.

bardliao:
> Ah yes, there's something wrong here. I see this in the conf
> 
> ```
> SectionVendorTuples."CONTROLMIXER5.5_tuples_w" {
> 	tokens "sof_led_tokens"
> 	tuples."word" {
> 		SOF_TKN_MUTE_LED_USE		"1"
> 		SOF_TKN_MUTE_LED_DIRECTION	"1"
> 	}
> }
> ```
> 
> this is just wrong and likely a combination of m4 being so difficult to read and copy/paste.
> 
> @ranj063 @bardliao, do you have any idea why those tokens are set in SoundWire topologies?

No, I have no idea, sorry. But it must be for the mute led feature.

ranj063:
> 
This was implemented by @juimonen . He should have more details

perexg:
I put this setup to UCM, but the driver defaults should be fixed, too.

PR #273, opened 2023-02-03
ucm: USB-Audio - Add support for Focusrite Scarlett 2i2 gen3

anarsoul:
Signed-off-by: Vasily Khoruzhick <anarsoul@gmail.com>


anarsoul:
[alsa-info.txt](https://github.com/alsa-project/alsa-ucm-conf/files/10576970/alsa-info.txt)

alsa-info.txt is attached.

ucm validator currently fails on master, so I just tested it locally with `alsaucm -c hw:2 dump text`

PR #287, opened 2023-03-08
Use case for the Arturia Minifuse 4 Audiointerface 

bernsnf:
Use case for the Arturia Minifuse 4 Audiointerface 
https://www.arturia.com/products/audio/minifuse/minifuse4

Based on the existing Minifuse 2 Profile.
#145 

I created a separate file because of the additional input- and output-Pairs result in different loopback channels.

Without the usecase outputs threated as surround channels.

### DIFF
--- a/ucm2/USB-Audio/Arturia/Minifuse-4-HiFi.conf
+++ b/ucm2/USB-Audio/Arturia/Minifuse-4-HiFi.conf
@@ -1,0 +1,180 @@
+Include.pcm_split.File "/common/pcm/split.conf"
+Macro [
+	{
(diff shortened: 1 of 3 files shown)


bernsnf:
alsa-info output
http://alsa-project.org/db/?f=7c09a721901f357d4d5a31d26263818dc7813f5e

perexg (changes requested):
It looks fine, but please, provide little corrections:

- add Signed-off-by line with your e-mail
- remove whitespaces (use tabelators - see `git diff --check`)

PR #289, opened 2023-03-11
USB-Audio: ALC4080 - Add MSI MAG B650M MORTAR WIFI

cloudy-dev:
USB ID: 0db0:7696

Adds support for the MSI MAG B650M MORTAR WIFI motherboard front & back 3.5mm jacks.


XeCycle:
I have this exact board, and TOSLINK does not work.  How can I contribute?

PR #302, opened 2023-04-10
[V2] sof-hda-dsp: Add speaker led support

Kobatw:
Replace hard code with variable.
before: sysw "-/class/sound/ctl-led/speaker/card1/attach:Master Playback Switch"
after:    sysw "-/class/sound/ctl-led/speaker/card${CardNumber}/attach:Master Playback Switch"

### DIFF
--- a/ucm2/Intel/sof-hda-dsp/sof-hda-dsp.conf
+++ b/ucm2/Intel/sof-hda-dsp/sof-hda-dsp.conf
@@ -61,6 +61,19 @@ If.speaker {
 	]
 }
 
+If.DellMuteLed {
+	Condition {
+                Type String
+                Haystack "$${sys:class/leds/platform::mute/device}"
+                Needle "dell-laptop"
+        }
+        True {
+                FixedBootSequence [
+                        sysw "-/class/sound/ctl-led/speaker/card${CardNumber}/attach:Master Playback Switch"
+                ]
+        }
+}
+
 If.headphone {
 	Condition {
 		Type ControlExists


ucm2/Intel/sof-hda-dsp/sof-hda-dsp.conf
```diff
@@ -61,6 +61,16 @@ If.speaker {
 	]
 }
 
+If.speakersw {
+        Condition {
+                Type ControlExists
+                Control "name='Master Playback Switch'"
```
plbossart, line 67: it's a bit inconsistent with the unconditional use of 'Master Playback Switch' line 50

Kobatw: does it look better if the sysw move to line 50 and remove If.speakersw block?
```
If.master {                                                                      
        Condition {                                                              
                Type ControlExists                                               
                Control "name='Master Playback Volume'"                          
        }                                                                        
        True.BootSequence [                                                      
                cset "name='Master Playback Volume' 60%"                         
                cset "name='Master Playback Switch' on"                          
                sysw "-/class/sound/ctl-led/speaker/card${CardNumber}/attach:Master Playback Switch"
        ]                                                                           
}  
```

plbossart: probably best to let @perexg comment, I don't remember the nuance between BootSequence and FixedBootSequence and whether the Volume always comes with a Switch.

perexg:
It's a bad idea to change this settings for all hardware. Usually, the speaker LED is preset in the HDA kernel driver. The HDA codec driver should call `snd_hda_gen_add_mute_led_cdev` function.

Kobatw:
@perexg as i known, the particular sku would trigger speaker-mute event through EC and is handled by dell-wmi subsystem.
the hda codec driver would not be notified and handle this.

perexg:
> @perexg as i known, the particular sku would trigger speaker-mute event through EC and is handled by dell-wmi subsystem. the hda codec driver would not be notified and handle this.

See https://github.com/torvalds/linux/blob/0d3eb744aed40ffce820cded61d7eac515199165/sound/pci/hda/thinkpad_helper.c as an example, how it's handled the LED trigger settings preset for Lenovo ACPI.

I'm curious where is located the Speaker LED driver for this Dell platform in the current kernel because `LED_AUDIO_MUTE` class is registered only in those drivers:

```
drivers/platform/x86/thinkpad_acpi.c:	[LED_AUDIO_MUTE] = {
drivers/platform/x86/thinkpad_acpi.c:	return tpacpi_led_set(LED_AUDIO_MUTE, brightness != LED_OFF);
drivers/platform/x86/thinkpad_acpi.c:	[LED_AUDIO_MUTE] = {
sound/pci/hda/hda_generic.c:	if (spec->led_cdevs[LED_AUDIO_MUTE])
sound/pci/hda/hda_generic.c:		led_classdev_unregister(spec->led_cdevs[LED_AUDIO_MUTE]);
sound/pci/hda/hda_generic.c:	int idx = micmute ? LED_AUDIO_MICMUTE : LED_AUDIO_MUTE;
```

perexg:
Could you check, what string is returned for `readlink /sys/class/leds/platform\:\:mute` command ?

In case of Lenovo ACPI, the result is:

```
../../devices/platform/thinkpad_acpi/leds/platform::mut
```

Kobatw:
I registered the speaker led control in dell-laptop drvier.
Ref. https://patchwork.kernel.org/project/platform-driver-x86/patch/20230308062414.1048913-1-koba.ko@canonical.com/

```
u@u-Dell-G16-7630:~$ readlink -v /sys/class/leds/platform\:\:mute
../../devices/platform/dell-laptop/leds/platform::mute
```

perexg:
Ok, if this should go through UCM, this condition may be better to catch all dell-wmi hardware:

```
        Condition {
                Type String
                Haystack "$${sys:class/leds/platform::mute}"
                Needle "../../devices/platform/dell-laptop/leds/platform::mute"
        }
```

Please, fix your commit and do force push (git -f).

@tiwai : FYI. Maybe you can comment, if you like to see this in the HDA driver rather than UCM. Note that the Mic LED is handled in the similar way, but I don't know, if legacy HDA devices (no SOF) are affected (LED trigger is missing) for Dell WMI.

perexg:
Oops, sorry only the last path component is returned for symlink (`platform::mute`) which is not so helpful. This reverse check should work:

```
       Condition {
                Type String
                Haystack "$${sys:devices/platform/dell-laptop/leds/platform::mute/device}"
                Needle "dell-laptop"
        }
```

Kobatw:
@perexg Sorry, i didn't notice "device"
After add "device" with needle "dell-laptop", it works.
Thanks

perexg:
Thanks. It looks fine. Let's wait for @tiwai, if he has a comment for the HDA driver.

perexg:
@Kobatw : You have a little problem in your kernel patch - the error path (`goto fail_led;`) does not handle situation where the mic led (`micmute_led_cdev`) is registered. It is not freed in the error path.

chinyakao:
Hi @perexg ,

Thank you so much for your comments and review. 
As we wait for the @tiwai, do you know how to contact him? or others who could review this PR?

> Thanks. It looks fine. Let's wait for @tiwai, if he has a comment for the HDA driver.

tiwai:
Sorry for the late comment, as I overlooked it.
I'm fine with the approach with UCM. If we need to put a hook in the HDA driver side for the specific models (with SSID), it can be added at any time later, too.

PR #309, opened 2023-05-03
ucm2: USB-Audio: Add support for Focusrite Scarlett Solo devices

kepstin:
These are basically identical to the Scarlett 2i2 devices of the same generation; the difference is that the first input can only be used for a mic, and the second input can only be used for line/instrument. (On the 2i devices, both inputs are the same - they both have mic preamps, and both support line/instrument input.)

I've added a new profile for them so the Comment can be set to the correct device name, but because of the similarity between the Solo and 2i devices, I've re-used the existing Scarlett-2i-HiFi use case to set up the split input devices.


kepstin:
At this point, I don't know enough about the syntax of the ucm files to make some of the changes you've requested - I've pretty much had to copy/paste from existing configs and modify them in ways that looked obvious.

Is there any documentation for the file format available?

perexg:
Documentation: https://www.alsa-project.org/alsa-doc/alsa-lib/group__ucm__conf.html

But things should be straight like:

```
diff --git a/ucm2/USB-Audio/Focusrite/Scarlett-2i-HiFi.conf b/ucm2/USB-Audio/Focusrite/Scarlett-2i-HiFi.conf
index ab69bc7..207bedd 100644
--- a/ucm2/USB-Audio/Focusrite/Scarlett-2i-HiFi.conf
+++ b/ucm2/USB-Audio/Focusrite/Scarlett-2i-HiFi.conf
@@ -69,7 +69,7 @@ If.scarlett_2i4 {
 	}
 }
 
-SectionDevice."Mic1" {
+SectionDevice."${var:Input1Device}" {
 	Comment "Input 1"
 
 	Value {
@@ -85,7 +85,7 @@ SectionDevice."Mic1" {
 	}
 }
 
-SectionDevice."Mic2" {
+SectionDevice."${var:Input2Device}" {
 	Comment "Input 2"
 
 	Value {
diff --git a/ucm2/USB-Audio/Focusrite/Scarlett-2i.conf b/ucm2/USB-Audio/Focusrite/Scarlett-2i.conf
index ff63971..4f712dc 100644
--- a/ucm2/USB-Audio/Focusrite/Scarlett-2i.conf
+++ b/ucm2/USB-Audio/Focusrite/Scarlett-2i.conf
@@ -1,7 +1,10 @@
 Define.PlaybackChannels 4
 Define.DirectPlaybackChannels 4
 Define.DirectCaptureChannels 2
+Define.Input1Device "Mic1"
+Define.Input2Device "Mic2"
 Define.Generation 2
+Define.Model "2i4"
 
 If.scarlett_2i2 {
 	Condition {
@@ -10,6 +13,7 @@ If.scarlett_2i2 {
 		Needle "USB1235:8202"
 	}
 	True.Define {
+		Model "2i2"
 		PlaybackChannels 2
 		DirectPlaybackChannels 2
 	}
@@ -22,13 +26,48 @@ If.scarlett_2i2_gen3 {
 		Needle "USB1235:8210"
 	}
 	True.Define {
+		Model "2i2"
 		PlaybackChannels 2
 		DirectPlaybackChannels 2
 		Generation 3
 	}
 }
 
-Comment "Focusrite Scarlett 2i${var:PlaybackChannels} Gen ${var:Generation}"
+If.scarlett_solo {
+	If.focusrite-scarlett-solo {
+	Condition {
+		Type RegexMatch
+		String "${CardComponents}"
+		Regex "USB1235:8(01c|205|211)"
+	}
+	True {
+		Define {
+			PlaybackChannels 2
+			DirectPlaybackChannels 2
+			Model "Solo"
+			Generation 1
+			Input2Device "Line3"
+		}
+		If.gen2 {
+			Condition {
+				Type String
+				Haystack "${CardComponents}"
+				Needle "USB1235:8205"
+			}
+			True.Define.Generation 2
+		}
+		If.gen3 {
+			Condition {
+				Type String
+				Haystack "${CardComponents}"
+				Needle "USB1235:8211"
+			}
+			True.Define.Generation 2
+		}
+	}
+}
+
+Comment "Focusrite Scarlett ${var:Model} Gen ${var:Generation}"
 
 SectionUseCase."HiFi" {
 	Comment "Default"
diff --git a/ucm2/USB-Audio/USB-Audio.conf b/ucm2/USB-Audio/USB-Audio.conf
index c128f1a..09a7fdc 100644
--- a/ucm2/USB-Audio/USB-Audio.conf
+++ b/ucm2/USB-Audio/USB-Audio.conf
@@ -194,7 +194,7 @@ If.focusrite-scarlett-2i {
 	Condition {
 		Type RegexMatch
 		String "${CardComponents}"
-		Regex "USB1235:82(0[02]|10)"
+		Regex "USB1235:(801c|82(0[025]|1[01]))"
 	}
 	True.Define {
 		ProfileName "Focusrite/Scarlett-2i"

```

kepstin:
If you're gonna be renaming the input devices anyways, then the names "Mic1" and "Mic2" on the 2i devices don't make sense. On those devices, both inputs are Mic/Line/Instrument inputs, with selection between Mic and Line/Instrument being performed automatically depending on what type of plug is connected.

On both the 2i and Solo devices, the Line/Instrument inputs have a hardware mode selection switch to choose between Line level and Instrument (with preamp).

Would the names "Input1" and "Input2" be acceptable? That matches how the inputs are described in the product user manual, and removes any reference to a specific device connection on the multi-purpose inputs.

kepstin:
Ah, to respond to myself after reading the docs, it looks like the names have to correspond to the `SND_USE_CASE_DEV_*` constants from https://www.alsa-project.org/alsa-doc/alsa-lib/group__ucm.html ?

In that case, there really is no appropriate name in the list for either the inputs (which are multipurpose and can be used as "Mic", "Line", or "Instrument" - a function which doesn't really fit anything in the list) or the outputs (on the 2-channel devices the stereo output is sent to both Line and Headphone simultaneously, and on the 4-channel devices both stereo outputs go to Line, and there's a hardware switch to select one of the stereo outputs to also be sent to Headphone.

perexg:
Yes, I forgot that we cannot detect the functionality for those devices which is hidden in the hardware. PC just handles the PCM stream. Keep the Mic1/Mic2 for the dual capable inputs. The `Comment` may be also altered to follow the hardware capabilities.

kepstin:
Changes have been made based on feedback. The configuration for "solo" devices has been merged with the config for the "2i2" and "2i4" devices. The conditionals used for the device naming and configuring properties have been re-arranged so they're easier to follow, and the "Comment" is now set dynamically on the various playback/capture devices depending on capabilities.

There's a couple of remaining questions that I have:

* Currently, a 4-channel "dshare" device is being set up for playback on the 2i2 and solo devices which only do 2-channel output. Is this a problem? Avoiding this is possible, but requires either a big conditional block or splitting out a separate verb file for the 2i4 devices.
* The input devices have been left as "Mic". There doesn't seem to be much consistency on what type to use on this sort of audio interface; calling them "Line" would probably also be ok.

perexg:
It seems that the output should be 2ch only - see #329 and #314 . So something like this may work for Line1:

```
SectionDevice."Line1" {
        Comment "${var:Playback1Name}"

        Value {
                PlaybackPriority 200
        }
        If.2i4 {
                Condition {
                        Type String
                        String1 "${var:PlaybackChannels}"
                        String2 "4"
                }       
                True {
                        Macro.pcm_split.SplitPCMDevice {
                                Name "scarlett2i_stereo_out"
                                Direction Playback
                                HWChannels 4
                                Channels 2
                                Channel0 0
                                Channel1 1
                                ChannelPos0 FL
                                ChannelPos1 FR
                        }
                }
                False.Value {
                        PlaybackPCM "hw:${CardId}"
                }
        }
}
```

The Mic device name is probably ok. The multi-purpose feature is described in the comment string.

PR #316, opened 2023-05-24
ucm2: Rockchip: Add UCM support for ES8316 on Rock 5B

cristicc:
Add UCM configuration for the RK3588 SoC based Rock 5B board to enable the analog audio support provided by the Everest Semi ES8316 codec.

### DIFF
--- a/ucm2/Rockchip/rk3588-es8316/HiFi.conf
+++ b/ucm2/Rockchip/rk3588-es8316/HiFi.conf
@@ -1,0 +1,42 @@
+SectionDevice."Headphones" {
+	Comment "Headphones"
+
+	EnableSequence [
+		cset "name='Playback Polarity' 'Normal'"
+	]
+
+	DisableSequence [
+		cset "name='Playback Polarity' 'R Invert'"
+	]
+
+	Value {
+		PlaybackPriority 200
+		PlaybackPCM "hw:${CardId}"
+		PlaybackMixerElem "Headphone Mixer"
+		PlaybackMasterElem "DAC"
+		JackControl "Headphones Jack"
+	}
+}
+
+SectionDevice."Headset" {
+	Comment "Headset Microphone on IN2"
+
+	EnableSequence [
+		cset "name='Differential Mux' lin2-rin2"
+		cset "name='Left Headphone Mux' lin2-rin2"
+		cset "name='Right Headphone Mux' lin2-rin2"
+	]
+
+	DisableSequence [
+		cset "name='Differential Mux' lin1-rin1"
+		cset "name='Left Headphone Mux' lin1-rin1"
+		cset "name='Right Headphone Mux' lin1-rin1"
+	]
+
+	Value {
+		CapturePriority 200
+		CapturePCM "hw:${CardId}"
+		CaptureMixerElem "ADC PGA Gain"
+		CaptureMasterElem "ADC"
+	}
+}
--- a/ucm2/Rockchip/rk3588-es8316/rk3588-es8316.conf
+++ b/ucm2/Rockchip/rk3588-es8316/rk3588-es8316.conf
@@ -1,0 +1,26 @@
+Syntax 3
+
+SectionUseCase."HiFi" {
+	File "/Rockchip/rk3588-es8316/HiFi.conf"
+	Comment "Play HiFi quality Music"
+}
+
+BootSequence [
+	# Set HP vol to 0 dB (3/3)
+	cset "name='Headphone Playback Volume' 3"
+	# Set HP mixer vol to 0 dB
+	cset "name='Headphone Mixer Volume' 11"
+	# Set DAC vol to 0 dB (192/192)
+	cset "name='DAC Playback Volume' 192"
+
+	# Disable Auto Level Control
+	cset "name='ALC Capture Switch' off"
+	# Set ADC vol to 0 dB (192/192)
+	cset "name='ADC Capture Volume' 192"
+	# Set Mic amplifier to +16 dB
+	cset "name='ADC PGA Gain Volume' 7"
+
+	# Setup muxes / switches
+	cset "name='Left Headphone Mixer Left DAC Switch' on"
+	cset "name='Right Headphone Mixer Right DAC Switch' on"
+]
--- a/ucm2/conf.d/rk3588-es8316/rk3588-es8316.conf
+++ b/ucm2/conf.d/rk3588-es8316/rk3588-es8316.conf
@@ -1,0 +1,1 @@
+../../Rockchip/rk3588-es8316/rk3588-es8316.conf


perexg:
The `Analog` driver name is unacceptable. The driver should be fixed at first to be unique.

cristicc:
> The `Analog` driver name is unacceptable. The driver should be fixed at first to be unique.

Indeed, that was way too generic. The sound card will be registered now as:
```
  card 0: rk3588es8316 [rk3588-es8316]
```
instead of
```
  card 0: Analog [Analog]
```
The related kernel patch is: https://lore.kernel.org/all/20230707162822.676024-1-cristian.ciocaltea@collabora.com/
The UCM files have been updated accordingly.

Thanks!

cristicc:
The kernel patch has been merged in [next-20230711](https://git.kernel.org/pub/scm/linux/kernel/git/next/linux-next.git/commit/?id=7efcaf997ae624caeccd046b6266fc2b7c0b91dc).

amazingfate:
@cristicc Hi, I'm running kernel 6.10-rc3 and there is no audio output from headphone on rock5b. Here is the output of alsa-info: http://alsa-project.org/db/?f=c31168b38ae850297ed748529bb92d6320b128b0.
How can I check what is wrong?

cristicc:
@amazingfate I've just tested with `6.10.0-rc3-next-20240613` and didn't notice any audio issue.

I'd suggest you to ensure the correct UCM files have been deployed - I don't have an Ubuntu based setup to verify, but you should have the following file installed by the `alsa-ucm-conf` package:

```
/usr/share/alsa/ucm2/conf.d/rk3588-es8316/rk3588-es8316.conf
```

To make sure this is not a higher level setup issue, stop any running pipewire/wireplumber services and run the following commands to test audio using the ALSA utilities:

```sh
$ alsaucm -c rk3588-es8316 reset reload set _verb HiFi list _devices set _enadev Headphones set _enadev Headset
  0: Headphones
    Headphones
  1: Headset
    Headset Microphone on IN2
    
$ speaker-test -D hw:rk3588es8316,0 -F S16_LE -c 2 -t wav
speaker-test 1.2.8

Playback device is hw:rk3588es8316,0
Stream parameters are 48000Hz, S16_LE, 2 channels
WAV file(s)
Rate set to 48000Hz (requested 48000Hz)
Buffer size range from 32 to 131072
Period size range from 16 to 65536
Using max buffer size 131072
Periods = 4
was set period_size = 32768
was set buffer_size = 131072
 0 - Front Left
 1 - Front Right
Time per period = 0.691699
 0 - Front Left
 1 - Front Right
```

amazingfate:
Hi, I'm running kernel 6.10-rc4 on ubuntu 24.04. After disabling pipewire/wireplumber/pulseaudio I can get exactly the same output from the commands above:
```
jfliu@rock-5b:~$ systemctl --user stop pulseaudio pulseaudio.socket 
jfliu@rock-5b:~$ alsaucm -c rk3588-es8316 reset reload set _verb HiFi list _devices set _enadev Headphones set _enadev Headset
  0: Headphones
    Headphones
  1: Headset
    Headset Microphone on IN2
jfliu@rock-5b:~$ speaker-test -D hw:rk3588es8316,0 -F S16_LE -c 2 -t wav

speaker-test 1.2.9

Playback device is hw:rk3588es8316,0
Stream parameters are 48000Hz, S16_LE, 2 channels
WAV file(s)
Rate set to 48000Hz (requested 48000Hz)
Buffer size range from 32 to 131072
Period size range from 16 to 65536
Using max buffer size 131072
Periods = 4
was set period_size = 32768
was set buffer_size = 131072
 0 - Front Left
 1 - Front Right
Time per period = 0.701255
 0 - Front Left
 1 - Front Right
Time per period = 2.731747
 0 - Front Left
 1 - Front Right
Time per period = 3.410957
 0 - Front Left
 1 - Front Right
Time per period = 2.732679
 0 - Front Left
 1 - Front Right
Time per period = 3.410644
 0 - Front Left
 1 - Front Right
```
But there is no sound heard when running speaker-test. When I run 6.1 vendor kernel from rockchip's SDK I can hear sound from headphone. While vendor kernel is using a non-mainline driver `rockchip_multicodecs.c`.

cristicc:
@amazingfate One possible explanation would be that `BootSequence` section did not execute because the sound card state has been persisted before updating the UCM configuration.

Check if there's `/var/lib/alsa/asound.state` available on your system. If yes, then most likely clearing the state would fix the issue:

```sh
$ systemctl stop alsa-state alsa-restore
$ rm /var/lib/alsa/asound.state
$ reboot
```

amazingfate:
@cristicc After removing `/var/lib/alsa/asound.state` and reboot, I can hear white noise from lef ear, but still no sound from speaker-test.

cristicc:
Please make sure the alsa services are stopped, otherwise the sound card state will be dumped prior rebooting and you end up with the same configuration reloaded on next boot.

It would also help if you could attach the file so I could take a quick look at what might be actually wrong in there.

cristicc:
For some reason I could identify a couple of unexpected values in your asound.state:
* `ADC Capture Volume` set to 175 instead of 192
* `'ADC PGA Gain Volume` set to 0 instead of 7

Are you sure the above mentioned services were **not** running before removing the file and rebooting?

Nevertheless, I forced an update of my card state with yours and I could still properly run the speaker test (did several tests, after rebooting the board, checked again the state against the original, etc), so I'm running out of ideas.. :-(

perexg:
Just another hint, the BootSequence can be executed directly using this command (assuming card 0):

```
alsaucm -c hw:0 set _boot ""
```

Just do it one time, the ALSA state service should store updated values on reboot.

cristicc:
@perexg Oh, thanks, that's a great tip!

@amazingfate So, instead of stopping services and removing the state file, the alternative would be:
```
$ alsaucm -c rk3588-es8316 set _boot ""
$ reboot
```

I tested and confirm it works as expected. However, it seems you are dealing with a different issue. 
Did you use the same HW setup when you checked with downstream kernel? Could you try with a different headset?

cristicc:
@amazingfate I've just noticed you mentioned stopping pulseaudio, instead of pipewire.

Since you are running Ubuntu 24.04, isn't pipewire the default sound server? You should see the following processes:

```sh
$ ps -e | grep wire
   1122 ?        00:04:00 pipewire
   1123 ?        00:00:24 wireplumber
   1124 ?        00:05:34 pipewire-pulse
```

amazingfate:
>Since you are running Ubuntu 24.04, isn't pipewire the default sound server? 

Yes it has pipewire running by default. Since you told me to stop pipewire, I just mask all pipewire user services. Then I started pulseaudio, but speaker-test would complain device or resource busy, so I also stoppped and masked pulseaudio service.

Now after running `alsaucm -c rk3588-es8316 set _boot ""` and reboot, speaker-test would also output white noise only from left earphone.
asoud.state file is in attachment.

cristicc:
> speaker-test would complain device or resource busy

That's precisely the reason any other application using the sound card should be stopped.

> speaker-test would also output white noise only from left earphone

Your configuration seems fine now. If that's not a HW fault, I don't really get how it works with downstream kernel.

I asked some questions above, but you didn't answer:
* Did you use the very same HW setup (board, headset/headphones) when you checked with downstream kernel?
* Could you try with a different headset?

amazingfate:
>Did you use the very same HW setup (board, headset/headphones) when you checked with downstream kernel?

Yes, I tested with the same board and two headphones. I'm toggling kernel with apt command. After toggling to downstream kernel and start pulseaudio service, I can hear sound from music played by mpv.

>Could you try with a different headset?

I have tried two different headphones, which are both headphone/mic 2 in 1 for cellphones, same white noise from left ear when running speaker-test

cristicc:
Ok, that's really odd.

The noise is only present when using speaker-test, or also with mpv & pulseaudio/pipewire?
Does speaker-test work as expected with downstream kernel?

If yes, could you please generate the alsa-info reports after running speaker-test with both upstream & downstream kernels? I'd like to check if there are any relevant differences between them.

amazingfate:
Here is the alsa-info of downstream kernel: http://alsa-project.org/db/?f=c1f8843e9dfc3b180db82f8f49966502f39b26a0.
The alsa device is rockchipes8316 instead of rk3588es8316. I can run speaker-test but I can't check the headphone sound now because this info is generated from ssh session.

Here is the alsa-info of mainline kernel: http://alsa-project.org/db/?f=bc7e8d08cf667344cda5758e543d32619170188a

cristicc:
Couldn't spot any obvious issue while comparing the reports, hence I would give your kernel a try on my device and see if I could reproduce the behavior.

Is there a package with the kernel binaries available for download? Or the whole rootfs which contains it?

amazingfate:
> Couldn't spot any obvious issue while comparing the reports, hence I would give your kernel a try on my device and see if I could reproduce the behavior.
> 
> Is there a package with the kernel binaries available for download? Or the whole rootfs which contains it?

There are armbian daily build images for rock5b: https://github.com/armbian/os/releases
You can try this: https://github.com/armbian/os/releases/download/24.8.0-trunk.224/Armbian_24.8.0-trunk.224_Rock-5b_noble_edge_6.10.0-rc5_gnome-oibaf_desktop.img.xz

And there are also kernel debs from armbian's beta repo:
https://k-space.ee.armbian.com/beta/pool/main/l/linux-dtb-edge-rockchip-rk3588/
https://k-space.ee.armbian.com/beta/pool/main/l/linux-6.10.0-rc4/

cristicc:
I've just tested the armbian kernel using my existing rootfs and confirm the audio still works fine.
```
$ uname -a
Linux debian-rockchip-rock5b-rk3588 6.10.0-rc4-edge-rockchip-rk3588 #1 SMP PREEMPT Sun Jun 16 20:40:16 UTC 2024 aarch64 GNU/Linux
```

I guess the last remaining thing would be to verify the entire armbian stack within the build image above.

cristicc:
I went through a clean SD card installation and managed to successfully test audio after the very first boot:

```sh
$ grep _PRETTY_NAME /etc/os-release
ARMBIAN_PRETTY_NAME="Armbian 24.8.0-trunk.224 noble"

$ uname -a
Linux rock-5b 6.10.0-rc5-edge-rockchip-rk3588 #1 SMP PREEMPT Sun Jun 23 21:08:54 UTC 2024 aarch64 aarch64 aarch64 GNU/Linux

$ alsaucm -c rk3588-es8316 set _boot ""
ALSA lib parser.c:2783:(load_toplevel_config) Unable to find the top-level configuration file '/usr/share/alsa/ucm2/ucm.conf'.
ALSA lib main.c:1554:(snd_use_case_mgr_open) error: failed to import rk3588-es8316 use case configuration -2
alsaucm: error failed to open sound card rk3588-es8316: No such file or directory

$ apt install alsa-ucm-conf
$ alsaucm -c rk3588-es8316 set _boot ""
$ speaker-test -D hw:rk3588es8316,0 -F S16_LE -c 2 -t wav
```

I've no explanation why it doesn't work on your side, other than having a potential HW fault. 
Obviously, the big question remains: why is that not exhibited when testing with downstream kernel?

amazingfate:
I'm testing with rock5b v1.3. I have sent this board to radxa to repair its broken headphone jack long time ago. But there are many other boards like rock5a, rock5c, and armsom sige7.
I remember that rock5c also doesn't have headphone sound. I will check these board later.

amazingfate:
@cristicc I tested with another rock5b and that has sound from speaker-test. So there should be hw falt with my v1.3 board. I'm sorry for your time wasting on my fault.

cristicc:
@amazingfate no problem, glad you figured this out! But I still don't understand how does it work with downstream kernel?!

Issue #326, opened 2023-06-21
Thinkpad Z16 Gen2 Mic mute LED not switching on when press F4


pma1:
The mic and sound output works fine.
alsamixer can switch on/off the F4/Mic-mute LED automatically.
But when press Fn+F4, the mic-mute led is not working.

/sys/class/sound/ctl-led/mic/card1# cat list
129

pma1:
numid=129,iface=MIXER,name='Capture Switch'

pma1:
alsa-info: http://alsa-project.org/db/?f=09c838933f2c3f4652090cdd6f3db7ec91aaf946

pseudocc:
also affects Thinkpad Z13 Gen2, alsa-info: https://alsa-project.org/db/?f=39cf33ce78a8a553ef0218e753b64a6a4c648fe8

perexg:
For Thinkpads, the ACPI driver should send KEY_MICMUTE (input subsystem) - see https://github.com/torvalds/linux/blob/master/drivers/platform/x86/thinkpad_acpi.c . This should be implemented for other platforms, too. The muting is implemented in gnome (window manager) / sound server.

In other words, this is not related to the sound subsytem at all.

pma1:
Mic mute LED on ThinkPad also rely on snd_ctl_led:
sound/pci/hda/thinkpad_helper.c

ucm2 uses "Mic ACP LED Capture Switch" on Thinkpad AMD platform.
There are similar issue #100 and #136
https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1849#note_1210933

pma1:
This is new AMD platform Pink Sardine, it enables new driver of DMIC sound/soc/amd/ps/ps-mach.c.

Without ucm2 conf:

1, Fallback profile is getting loaded and capture endpoint name is not listed as "Digital Microphone".
2, Mic Mute Led button is not reflected with Mute status.

It is just like the enablement of yellow carp and renoir

Sorry for the misunderstanding.

pma1:
The Fn+F4 generate ACPI event:
button/f20 F20 00000080 00000000 K

It's mapped in udev hwdb.d/60-keyboard.hwdb as KEY_MICMUTE.
The "Mic ACP LED Capture Switch" is not in /sys/class/sound/ctl-led/mic/card1/list
OSD will be shown in when switch Fn+F4.

I didn't see anything wrong in driver.

perexg:
[...]

PR #327, opened 2023-06-21
acp63: add initial support for AMD Pink Sardine - ACP63

pma1:
Link: #326


pseudocc:
verified that this works well on Thinkpad Z16/Z13 Gen2.

PR #328, opened 2023-06-22
Intel/sof-essx8336: Fix HiFi.conf

xry111:
1. <del>Enable Left/Right Headphone Mixer Left/Right DAC Switch in BootSequence.  Despite they seem enabled in sof-essx8336.conf, if I don't add this I'll need to enable them with alsamixer.  I guess they are killed by disdevall.</del>
2. Switch Headphone on/off in Headphones enable/disable sequence.  This should be obvious.
3. "Headset Switch" is not recognized and triggers errors running "alsaucm -c hw:0 set _verb Hifi".  Change it to "Headset Mic Switch". And in the disable sequence we should turn it off, not on.

### DIFF
--- a/ucm2/Intel/sof-essx8336/HiFi.conf
+++ b/ucm2/Intel/sof-essx8336/HiFi.conf
@@ -82,6 +82,7 @@ SectionDevice."Speaker" {
 	DisableSequence [
 		cset "name='Speaker Switch' off"
 	]
+
 	Value {
 		PlaybackPriority 100
 		PlaybackPCM "hw:${CardId}"
@@ -97,6 +98,14 @@ SectionDevice."Headphones" {
 
 	ConflictingDevice [
 		"Speaker"
+	]
+
+	EnableSequence [
+		cset "name='Headphone Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='Headphone Switch' off"
 	]
 
 	Value {
@@ -123,12 +132,12 @@ SectionDevice."Headset" {
 	}
 
 	EnableSequence [
-		cset "name='Headset Switch' on"
+		cset "name='Headset Mic Switch' on"
 		cset "name='Digital Mic Mux' 'dmic disable'"
 	]
 
 	DisableSequence [
-		cset "name='Headset Switch' on"
+		cset "name='Headset Mic Switch' off"
 	]
 
 	Value {


perexg:
The `disdevall` calls only disable sequences for defined devices. So it looks like that another control brings them off. Could you try to manually turn off the controls one by one and look, if `Left Headphone Mixer Left DAC Switch` stays on?

```
cset "name='Internal Mic Switch' off"
cset "name='Speaker Switch' off"
cset "name='Headset Switch' on"   <-- seems like  a bug in the original configuration
```

xry111:
> The `disdevall` calls only disable sequences for defined devices. So it looks like that another control brings them off. Could you try to manually turn off the controls one by one and look, if `Left Headphone Mixer Left DAC Switch` stays on?
> 
> ```
> cset "name='Internal Mic Switch' off"
> cset "name='Speaker Switch' off"
> cset "name='Headset Switch' on"   <-- seems like  a bug in the original configuration
> ```

It seems `disdevall` is not guilt.  Even if I comment out `disdevall ""` they are still disabled by default.

perexg:
> Hmm, it looks like the BootSequence is only executed by `alsactl init`. Is it expected that everyone should run `alsactl init`?

It's correct. The controls should be saved to the `/var/lib/alsa/asound.state` and restored from this file after first initialization from the `BootSequence`. Your system should have one of `alsa-state` or `alsa-restore` systemd services active. This mechanism allows "custom" initialization made by users.

xry111:
> > Hmm, it looks like the BootSequence is only executed by `alsactl init`. Is it expected that everyone should run `alsactl init`?
> 
> It's correct. The controls should be saved to the `/var/lib/alsa/asound.state` and restored from this file after first initialization from the `BootSequence`. Your system should have one of `alsa-state` or `alsa-restore` systemd services active. This mechanism allows "custom" initialization made by users.

On my system the `asound.state` is saved and loaded by udev, so things are fine after running `alsactl init`.  I've dropped these two lines from the PR.

perexg:
So things are working without any extra command requirement with this PR update?

xry111:
> So things are working without any extra command requirement with this PR update?

For a "normal" distro I guess so.  But for a "minimal" distro `alsactl` can be installed after installing the distro, in this case we need to manually run `alsactl init` or `alsactl restore`, or rebooting the system so `systemd` will run `alsactl restore`.

PR #335, opened 2023-07-18
X13s volume fixes

Srinivas-Kandagatla:
This patchset fixes some of the Volume issues that are seen on X13s, most of the existing ucm was controlling Digital Gains instead of Analog Gains. With this patchset the volume on both speakers are much better and not saturated.


ucm2/codecs/qcom-lpass/wsa-macro/SpeakerEnableSeq.conf
```diff
@@ -5,6 +5,6 @@ EnableSequence [
 	cset "name='WSA_RX1 INP0' RX1"
 	cset "name='WSA_COMP1 Switch' 1"
 	cset "name='WSA_COMP2 Switch' 1"
-	cset "name='WSA_RX0 Digital Volume' 68"
-	cset "name='WSA_RX1 Digital Volume' 68"
+	cset "name='WSA_RX0 Digital Volume' 84"
```
perexg, line 8: Won't be better to remove this settings from the EnableSequence and keep the initailization only in the BootSequence ?

Srinivas-Kandagatla: Makes sense, will update the PR with suggested changes.

Srinivas-Kandagatla: Sorry for the delay, branch is now pushed with suggested change.

jhovold:
@Srinivas-Kandagatla , your updated pull-request is not functionally equivalent to your first ones and causes the speaker volume to again be set too low.

I tried adding back the volume csets to ucm2/codecs/qcom-lpass/wsa-macro/SpeakerEnableSeq.conf and that fixed the issue.

Please investigate and respin ASAP so we can get this fixed as audio is currently broken on the X13s without these changes merged.

Srinivas-Kandagatla:
@jhovold thanks for testing and reporting this, I think I found one issue here. We seems to be setting the same mixer WSA_RX0/WSA_RX1 Digital Volume in SpeakerDisableSeq.conf as well. So we will end up with low volume because of this. I have fixed this now, Will push to this PR with those changes.

ucm2/codecs/wsa883x/init.conf
```diff
@@ -0,0 +1,18 @@
+# WSA883x specific volume control settings
+
+BootSequence [
+	cset "name='WSA_RX0 Digital Volume' 84"
+	cset "name='WSA_RX1 Digital Volume' 84"
+]
+
```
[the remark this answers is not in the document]
Srinivas-Kandagatla: No, it should be removed.

jhovold:
The updated fixes seem to work as intended.

Note that with the addition of rx-macro/init.conf you now have a redundant initialisation of "RX_RX0 Digital Volume" in wcd938x/init.conf instead. Not sure if you want to respin this PR again or if you can just drop that as a follow-up clean up.

jhovold:
Still works after dropping the redundant initialisation so these should be good to go:

Reviewed-by: Johan Hovold <johan+linaro@kernel.org>
Tested-by: Johan Hovold <johan+linaro@kernel.org>

Note that users will need to clear their asound.state after updating the ucm configuration, and that after doing so headphone output is broken until a second reboot. So there still appears to be some bugs in these configuration files, but as I assume this issue was not introduced by this series that should be fixed separately.

@perexg, is it possible to get this merged and included in the next release?

PR #347, opened 2023-09-05
ucm2: Sync Librem 5 config with Purism's downstream one

Newbytee:
The config currently in this repo has several issues. The Purism
downstream one resolves these.

Taken from https://source.puri.sm/Librem5/librem5-base, commit f5b51beb144f76ef3bc483b74e19867bd6364d32

### DIFF
--- a/ucm2/NXP/iMX8/Librem_5/HiFi.conf
+++ b/ucm2/NXP/iMX8/Librem_5/HiFi.conf
@@ -1,12 +1,38 @@
 SectionVerb {
 	EnableSequence [
 		disdevall ""
+		cset "name='Speaker Mixer Switch' off,off"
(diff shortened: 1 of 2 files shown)


perexg:
From the first look this is very unclean update.

Some modifications are straight, but the big ones seems a bit step back.

Some comments:

   Header to Header1 is not finished (not all ConflictingDevices blocks were modified)
   The `disdevall ""` in EnableSequence will call DisableSequences for all devices, so it's not required to duplicate the work.
   Empty DisableSequence block in Verb.

agx:
@Newbytee  Purism's downstream config still relies on unmerged kernel patches for the jack mic detection so it's not upstreamable as is. 

/cc @dos1 as in my recollection he wanted to upstream the kernel bits before updating UCM.

PR #358, opened 2023-10-11
Add UCM2 configuration for Behringer UMC404HD

Janhouse:
The only difference between UMC204HD and UMC404HD are the 2 additional inputs and a different USB device ID. Tested with my 404HD, works great.

### DIFF
--- a/ucm2/USB-Audio/Behringer/UMC404HD-HiFi.conf
+++ b/ucm2/USB-Audio/Behringer/UMC404HD-HiFi.conf
@@ -1,0 +1,200 @@
+Include.pcm_split.File "/common/pcm/split.conf"
+
+Macro [
+	{
+		SplitPCM {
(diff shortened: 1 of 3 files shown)


Janhouse:
@perexg does this look OK?

tam1m:
I was about send the same PR when I found yours. This works perfectly fine. The one thing I noticed in your code is capture priorities for MIC 1 to 4. MIC1 CapturePriority set to 200 and all the others at 100. Guess, this was just oversight from copying the UMC204HD config?

Janhouse:
@tam1m thanks for the comment. I am actually not sure what the priority should be. I presume higher number means higher priority. Can't the rest of the inputs stay at 100? :slightly_smiling_face:

perexg:
Thanks. The priorities should be different, but they don't play any role for devices without jack detection mechanism. I've changed them anyway.

PR #362, opened 2023-10-23
Add UCM2 configuration for Topping DX3 Pro+

ZVNexus:
Fixes it showing up as "digital" and "analog" output in Settings, should just be "Headphones".


perexg:
Note that I moved this config to `USB-Audio/Common/HeadphonesOnly.conf` in  93d015d12c68a0516d181a5c2682037f1ec256a7. It would be nice, if you can re-test with the latest UCM tree.

ZVNexus:
@perexg Retested after changes to make it generic, didn't take effect but that was because of the ProfileName. Pushed a new PR to fix it, and after that it shows up as expected.

image

Issue #351, opened 2023-09-12
AIF2 ADC Stereo Capture Route is uninitialised in alsa-ucm-conf


broukema:
DESCRIPTION:
`AIF2 ADC Stereo Capture Route` is not initialised in
`alsa-ucm-conf`. This implies that if its setting is significant, any
accidental change by the user (using any of the various `alsa` tools)
to a wrong setting will not be corrected by any of the three
`PinePhone/*.conf` files, leading, in particular, to audio errors during a phone
call. This was discovered by testing variations on the Mobian version
of the PinePhone alsa-ucm-conf files.

CODING ARGUMENT:
All parameters that can significantly affect the audio system should be
either initialised or overridden in configuration files that the user
expects to necessarily work, such as those for phone calls.
In `alsa-ucm-conf` commit b68aa52 (one beyond v1.2.10):
````
$ grep -ri "AIF2 ADC Stereo Capture" .
$
````
shows that `AIF2 ADC Stereo Capture Route` is not set for any device at all
(PinePhone or other).

HARDWARE ARGUMENT:
For the PinePhone (PP), Megi recommends in https://xnux.eu/devices/feature/audio-pp.html
that `AIF2 ADC Stereo Capture Route` is set to `Mix Mono`.

TESTING ARGUMENT:
Experimentation with different stored alsa settings (from
````
sudo alsactl store -f - > YYYYMMDD_example_alsa_settings
````
) showed that settings files whose only substantial difference was in the setting
of `AIF2 ADC Stereo Capture Route` had the effect that audio from the PP was heard on the
remote phone when the value was `Mix Mono` and was not heard with the value `Stereo`.
Testing was done with the equivalent of
https://codeberg.org/boud/pinephone_hacks/src/commit/7c7b6660a4446a24747f2c2edbe1d57d863ffb9f/audio_hacks/sort_of_restart_audio
with the settings file
https://codeberg.org/boud/pinephone_hacks/src/commit/7c7b6660a4446a24747f2c2edbe1d57d863ffb9f/audio_hacks/20230806_example_alsa_settings#L271
that has the `Mix Mono` value selected.

RECOMMENDATIONS:
* `AIF2 ADC Stereo Capture Route` should be initialised or set to *something* in one of the
.conf files (in the PP case, most likely in `VoiceCall.conf`), at least for the PinePhone,
to prevent users who experiment and have 
difficulty remembering which of the 50 PinePhone alsa parameters should be set in which way from
[...]

PR #365, opened 2023-11-01
Initialise AIF2 ADC Stereo Capture Route

broukema:
As pointed out in issue #351 [1], the parameter 'AIF2 ADC Stereo Capture Route' was previously uninitialised. I'm not very convinced that this particular value (Mix Mono) is the best combination with the other PinePhone settings, but better that there is at least a value rather than no value, so that people can provide bug reports based on tests that are closer to being reproducible. As described in [1], there is circumstantial evidence favouring this value.

This commit does not set a value for HiFi.conf - presumably the value is irrelevant for HiFi, which does not appear to be as buggy as VoiceCall.

[1] https://github.com/alsa-project/alsa-ucm-conf/issues/351

### DIFF
--- a/ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
+++ b/ucm2/Allwinner/A64/PinePhone/VoiceCall.conf
@@ -2,6 +2,10 @@ SectionVerb {
 	EnableSequence [
 		cset "name='AIF2 Digital DAC Playback Switch' on"
 		cset "name='AIF2 ADC Mixer ADC Capture Switch' on"
+		#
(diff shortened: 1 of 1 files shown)


broukema:
I now have a working asound.state snapshot file together with a small modification of the three *.conf files here upstream that together gave valid audio calls several times in a row (with very little remote-to-remote echo). There are many more missing parameters, but I checked that `AIF2 ADC Stereo Capture Route` set to `Mix Mono` works correctly for me. It might be worth waiting to see if I can propose a much more complete merge request, but this one on its own appears correct to me. So I'll switch it back to 'ready for review'.

PR #374, opened 2024-01-07
Also search for ucm2/conf.d/[${CardDriver}|${KernelDriver}]/${CardName}.conf

Daniel-Abrecht:
The CardLongName may be derived from various DMI properties of the system, that have little to do with the card itself. Sometimes, that may be a good thing, but sometimes, it may contain the completely generic "Unknown-UnknownProduct-", see https://gitlab.com/postmarketOS/pmaports/-/issues/1498#note_1716959721 for example. At least for devices using device tree and simple-card, the name of the card (CardName), which then comes from the device tree, which can be provided by the kernel too, may be a lot more relevant, and does actually identify the device.


perexg:
Why the simple-card driver does not set the CardLongName (GUI string) from DT, too? The DMI fallback for card's long name is there just for ASoC laziness.

Daniel-Abrecht:
I don't think the simple-card driver can set the CardLongName from the DT right now. It would have to set `long_name` in `struct snd_soc_card`, but I couldn't find any reference to `long_name` in `sound/soc/generic/simple-card.c`. I think that would have to be added to the `simple_parse_of` function in that file, it seams `name` is set there.

Shall I send a patch for that?

perexg:
If you look to conf.d/simple-card there is `rockchip,es8316-codec.conf` link. No idea, how this does work. It seems like that the long name is used for this match or things are really wrong. The simple-card driver should allow to modify the long name, too.

Another option is to create `conf.d/simple-card/simple-card.conf` and use UCM configuration for the more precise matching (`${CardName}` substitution and condition). Look to `conf.d/USB-Audio/USB-Audio.conf` for an example or `conf.d/acp5x/Valve-Jupiter-1.conf` (DMI matching) etc.

Daniel-Abrecht:
I think `rockchip,es8316-codec.conf` is used by the pinebook-pro. In it's dtb, on line 133 `https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/arch/arm64/boot/dts/rockchip/rk3399-pinebook-pro.dts#n133` it says `simple-audio-card,name = "rockchip,es8316-codec";`. That only sets the name, not the long name, though.

So, the circumstances in which this works are probably going to be about the same as for the pinephone-pro. For distros where the kernel is loaded directly by uboot without EFI, there will be no DMI, and the `long_name` is going to fall back to `name`, here: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/sound/soc/soc-core.c?id=052d534373b7ed33712a63d5e17b2b6cdbce84fd#n2268 (line 2268).

It seams there is also a blacklist of invalid DMI names in the kernel, I'll add `Unknown` and `Unknown Product` to that list.

I will also look into doing the matching in `conf.d/simple-card/simple-card.conf`, and will update this pull request later once I've figured that out.

PR #378, opened 2024-01-11
USB-Audio: add MOTU M6 config

davide125:
This card shares the same ID as the MOTU M2 and M4, so we use the long card name to identify it.


davide125:
`alsa-info.sh` output is here: http://alsa-project.org/db/?f=4a7060bbf2e0ba651a82653f583a376a99252fef (this includes a bunch of other cards, only the M6 stuff is relevant for this PR).

PR #402, opened 2024-03-11
sof-soundwire: rt1308: Fix single amp configuration

ujfalusi:
If the device have single rt1308 amp then the codec's Channel Select should be set to "LR" and not to "LL". "LL" is only valid if we have 2 amps, in that case one is the Left and the other is the Right channel.

This fixes for example Dell Latitude 9520 which only played on the Left speaker, Right is silent.

### DIFF
--- a/ucm2/sof-soundwire/rt1308.conf
+++ b/ucm2/sof-soundwire/rt1308.conf
@@ -28,8 +28,9 @@ DefineMacro.rt1308spk.If.0 {
 SectionDevice."Speaker" {
 	Comment	"Speaker"
 
-	Macro.num1.rt1308spk { ForAmps "[12]" Amp 1 Sel LL }
-	Macro.num2.rt1308spk { ForAmps "2" Amp 2 Sel RR }
+	Macro.num1.rt1308spk { ForAmps "1" Amp 1 Sel LR }
+	Macro.num2.rt1308spk { ForAmps "2" Amp 1 Sel LL }
+	Macro.num3.rt1308spk { ForAmps "2" Amp 2 Sel RR }
 
 	EnableSequence [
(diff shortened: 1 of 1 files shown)


ujfalusi:
> Shorter definition:
> 
> ```
> Macro.num1.rt1308spk { ForAmps "1" Amp 1 Sel LR }
> Macro.num2.rt1308spk { ForAmps "2" Amp 1 Sel LL }
> Macro.num3.rt1308spk { ForAmps "2" Amp 2 Sel RR }
> ```
> 
> Could you test, please ?

Does not work.
after boot to see if UCM does something:
```
amixer -c0 sset 'rt1308-1 RX Channel Select' 'RR'
```
The log in (PW starts)
```
# amixer -c0 | grep -A3 rt1308
Simple mixer control 'rt1308-1 DAC L',0
  Capabilities: pswitch pswitch-joined
  Playback channels: Mono
  Mono: Playback [off]
Simple mixer control 'rt1308-1 DAC R',0
  Capabilities: pswitch pswitch-joined
  Playback channels: Mono
  Mono: Playback [off]
Simple mixer control 'rt1308-1 RX Channel Select',0
  Capabilities: enum
  Items: 'LR' 'LL' 'RL' 'RR'
  Item0: 'RR'
```
With the PR it ends up at "LR" and the DACs are unmuted, but oddly it looks like during PW startup the Channel Select takes "LL" for a moment before settling on the correct "LR".

ujfalusi:
@perexg, I stand corrected, your suggestion works like this PR, it ends up in the correct configuration.
I will update the PR, thanks for the suggestion (first ucm2 change ever on my side).

ujfalusi:
Changes since v1:
- Use the shorter and cleaner version suggested by @perexg

PR #409, opened 2024-04-15
ucm2: soundwire: add rt722 SDCA device

jack-cy-yu:
Add support of ucm to rt722 SDCA.

### DIFF
--- a/ucm2/sof-soundwire/HiFi.conf
+++ b/ucm2/sof-soundwire/HiFi.conf
@@ -8,6 +8,14 @@ SectionVerb {
 	Value.TQ "HiFi"
 }
 
+If.multicodec {
+	Condition {
+		Type String
+		Empty "${var:MultiCodec1}"
+	}
+	False.Include.multicodec.File "/sof-soundwire/${var:MultiCodec1}.conf"
+}
+
 If.spkdev {
 	Condition {
 		Type String
--- a/ucm2/sof-soundwire/rt722.conf
+++ b/ucm2/sof-soundwire/rt722.conf
@@ -1,0 +1,97 @@
+# Use case Configuration for sof-soundwire card
+
+If.codecmic {
+	Condition {
+		Type RegexMatch
+		Regex "(rt722(-sdca)?)"
+		String "${var:MultiMicShadow}"
+	}
+	True {
+		SectionDevice."Mic" {
+			Comment	"SoundWire microphones"
+
+			EnableSequence [
+				cset "name='rt722 FU1E Capture Switch' 1"
+			]
+
+			DisableSequence [
+				cset "name='rt722 FU1E Capture Switch' 0"
+			]
+
+			Value {
+				CapturePriority 100
+				CapturePCM "hw:${CardId},4"
+				CaptureSwitch "rt722 FU1E Capture Switch"
+				CaptureVolume "rt722 FU1E Capture Volume"
+				CaptureMixerElem "rt722 FU1E"
+			}
+		}
+	}
+}
+
+If.codecspk {
+	Condition {
+		Type RegexMatch
+		Regex "(rt722(-sdca)?)"
+		String "${var:MultiSpeakerShadow}"
+	}
+	True {
+		SectionDevice."Speaker" {
+			Comment "Speaker"
+
+			EnableSequence [
+				cset "name='Speaker Switch' on"
+			]
+
+			DisableSequence [
+				cset "name='Speaker Switch' off"
+			]
+
+			Value {
+				PlaybackPriority 100
+				PlaybackPCM "hw:${CardId},2"
+				PlaybackMixerElem "rt722 FU06"
+				PlaybackVolume "rt722 FU06 Playback Volume"
+			}
+		}
+	}
+}
+
+SectionDevice."Headphones" {
+	Comment	"Headphones"
+
+	EnableSequence [
+		cset "name='Headphone Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='Headphone Switch' off"
+	]
+
+	Value {
+	      PlaybackPriority 200
+	      PlaybackPCM "hw:${CardId}"
+	      JackControl "Headphone Jack"
+	}
+}
+
+SectionDevice."Headset" {
+	Comment "Headset Microphone"
+
+	EnableSequence [
+		cset "name='rt722 FU0F Capture Switch' 1"
+	]
+
+	DisableSequence [
+		cset "name='rt722 FU0F Capture Switch' 0"
+	]
+
+	Value {
+		CapturePriority 200
+		CapturePCM "hw:${CardId},1"
+		JackControl "Headset Mic Jack"
+		CaptureSwitch "rt722 FU0F Capture Switch"
+		CaptureVolume "rt722 FU0F Capture Volume"
+		CaptureMixerElem "rt722 FU0F"
+	}
+}
--- a/ucm2/sof-soundwire/sof-soundwire.conf
+++ b/ucm2/sof-soundwire/sof-soundwire.conf
@@ -49,6 +49,59 @@ DefineRegex {
 	}
 }
 
+DefineRegex.MultiCodec {
+	Regex "(rt722(-sdca)?)"
+	String "${var:SpeakerCodec1} ${var:HeadsetCodec1} ${var:MicCodec1}"
+}
+
+If.multi_init {
+	Condition {
+		Type String
+		Empty "${var:MultiCodec1} "
+	}
+	False {
+		Define {
+			MultiSpeakerShadow "${var:SpeakerCodec1}"
+			MultiMicShadow "${var:MicCodec1}"
+		}
+		Include.multi_init.File "/codecs/${var:MultiCodec1}/init.conf"
+	}
+}
+
+If.multi_speaker {
+	Condition {
+		Type RegexMatch
+		Regex "${var:SpeakerCodec1}"
+		String "${var:MultiCodec1}"
+	}
+	True {
+		Define.SpeakerCodec1 ""
+	}
+}
+
+If.multi_mic {
+	Condition {
+		Type RegexMatch
+		Regex "${var:MicCodec1}"
+		String "${var:MultiCodec1}"
+	}
+	True {
+		Define.MicCodec1 ""
+	}
+}
+
+If.multi_headset {
+	Condition {
+		Type RegexMatch
+		Regex "${var:HeadsetCodec1}"
+		String "${var:MultiCodec1}"
+	}
+	True {
+		Define.HeadsetCodec1 ""
+	}
+}
+
+
 If.hs_init {
 	Condition {
 		Type RegexMatch
@@ -146,3 +199,5 @@ If.HdmiIec61937 {
 	}
 	True.Macro.save_hdmi_cfg.HdmiPCMSave { Name "42-sof-hdmi" }
 }
+
+
--- a/ucm2/codecs/rt722/init.conf
+++ b/ucm2/codecs/rt722/init.conf
@@ -1,0 +1,11 @@
+# RT722-sdca specific volume control settings
+
+BootSequence [
+	cset "name='rt722 ADC 24 Mux' 'DMIC1'"
+	cset "name='rt722 ADC 25 Mux' 'DMIC2'"
+	cset "name='rt722 ADC 22 Mux' 'MIC2'"
+	cset "name='rt722 FU1E Capture Volume' 63"
+	cset "name='rt722 FU06 Playback Volume' 60"
+	cset "name='rt722 FU05 Playback Volume' 35"
+	cset "name='rt722 FU0F Capture Volume' 63"
+]


ucm2/sof-soundwire/rt722-0.conf
```diff
@@ -0,0 +1,18 @@
...
+
+        DisableSequence [
+                cset "name='Speaker Switch' off"
+        ]
+
+        Value {
+              PlaybackPriority 100
+              PlaybackPCM "hw:${CardId},2"
+        }
+}
```
plbossart, line 18: the file naming is confusing, why not use the function name, i.e.

rt722-amp.conf
rt722-jack.conf
rt722-mic.conf

ucm2/sof-soundwire/sof-soundwire.conf
```diff
@@ -47,7 +47,7 @@ DefineRegex {
 If.hs_init {
 	Condition {
 		Type RegexMatch
-		Regex "(cs42l43|rt5682|rt700|rt711|rt713(-sdca)?)"
+		Regex "(rt722|cs42l43|rt5682|rt700|rt711|rt713(-sdca)?)"
```
plbossart, line 50: keep all RT stuff in a logical order instead of rt722 first?

ucm2/sof-soundwire/rt722-0.conf
jack-cy-yu: Because the speaker structure defined in HiFi.conf is like below

If.spkdev {
	Condition {
		Type String
		Empty "${var:SpeakerCodec1}"
	}
	False.Include.spkdev.File "/sof-soundwire/${var:SpeakerCodec1}-${var:SpeakerAmps1}.conf"
}

micdev and hsdev also has its structure defined in HiFi.conf, so I follow these just like what rt712 did.

ucm2/sof-soundwire/sof-soundwire.conf
jack-cy-yu: OK, I'll modify it.

ucm2/sof-soundwire/rt722-0.conf
plbossart: rt722-0.conf means nothing to me...

ucm2/sof-soundwire/HiFi.conf
```diff
@@ -10,25 +10,31 @@ SectionVerb {
...
 
 If.hsdev {
 	Condition {
-		Type String
-		Empty "${var:HeadsetCodec1}"
+		Type RegexMatch
+		Regex "(rt722(-sdca)?)"
+		String "${var:HeadsetCodec1}"
 	}
+	True.Include.hsdev.File "/sof-soundwire/${var:HeadsetCodec1}-jack.conf"
```
plbossart, line 37: @perexg this needs your review.

RT722 is one of the first 'integrated' codecs with multiple "functions" in the same chip. Before we had one function per chip, but some for the same rt722 we have mic,jack,amp capabilities.

So we probably need a mechanism to identify such functions without changing the main filters.

ucm2/sof-soundwire/rt722-0.conf
plbossart: @bardliao @jack-cy-yu I wonder if the amixer 'components' string is correct now that I think of it, could it be that we report zero amps?

Can you share the result of amixer -Dhw:L info ?

ucm2/sof-soundwire/HiFi.conf
bardliao: I think we don't need to split rt722.conf into rt711-spk.conf, rt722-mic.conf, and rt722-sdca-jack.conf. We can put SectionDevice."Mic" and SectionDevice."Speaker" in rt722.conf and SectionDevice."Headphones" and SectionDevice."Headset" in rt722-sdca.conf.

bardliao: Or we could add suffix in machine driver, and use rt722-amp.conf, rt722-jack.conf, and rt722-mic.conf.

ucm2/sof-soundwire/rt722-0.conf
bardliao: @plbossart 
Card hw:0 'sofsoundwire'/'Intel Soundwire SOF'
  Mixer name	: 'Intel Meteor Lake HDMI'
  Components	: 'HDA:8086281d,80860101,00100000 cfg-spk:2 cfg-amp:0 spk:rt722 hs:rt722-sdca mic:rt722'
  Controls      : 21
  Simple ctrls  : 16

ucm2/sof-soundwire/rt722-spk.conf
```diff
@@ -0,0 +1,18 @@
...
+                cset "name='Speaker Switch' on"
+        ]
+
+        DisableSequence [
+                cset "name='Speaker Switch' off"
+        ]
+
+        Value {
+              PlaybackPriority 100
+              PlaybackPCM "hw:${CardId},2"
```
perexg, line 16: No hardware volume control?

ucm2/sof-soundwire/rt722-0.conf
perexg: The driver serves one chip or two chips (rt722 / rt722-sdca) ?

bardliao: > The driver serves one chip or two chips (rt722 / rt722-sdca) ?

One chip, but currently, machine driver will add "-sdca" suffix for headset.

plbossart: @bardliao this configuration makes no sense "cfg-spk:2 cfg-amp:0". We're missing something here, the amps cannot be zero. and I thought we stopped reporting cfg-spk?

perexg: I think that we should end with one `rt722.conf` file and use a different identifier in components (like `cfg-codec:rt722`) to not follow the scheme with separate chips. All other components without use should be removed.

plbossart: @perexg "cfg-codec:rt722" may only work if this is the only chip in the system. I vaguely remember the possibility of having an additional discrete amplifier in addition to RT722 for e.g. tweeters.

With SDCA the 'functions' may be part of a single chip or a set of discrete ones. There's no general rule on what the partition might be.

bardliao: > @bardliao this configuration makes no sense "cfg-spk:2 cfg-amp:0". We're missing something here, the amps cannot be zero. and I thought we stopped reporting cfg-spk?

The output is from @jack-cy-yu. He may not use the latest sof-dev kernel. The chg-amp:0 issue was fixed by https://github.com/thesofproject/linux/pull/4909

perexg: > @perexg "cfg-codec:rt722" may only work if this is the only chip in the system. I vaguely remember the possibility of having an additional discrete amplifier in addition to RT722 for e.g. tweeters.
> 
> With SDCA the 'functions' may be part of a single chip or a set of discrete ones. There's no general rule on what the partition might be.

We can extend rt722.conf (add include) or the sof-soundwire.conf file when we know this hardware. From my view, trying to separate functions for one driver seems a bit more difficult than maintain whole configuration in one file.

We can do something like (assuming that all other codec indentifications will be removed from components string, keeping only `codec:rt722` there:

```
diff --git a/ucm2/sof-soundwire/HiFi.conf b/ucm2/sof-soundwire/HiFi.conf
index ffea077..8777ee6 100644
--- a/ucm2/sof-soundwire/HiFi.conf
+++ b/ucm2/sof-soundwire/HiFi.conf
@@ -8,6 +8,14 @@ SectionVerb {
        Value.TQ "HiFi"
 }
 
+If.codec {
+       Condition {
+               Type String
+               Empty "${var:Codec1}"
+       }
+       False.Include.codec.File "/sof-soundwire/${var:Codec1}.conf"
+}
+
 If.spkdev {
        Condition {
                Type String
diff --git a/ucm2/sof-soundwire/sof-soundwire.conf b/ucm2/sof-soundwire/sof-soundwire.conf
index 35a262c..1ed888c 100644
--- a/ucm2/sof-soundwire/sof-soundwire.conf
+++ b/ucm2/sof-soundwire/sof-soundwire.conf
@@ -18,6 +18,10 @@ Define {
 }
 
 DefineRegex {
+       Codec {
+               Regex " codec:([a-z0-9]+(-sdca)?)"
+               String "${CardComponents}"
+       }
        SpeakerCodec {
                Regex " spk:([a-z0-9]+(-sdca)?)"
                String "${CardComponents}"
@@ -44,6 +48,15 @@ DefineRegex {
        }
 }
 
+If.codec_init {
+       Condition {
+               Type RegexMatch
+               Regex "(rt722)"
+               String "${var:Codec1}"
+       }
+       True.Include.codec_init.File "/codecs/${var:Codec1}/init.conf"
+}
+
 If.hs_init {
        Condition {
```

jack-cy-yu: I've followed the suggestion of @perexg  and added the patch like attachment, it works on my MTL platform, but I declare the variable "Codec1" from 

 DefineRegex {
+	Codec {
+		Regex " spk:([a-z0-9]+(-sdca)?)"
+		String "${CardComponents}"
+	}
 ...
 }

we need "Regex " **codec**:([a-z0-9]+(-sdca)?)" " to replace "Regex " **spk**:([a-z0-9]+(-sdca)?)"".  @plbossart @bardliao Could you help to add a "codec:rt722" in components string? 

[0001-ucm2-soundwire-add-rt722-SDCA-device.zip](https://github.com/alsa-project/alsa-ucm-conf/files/15091751/0001-ucm2-soundwire-add-rt722-SDCA-device.zip)

bardliao: > we need "Regex " **codec**:([a-z0-9]+(-sdca)?)" " to replace "Regex " **spk**:([a-z0-9]+(-sdca)?)"". @plbossart @bardliao Could you help to add a "codec:rt722" in components string?
> 

Yes, I will submit a machine driver change for it.

plbossart: I really didn't get how we will support the case with two chips for the main/companion amplifier topology. It's not an hypothetical design, it's what OEMs want to scale their platforms between mid/high tiers.

perexg: The driver should set a hint for UCM in the components string and UCM configuration should be extended for both cases. There are several possibilities to describe this situation, but I guess, we should do it when we have a working driver.

plbossart: I am fine with pushing the main/companion amplifier work to later, but the term 'codec' means absolutely nothing in the case of RT722. It's a multi-function device, so we don't know if we are referring to the jack, amp or mic function, or all of the above.

perexg: Judging from the rt722-sdca.c source file (datasheet is not probably publicly available), the comments in `rt722_sdca_controls` table are straight forward about the expected usage. I don't comment the ugly control names transferring any abstraction to upper layers. We lost this abstraction from 0day for ASoC drivers.

plbossart (approved):
LGTM, I think all the driver changes are already on their way to 6.10, so this matching UCM update would be neat for end-users.

perexg:
It looks screwed. What's the final components string?

plbossart:
@bardliao can you share what your latest component string would be?

bardliao:
@perexg @plbossart Components	: 'HDA:8086281d,80860101,00100000  cfg-amp:1 iec61937-pcm:7,6,5 hs:rt722 spk:rt722 mic:rt722'

perexg:
In this case, codecs/rt722/init.conf and rt722.conf are included/evaluated multiple times which is not ideal. Some ideas to optimize this:

```
DefineRegex.MultiCodec {
       Regex "(rt722(-sdca)?)"
       String "${var:SpeakerCodec1} ${var:HeadsetCodec1} ${var:MicCodec1}"
}

If.multi_init {
        Condition {
                Type String
                Empty "${var:MultiCodec1} "
        }
        False {
                 Define {
                        SpeakerCodecShadow "${var:SpeakerCodec1}"
                        HeadsetCodecShadow "${var:HeadsetCodec1}"
                        MicCodecShadow "${var:MicCodec1}"
                        SpeakerCodec1 ""   # FIXME: Add If - discard when SpeakerCodec1 == MultiCodec1
                        HeadsetCodec1 ""   # FIXME: dtto
                        MicCodec1 ""       # FIXME: dtto
                 }
                 Include.multi_init.File "/codecs/${var:MultiCodec1}/init.conf"
        }
}
```


... rt722.conf should use `...Shadow` variables for the decision. Also HiFi conf should be updated to include `MultiCodec1` file.

ucm2/sof-soundwire/sof-soundwire.conf
```diff
@@ -146,3 +152,49 @@ If.HdmiIec61937 {
...
+	Condition {
+		Type String
+		Empty "${var:MultiCodec1} "
+	}
+	False {
+		Include.multi_init.File "/codecs/${var:MultiCodec1}/init.conf"
+	}
+}
+
+If.multi_speaker {
```
perexg, line 166: Those sections must be moved before ????Codec1 variables use to avoid multiple init.conf inclusions. So before "If.hs_init", "If.mic_init".

ucm2/sof-soundwire/rt722.conf
```diff
@@ -0,0 +1,97 @@
+# Use case Configuration for sof-soundwire card
+
+If.codecmic {
+	Condition {
+		Type RegexMatch
+		Regex "(rt722(-sdca)?)"
+		String "${var:MicCodec1}"
```
perexg, line 7: The `MicCodec1` variable may be empty when `MultiCodec1` is set. You should define/use another variable to make this decision.

perexg (changes requested):

PR #410, opened 2024-04-16
ucm2: HDA: Use Master volume control for dual speakers

tuxedoxt:
This PR extends the existing 'activate Bass Speaker' case with additional configuration

- Use Master mixer for volume control (instead of only using Speaker)
- Configure Speaker Volume and Bass Speaker volume to max since Master volume regulation is used

It fixes a case where only the first speaker volume is changed using the system volume control.

### DIFF
--- a/ucm2/HDA/HiFi-analog.conf
+++ b/ucm2/HDA/HiFi-analog.conf
@@ -146,12 +146,19 @@ If.spk {
 				EnableSequence [
 					cset "name='Speaker Playback Switch' on"
 					cset "name='Bass Speaker Playback Switch' on"
+					cset "name='Speaker Playback Volume' 100%"
+					cset "name='Bass Speaker Playback Volume' 100%"
 				]
 
 				DisableSequence [
 					cset "name='Speaker Playback Switch' off"
 					cset "name='Bass Speaker Playback Switch' off"
 				]
+				Value {
+					PlaybackMixerElem "Master"
+					PlaybackVolume "Master Playback Volume"
+					PlaybackSwitch "Master Playback Switch"
+				}
 			}
 			False {
 				EnableSequence [
@@ -161,16 +168,18 @@ If.spk {
 				DisableSequence [
 					cset "name='Speaker Playback Switch' off"
 				]
+				Value {
+					PlaybackMixerElem "Speaker"
+					PlaybackVolume "Speaker Playback Volume"
+					PlaybackSwitch "Speaker Playback Switch"
+				}
 			}
 		}
 
 		Value {
 			PlaybackPriority 100
 			PlaybackPCM "hw:${CardId}"
-			PlaybackMixerElem "Speaker"
 			PlaybackMasterElem "Master"
-			PlaybackVolume "Speaker Playback Volume"
-			PlaybackSwitch "Speaker Playback Switch"
 		}
 	}
 }


tuxedoxt:
Alternative solution suggestions are also appreciated.

tuxedoxt:
Bump, is it feasible to get this included or is an alternative way preferred?

GERJanB:
Is there any update on this when this will be merged?

Prunkles:
This does not work on my Tuxedo Sirius 16 Gen1 under NixOS. I applied this patch using the ALSA_CONFIG_UCM2 environment variable like described [here](https://discourse.nixos.org/t/help-with-avoiding-50gb-worth-of-rebuils/41655/3), and have a patched kernel with [this](https://gitlab.com/tuxedocomputers/development/packages/linux/-/commit/0f49d4c07f572e8d3c9582b9a75b20166fed7695), but with the patch from this PR sound just disappears entirely.
If there is anything I can provide for troubleshooting, I will do so.

tuxedoxt:
@Prunkles 
> This does not work on my Tuxedo Sirius 16 Gen1 under NixOS. I applied this patch using the ALSA_CONFIG_UCM2 environment variable like described [here](https://discourse.nixos.org/t/help-with-avoiding-50gb-worth-of-rebuils/41655/3) (and have a patched kernel with [this](https://gitlab.com/tuxedocomputers/development/packages/linux/-/commit/0f49d4c07f572e8d3c9582b9a75b20166fed7695)) but with the patch from this PR sound just disappears entirely. If there is anything I can provide for troubleshooting, I will do so.

1. Check that alsamixer provides two volume knobs, that is "Bass speakers" in addition to the normal speakers entry for codec "Conexant CX11970" (probably `alsamixer -c 2`), and that the volume knobs regulate the sound of the individual speakers.
2. Providing the speaker knobs are there and works, this PR means to set both at max, and using the system volume control should regulate the master volume knob seen in alsamixer while the two speaker knobs stay at max.

If (1) doesn't work the kernel patch did not apply. If (2) doesn't work this ucm2 config change didn't apply.

perexg:
Could you both attach output from `amixer -c sofhdadsp contents`  command? If this command does not work, replace `sofhdadsp` string with card number. Thanks.

tuxedoxt:
> Could you both attach output from `amixer -c sofhdadsp contents` command? If this command does not work, replace `sofhdadsp` string with card number. Thanks.

Here's from the devices of the original issue. Both use same card and double speaker pairs (sides+top speaker combo in a laptop for context).

[amixer-contents-sirius-gen1.txt](https://github.com/user-attachments/files/17290796/amixer-contents-sirius-gen1.txt)
[amixer-contents-sirius-gen2.txt](https://github.com/user-attachments/files/17290797/amixer-contents-sirius-gen2.txt)

Prunkles:
@perexg
I actually didn't mention that there is no sound anyway on system startup, until I restart wireplumber and pipewire. 

Without this patch, right after the system is started (no sound): [amixer_no-ucm-patch_pre-pw-retart.out.txt](https://github.com/user-attachments/files/17291262/amixer_no-ucm-patch_pre-pw-retart.out.txt)
Without this patch, after pipewire/wireplumber are restarted (is sound): [amixer_no-ucm-patch_post-pw-restart.out.txt](https://github.com/user-attachments/files/17291261/amixer_no-ucm-patch_post-pw-restart.out.txt)
With this patch, right after the system is started (no sound): [amixer_ucm-patch_pre-pw-restart.out.txt](https://github.com/user-attachments/files/17291266/amixer_ucm-patch_pre-pw-restart.out.txt)
With this patch, after pipewire/wireplumber are restarted (no sound) (the output is the same by the way): [amixer_ucm-patch_post-pw-retart.out.txt](https://github.com/user-attachments/files/17291265/amixer_ucm-patch_post-pw-retart.out.txt)

So I guess it may be pipewire/wireplumber issue, but I am not sure how to deal with it either

Prunkles:
> @Prunkles
> 
> > This does not work on my Tuxedo Sirius 16 Gen1 under NixOS. I applied this patch using the ALSA_CONFIG_UCM2 environment variable like described [here](https://discourse.nixos.org/t/help-with-avoiding-50gb-worth-of-rebuils/41655/3) (and have a patched kernel with [this](https://gitlab.com/tuxedocomputers/development/packages/linux/-/commit/0f49d4c07f572e8d3c9582b9a75b20166fed7695)) but with the patch from this PR sound just disappears entirely. If there is anything I can provide for troubleshooting, I will do so.
> 
>     1. Check that alsamixer provides two volume knobs, that is "Bass speakers" in addition to the normal speakers entry for codec "Conexant CX11970" (probably `alsamixer -c 2`), and that the volume knobs regulate the sound of the individual speakers.
> 
>     2. Providing the speaker knobs are there and works, this PR means to set both at max, and using the system volume control should regulate the master volume knob seen in alsamixer while the two speaker knobs stay at max.
> 
> 
> If (1) doesn't work the kernel patch did not apply. If (2) doesn't work this ucm2 config change didn't apply.

There is the "Bass speakers" knob: with or without this PR, 100%-ed, but it doesn't do anything.

tuxedoxt:
> 
> There is the "Bass speakers" knob: with or without this PR, 100%-ed, but it doesn't do anything.

For you it sounds like there's a problem unrelated to this PR then to solve first. To debug. Be sure to turn the other speaker off to hear any difference with the "bass speaker" knob. And of course make sure that it's not muted.

Prunkles:
@tuxedoxt
> For you it sounds like there's a problem unrelated to this PR then to solve first. To debug. Be sure to turn the other speaker off to hear any difference with the "bass speaker" knob. And of course make sure that it's not muted.

Ooh I see. I didn't know you can mute stuff in alsamixer. It solves the entire system mute, but "Bass speakers" still doesn't affect anything

The thing is, I actually once managed to make bass speakers work a long time ago, when I installed the patched tuxedo kernel. And I was able to change volumes of bass and side speakers separately in alsamixer, but when I changed the master knob the other two were changing in a strange manner, so I guess this PR is addressed to exactly this issue.
But when I once tried to plug in and then out headphones, bass speakers stopped to work, and even reboot didn't help. I'm still trying to understand why it stopped working entirely, but I think this is beyond the scope of this PR

perexg:
Those controls exist in all alsa-info dumps:

```
numid=3,iface=MIXER,name='Speaker Playback Volume'
  ; type=INTEGER,access=rw---R--,values=2,min=0,max=74,step=0
  : values=74,74
  | dBscale-min=-74.00dB,step=1.00dB,mute=0
numid=5,iface=MIXER,name='Bass Speaker Playback Volume'
  ; type=INTEGER,access=rw---R--,values=2,min=0,max=74,step=0
  : values=74,74
  | dBscale-min=-74.00dB,step=1.00dB,mute=0
```

I don't like the master volume misuse solution because it may affect also different use-cases (like separate output from Headphones) in future. The goal may be to create a "virtual" control using alsa-lib which maps the `Speaker Volume` to `Bass Speaker Volume`. The question is if 1:1 mapping is sufficient to make the implementation easier for the first code shot. Is 1:1 mapping okay or we need to deal with a more complex mapping (level shifting etc.)?

tuxedoxt:
Thanks for the feedback. Right, those are the two controls in question. The second pair is not technically a bass speaker I believe, but get labelled such on the way somewhere.

>  The question is if 1:1 mapping is sufficient to make the implementation easier for the first code shot. Is 1:1 mapping okay or we need to deal with a more complex mapping (level shifting etc.)?

Yes, 1:1 mapping should be enough for this one case. I do not have enough experience with systems with multiple speaker pairs to judge if level shifting would be necessary at this level in a general case.

I made one more observation however. Taking another look at the HiFi profile and looking at how it is identified I made one test inverting the condition to use it here

https://github.com/alsa-project/alsa-ucm-conf/blob/a03c538f5361b52f4ac2056414c58b0cf3017931/ucm2/HDA/HDA.conf#L22-L31

such that it does not get used.

Fallback one for the card with the multiple speaker pairs gets assigned "Analog Stereo Duplex" and this profile actually does have the behaviour regulating both pairs through the master control (except for 0% where both speakers also get muted). Furthermore there's an "Analog Surround 4.0 Output" profile where the individual levels can be configured. At least to some extent, temporarily.

Ex:

kde-audio-surround40

perexg:
I created PR #533 which should resolve all issues. Please, continue there. The latest (unreleased) alsa-lib is required to support those new extensions.

Note for your last comment: You turned off UCM completely, so you're using legacy pulseaudio/pipewire code. The digital microphone for ACP is probably not available in this mode and also digital microphone for sof-hda-dsp driver is not available.

tuxedoxt:
> I created PR #533 which should resolve all issues. Please, continue there. The latest (unreleased) alsa-lib is required to support those new extensions.
> 
Ah great, I will have a look, thanks.

> Note for your last comment: You turned off UCM completely, so you're using legacy pulseaudio/pipewire code. The digital microphone for ACP is probably not available in this mode and also digital microphone for sof-hda-dsp driver is not available.

Correct, I believe there was a bunch of stuff that didn't work normally. It was more for comparison.

PR #413, opened 2024-04-30
common: pcm: hdmi/sof-hda-dsp/sof-soundwire: Syntax 7 dependent macro to generate the hdmi device sections

ujfalusi:
Hi,

The PR will introduce a generic `HdmiDevice` macro, which can replace duplicated conditions in sof UCM files and potentially used by other devices to generate the needed section for the hdmi PCM devices to allow passthrough.

The PR is marked as DNM and is in draft since the change depends on Syntax 7 support, which is only available in git version of alsa-lib, it is not a simple ask for a user (or developer) to use it and the change would break UCM for released alsa-lib.

@perexg, is this something that you were suggesting? I think the macro and it's definition is generic enough, I hope that it will be usable for non SOF setups.

### DIFF
--- a/ucm2/Intel/sof-hda-dsp/sof-hda-dsp.conf
+++ b/ucm2/Intel/sof-hda-dsp/sof-hda-dsp.conf
@@ -1,4 +1,4 @@
-Syntax 6
+Syntax 7
 
 Include.card-init.File "/lib/card-init.conf"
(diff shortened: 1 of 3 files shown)


perexg:
It looks fine. Let's merge this when alsa-lib is released.

PR #414, opened 2024-05-08
ucm2: Qualcomm: x1e80100: add recording via DMIC01

krzk:
Add support for Digital Microphones DMIC01 on Qualcomm X1E80100 CRD board.


ucm2/Qualcomm/x1e80100/HiFi.conf
```diff
@@ -53,3 +53,17 @@ SectionDevice."Headphones" {
 		JackHWMute "Speaker"
 	}
 }
+
+SectionDevice."DMic01" {
```
perexg, line 57: Naming does not follow convention: https://github.com/alsa-project/alsa-lib/blob/master/include/use-case.h#L135 . Description should be put to the comment only.

krzk: Ack, I'll fix it.

Issue #415, opened 2024-05-08
Motu 828 (28 x 32) support


beatboxa:
The newest [Motu 828 (28x32)](https://motu.com/en-us/products/828/) does not appear to have a UCM yet.  It is a class compliant USB device; so as a result, it generally "works" at a very basic level; but default channel mappings and profiles do not. 

The system appears to default to only 2 channels in and 2 channels out.  So, for example, surround playback seems to mix into either the left or right channel.

In qpwgraph, it shows 30 inputs (I believe this includes an additional 2-channel loopback at the end) and 32 outputs, all marked as "AUX#," where # starts with 0.  I was able to identify which AUX# corresponds to which input or output on the device, through the use of qpwgraph and the device's built-in digital VU meter.

I have tried to create a UCM or even an asoundrc; but I have no idea what I'm doing.  The documentation and examples are not intuitive for me.

Could I collaborate with someone in order to get this audio interface supported?  I'd be happy to run any commands and tests you advise on.

perexg:
Show `amixer -c 0 info` output (replace zero with your card number - can be determined using e.g. `aplay -l`).

And show the channel map (separate for the playback and capture direction). It means which channel is router to which input/output connector on your device.

beatboxa:
output of:     amixer -c 0 info:

```
Card sysdefault:0 'D828'/'MOTU 828 at usb-0000:59:00.0-6, high speed'
  Mixer name	: 'USB Mixer'
  Components	: 'USB07fd:000e'
  Controls      : 1
  Simple ctrls  : 0
```

beatboxa:
Channel Map:  Output
(AUX# on the left is alsa.  Name on the right is from the device)

Images for reference:
![Motu_828-rear_connections](https://github.com/alsa-project/alsa-ucm-conf/assets/6416565/74fd2492-a7ed-4fdc-9eff-8aeb86900e5d)
![Motu_828-front_connections](https://github.com/alsa-project/alsa-ucm-conf/assets/6416565/00563dcf-16e3-41bf-9a52-2ddc8cd04a34)
![Motu_828-connection_list](https://github.com/alsa-project/alsa-ucm-conf/assets/6416565/75a90995-ece9-4fc8-869d-9177db39edb4)

#(On the back of the device, these are labelled "MAIN OUT (A)" and are XLR connectors):
AUX0 = Main Out (L)
AUX1 = Main Out (R)

[...]

PR #416, opened 2024-05-13
MOTU 828 (28x32) ucm2 configuration

beatboxa:
I am trying to create a UCM for the Motu 828 (28 x 32), which was released in 2024.  (Caution: there have been several versions of the 828 over the past few decades.  I am referring to the 2024 model).  

I am not particularly technical when it comes to code and I had to google what a Pull Request is.  And I am struggling on how to do this.  I had previously created an issue here:  https://github.com/alsa-project/alsa-ucm-conf/issues/415

I have made 2 different attempts at creating a ucm configuration:  1 with and 1 without use of the split pcm macros.  I based the 'macro version' loosely on the Motu M6 ucm.  And I created the one without based on a guide I found online.  The version without macros seemed to get further; but for purposes here, I will attempt the Macro version.

Here are the files and changes I have created.  Unfortunately, when I run the command `spa-acp-tool -c 0 -vvvv info`, the results are:
`spa-acp-tool -c 0 -vvvv info

variable '${var:__Device}' is not defined in this context!
error: /USB-Audio/MOTU/D828-HiFi.conf failed to parse device
error: failed to import hw:0 use case configuration -22
variable '${var:__Device}' is not defined in this context!
error: /USB-Audio/MOTU/D828-HiFi.conf failed to parse device
error: failed to import 828 use case configuration -22
UCM not available for card 828
`

At which point the system goes into an alsa fallback / default.

What am I doing wrong?

### DIFF
--- a/ucm2/USB-Audio/MOTU/D828-HiFi.conf
+++ b/ucm2/USB-Audio/MOTU/D828-HiFi.conf
@@ -5,10 +5,12 @@
 #except for the two front microphone ports (each mono)
 #and the two rear optical ports (each is 8-channel ADAT)
 
-#***** Still needs to be done:  ADAT mode of optical ports (in+out, 8-channels, to surround71)
-#I have removed these ports for now
-
-
+#***** Still needs to be done*****:
+#1)  ADAT mode for banks A & B of optical ports (each bank has in + out, 8-channels).  I have removed these ports for now, as the macro only supports 2 channel parameters
+#2)  Related to the above:  a conflict switch so that optical Bank A can switch between 2-channel TOSLink and 8-channel ADAT
+#3)  Conflict switch for front microphones, to switch between 2 mono microphones or a single stereo microphone
+#4)  Surround mapping (7.1 configuration, such that channels 0,1,4,5,6,7,8,9 map to FL,FR,FC,LFE,SL,SR,RL,RR (respectively)
+#5)  Surround mixdown, such that line outs play surround, while headphones mixdown surround to stereo
 
 Include.pcm_split.File "/common/pcm/split.conf"
 
@@ -34,6 +36,7 @@ Macro [
 			HWChannelPos1 MONO
 		}
 	}
+
 	{
 		SplitPCM {
 			Name "motu828_stereo_in"
@@ -246,7 +249,7 @@ SectionDevice."Mic1" {
 }
 
 SectionDevice."Mic2" {
-	Comment "Mic / Line / Instrument In 1"
+	Comment "Mic / Line / Instrument In 2"
 
 	Value {
 		CapturePriority 190
@@ -358,7 +361,7 @@ SectionDevice."Line10" {
 
 #On the back of the MOTU 828, this is a single RCA connector, labeled "S/PDIF IN":
 
-SectionDevice."SPDIF4" {
+SectionDevice."SPDIF5" {
 	Comment "S/PDIF In (RCA)"
 
 	Value {
@@ -382,11 +385,11 @@ SectionDevice."SPDIF4" {
 #   8-channel ADAT
 #in the MOTU 828's menu
 
-SectionDevice."SPDIF5" {
+SectionDevice."SPDIF6" {
 	Comment "Optical A In (S/PDIF)"
 
 	Value {
-		CapturePriority 120
+		CapturePriority 130
 	}
 	Macro.pcm_split.SplitPCMDevice {
 		Name "motu828_stereo_in"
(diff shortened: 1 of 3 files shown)


beatboxa:
After some changes, I also had to apply this patch;  https://github.com/NixOS/nixpkgs/pull/272991

And now, it appears to go further in the chain.  It appears to find the UCM profiles, sets the ucm verb to HiFi, add ports.  Eventually, however, I get to this:

ALSA device (null) roles: null
Check device Mic1 conformance with 0 other devices
First device in combination, number 1
Add port [In] Mic1: Mic / Line / Instrument In 1
Port [In] Mic1 direction input, priority 200
Adding profile HiFi to port [In] Mic1.
Check device Mic1 conformance with 0 other devices
First device in combination, number 1
Port [In] Mic1 direction input, priority 200
ALSA device (null) roles: null
Found 0 jacks.
activate profile: pro-audio (2)
Device: Pro mapping 'Pro' (pro-output-0).
Device: Pro mapping 'Pro' (pro-input-0).
No ELD device found for port [Out] SPDIF1 (0).
No ELD device found for port [Out] Headphones2 (0).
No ELD device found for port [Out] Headphones1 (0).
No ELD device found for port [Out] Line5 (0).
No ELD device found for port [Out] Line4 (0).
No ELD device found for port [Out] Line3 (0).
No ELD device found for port [Out] Line2 (0).
No ELD device found for port [Out] Line1 (0).
No ELD device found for port [In] SPDIF4 (0).
No ELD device found for port [In] Line10 (0).
No ELD device found for port [In] Line9 (0).
No ELD device found for port [In] Line8 (0).
No ELD device found for port [In] Line7 (0).
No ELD device found for port [In] Line6 (0).
No ELD device found for port [In] Mic2 (0).
No ELD device found for port [In] Mic1 (0).
card 0: profiles:3 devices:18 ports:16
    properties: (11)
        use-ucm = "true"
        verbose = "true"
        alsa.card = "0"
        alsa.card_name = "828"
        alsa.long_card_name = "MOTU 828 at usb-0000:59:00.0-9, high speed"
        alsa.driver_name = "snd_usb_audio"
        alsa.mixer_name = "USB Mixer"
        alsa.components = "USB07fd:000e"
        alsa.id = "D828"
        device.string = "0"
        device.description = "828"

beatboxa:
The above configuration seems to work as a baseline.  Note the comments at the beginning of the file:  there are some future features to be done, including adding missing ports--namely two banks of 8-channel ADAT ports--because the macro only appears to work with up to 2 channels of parameters.

beatboxa (commented):
This is a clean and working instance.
However, there are 16 input and 16 output channels missing
(8x Optical A in, 8x Optical A out, 8x Optical B in, 8x Optical B out).

I don't know how to do an 8-channel source or sink.

perexg:
Comments: I added support up to 32 channels / 8 logical channels to PR #417 . Note that this requires recent (not yet released) alsa-lib and `Syntax 7` in `USB-Audio.conf`.

To describe conflicts - use `ConflictingDevice` section. Look to other configuration files for an example. E.g. if two devices are sharing same channels, mark them as conflicting. It's enough to add this section to one conflicting device. The dependencies will be resolved automatically.

perexg:
The sound servers should show the device list sorted according the priority value.  It looks like a problem somewhere else (pipewire or pulseaudio). I suggest to check the list also using the `alsaucm` tool like `alsaucm -c hw:1 dump text | grep -E "(Device.|Priority)"`.

beatboxa:
In testing the results, I don't think the priority has anything to do with the order that the devices are shown.  I can go into the D828-HiFi file, change the order of a device, without changing the priority.

The result is that for the alsaucm command you recommended, the order is consistent with the order written in the D828-HiFi file (top to bottom, not by priority); but the spa-acp-tool and in gnome settings, the order is reversed.

So, for example, in my latest D828-HiFi.conf file, in one part, I have the following consecutive order for 3 devices (top to bottom):

1. Line 3 - Priority 175
2. Mics 1-2 - Priority 195
3. Optical A spdif - Priority 120

In alsaucm, these show up in the order listed above.

However, in spa-acp-tool and in gnome settings (under sound), these show up as:

1. Optical A spdif
2. Mics 1-2
3. Line 3

(reverse ordered)

I had earlier linked a separate issue; and within it, there is this comment that seems to have the same observations as mine:  https://github.com/alsa-project/alsa-lib/issues/245#issuecomment-1168910073  
It appears that somewhere between alsa-lib 1.2.4 and 1.2.7, the probe order has reversed.

For now, I will reorder my devices as I add mono/stereo conflict resolution, so that they show up correctly in my GUI's settings panel.  It is relatively simple to reverse the order again, should there be a future change to alsa-lib or whatever the root cause turns out to be.

beatboxa (commented):
Made some updates:
- Mono / stereo line sources
- Naming convention
- Reordering
- comments

perexg:
1) order is wrong from the configuration POV, the sound servers should keep the UCM device order or sort devices by priority - you should report your issue there

2) UCM device naming has rules - https://github.com/alsa-project/alsa-lib/blob/master/include/use-case.h#L132-L144

PR #423, opened 2024-06-04
ucm2: Qualcomm: x1e80100: add USB DisplayPort playback

krzk:
Add two DisplayPort (over USB) playback devices, conflicting with the headset, because they use the same Multimedia1 frontend.

### DIFF
--- a/ucm2/Qualcomm/x1e80100/HiFi.conf
+++ b/ucm2/Qualcomm/x1e80100/HiFi.conf
@@ -3,7 +3,9 @@
 
 SectionVerb {
 	EnableSequence [
-		cset "name='RX_CODEC_DMA_RX_0 Audio Mixer MultiMedia1' 1"
+		cset "name='DISPLAY_PORT_RX_0 Audio Mixer MultiMedia1' 0"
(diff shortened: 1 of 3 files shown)


ucm2/Qualcomm/x1e80100/HiFi.conf
```diff
@@ -67,3 +81,48 @@ SectionDevice."Mic" {
 		CapturePCM "hw:${CardId},3"
 	}
 }
+
+SectionDevice."DisplayPort0" {
```
krzk, line 85: Maybe this should be called HDMI? https://www.alsa-project.org/alsa-doc/alsa-lib/group___use.html#gab4e4a2b3d44883b1ffb5406529746faf Although technically it is USB/DisplayPort.

perexg:
HDMI device name is only right name for this hardware. I just clarified that HDMI devices are also for DP: https://github.com/alsa-project/alsa-lib/commit/b359ed813e012c38d12960c92940c418859b2806

From the application perspective, the usage is similar (digital streams are equal), just the physical connection is different which should be identified in `Comment` for the user GUI.

krzk:
Ack, thanks. I'll fix it.

PR #430, opened 2024-07-06
Fix labeling of BOSS Katana ports

aledomu:
I'd like to rename the MIDI ports as well but I don't know how. If you can point me where I can learn about it, I'll be glad to do so and mark this PR as finished.

I guess this fix works the same for all Katana models from all generations, but I don't have the PID available to enable it, nor the devices for testing.

Edit: Didn't post the [`alsa-info.sh` output](https://gist.github.com/aledomu/ac3b2741ca2426bc6fbc057907ad0dab).


perexg:
MIDI ports cannot be renamed. You may modify the USB kernel driver for this device.

aledomu:
Nevermind, then.

PR #438, opened 2024-08-21
USB-Audio: Add support for Focusrite 4th Gen devices

ChrisLane:
Extending the existing support for Focusrite Scarlett Solo and 2i2 interfaces with support for the 4th Gen devices.

I have a Focusrite Scarlett 2i2 which is what I have tested with and used to get the device ID.
For the Solo device, I grabbed the device ID from [linux-hardware.org](https://linux-hardware.org/?id=usb:1235-8218).

### DIFF
--- a/ucm2/USB-Audio/Focusrite/Scarlett-2i.conf
+++ b/ucm2/USB-Audio/Focusrite/Scarlett-2i.conf
@@ -55,6 +55,15 @@ If.gen3 {
 	True.Define.Generation "3rd"
 }
 
+If.gen4 {
+	Condition {
+		Type RegexMatch
+		String "${CardComponents}"
+		Regex "USB1235:821[89]"
+	}
+	True.Define.Generation "4th"
+}
+
 Comment "Focusrite Scarlett ${var:Model} ${var:Generation} Gen"
 
 SectionUseCase."HiFi" {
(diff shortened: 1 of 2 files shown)


pv:
Is the `HWChannels 2` for Mic1/2 input direction right in this file, for USB 1235:8218?

We've got a report https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/4597#note_2815906 that looks like `snd_pcm_hw_params_set_channels(pcm_handle, hwparams, 2)` for the CapturePCM "_ucm0001.hw:Gen,0" fails with error. The PCM can be opened with 4 channels however.

ChrisLane:
> See #515
> 
> Is the `HWChannels 2` for Mic1/2 input direction right in this file, for USB 1235:8218?
> 
> We've got a report [gitlab.freedesktop.org/pipewire/pipewire/-/issues/4597#note_2815906](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/4597#note_2815906) that looks like `snd_pcm_hw_params_set_channels(pcm_handle, hwparams, 2)` for the CapturePCM "_ucm0001.hw:Gen,0" fails with error. The PCM can be opened with 4 channels however.
> 
> Shall `CaptureChannels 2` in UCM device be interpreted as that the device can be opened with exactly 2 channels, or that it can be opened with some number of channels >= 2?

I have no idea I'm afraid, I've responded in the open issue you linked.

PR #441, opened 2024-09-12
ucm2: soundwire: rt722: add ConflictingDevice in each device

shumingfan:
This patch allows only one input and one output to be opened simultaneously.

### DIFF
--- a/ucm2/sof-soundwire/rt722.conf
+++ b/ucm2/sof-soundwire/rt722.conf
@@ -10,6 +10,10 @@ If.codecmic {
 		SectionDevice."Mic" {
 			Comment	"SoundWire microphones"
 
+			ConflictingDevice [
+				"Headset"
+			]
+
 			EnableSequence [
 				cset "name='rt722 FU1E Capture Switch' 1"
 			]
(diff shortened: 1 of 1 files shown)


perexg:
What problem are you trying to resolve? If hardware is capable to use all inputs/outputs simultatenously, we should not block this in UCM.

plbossart:
SDCA functions are supposed to operate independently, I am not sure why there would be a conflict @shumingfan  @bardliao? There may be bandwidth restrictions but that's for the Manager to solve with the multi-lane or higher clock rates capabilities.

bardliao:
That is about the bandwidth issue. Currently Intel platforms can support 3 devices work concurrently. We are working on the multi-lane and higher clock rates support. This PR can buy us some time.

plbossart:
the last thing you want is to play the clock with UCM changes, it's nearly impossible to synchronize UCM changes with kernel additions, so the best course of action is to do the right thing at the kernel level.

shumingfan:
@plbossart @bardliao OK, we focus on the kernel level first. closing

PR #445, opened 2024-09-28
ucm2: Add Roland Bridge Cast V2

Fumler:
Roland Bridge Cast recently had a firmware update which brings in e.g. virtual surround and is dubbed as "V2" and has a separate usb device id than the existing Bridge Cast in the repo. This PR adds the changes needed for the new firmware to be mapped correctly, as a new USB audio device.

Changelog: https://www.roland.com/global/support/by_product/bridge_cast/updates_drivers/7e9b7ea2-2c20-4749-a423-0ff28709cdac/

Thanks to [geoffreybennett](https://old.reddit.com/user/geoffreybennett) on reddit for helping me understand how to debug this stuff. Thread: https://old.reddit.com/r/linuxaudio/comments/1fq06z8/roland_bridge_cast_no_longer_having_separate/?cache-bust=1727469957022

### DIFF
--- a/ucm2/USB-Audio/Roland/BridgeCastV2-Hifi.conf
+++ b/ucm2/USB-Audio/Roland/BridgeCastV2-Hifi.conf
@@ -1,0 +1,170 @@
+Include.pcm_split.File "/common/pcm/split.conf"
+
+Macro [
+	{
+		SplitPCM {
+			Name "bc_stereo_out"
+			Direction Playback
+			Channels 2
+			HWChannels 14
+			HWChannelPos0 FL
+			HWChannelPos1 FR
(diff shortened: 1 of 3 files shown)


Fumler:
I opened the PR originally with only 6 channels for the "Game" section, but it seems to have 8, after trying to fix that it's no longer working for me. Will take a look again later.

Fumler:
The PR works as it is now, where the "Game L/R" section is configured with only two channels, but in reality this has eight channels. It stops working for me when I try to configure it with eight, so I must be doing something wrong. I don't really care about the virtual surround so unless you really want it feel free to merge as is.

PR #448, opened 2024-10-10
Add Toradex Hardware Configuration Files

hiagofranco:
Add necessary configuration files related to Toradex Hardware.

### DIFF
--- a/ucm2/IO-Boards/Toradex/apalis/eval-HiFi.conf
+++ b/ucm2/IO-Boards/Toradex/apalis/eval-HiFi.conf
@@ -1,0 +1,21 @@
+# Use case configuration for Toradex Apalis iMX8, Evaluation Board v1.2
+# This is a carrier board for the Apalis family, where any Apalis SoM (with
+# different SoCs as iMX6, iMX8...) can be connected to it.
+
+SectionDevice."Headphones" {
+	Comment "Headphones"
+
+	EnableSequence [
+		cset "name='Headphone Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='Headphone Switch' off"
+	]
+
+	Value {
+		PlaybackPCM "hw:${CardId}"
+		PlaybackVolume "Headphone Volume"
+		PlaybackSwitch "Headphone Switch"
+	}
+}
--- a/ucm2/IO-Boards/Toradex/apalis/eval.conf
+++ b/ucm2/IO-Boards/Toradex/apalis/eval.conf
@@ -1,0 +1,14 @@
+# Use case configuration for Toradex Apalis iMX8, Evaluation Board v1.2
+# This is a carrier board for the Apalis family, where any Apalis SoM (with
+# different SoCs as iMX6, iMX8...) can be connected to it.
+
+Syntax 4
+
+SectionUseCase."HiFi" {
+        File "/IO-Boards/Toradex/apalis/eval-HiFi.conf"
+        Comment "Default"
+}
+
+BootSequence [
+        cset "name='Headphone Volume' 50%"
+]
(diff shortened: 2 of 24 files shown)


ucm2/Toradex/apalis/eval-HiFi.conf
```diff
@@ -0,0 +1,13 @@
...
+	Comment "Headphones"
+
+	EnableSequence [
+		cset "name='Headphone Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='Headphone Switch' off"
+	]
+}
```
perexg, line 13: Overall: Missing PCM device specifications (Values) ....

ucm2/Toradex/apalis/eval.conf
```diff
@@ -0,0 +1,12 @@
...
+
+Syntax 4
+
+SectionUseCase."HiFi" {
+        File "/Toradex/apalis/eval-HiFi.conf"
+        Comment "Default"
+}
+
+BootSequence [
+        cset "name='Headphone Volume' 50%"
```
perexg, line 11: Overall: If hardware supports volume control, it should be speciified e.g. using `PlaybackMixerElem`.

ucm2/Toradex/apalis-imx6/HiFi.conf
```diff
@@ -0,0 +1,13 @@
+# Use case configuration for Toradex Apalis iMX6
```
perexg, line 1: Use ucm2/NXP paths (like current NXP/iMX8 etc.). We prefer to store configurations for similar hardware in one place.

ucm2/Toradex/verdin/dahlia-HiFi.conf
```diff
@@ -0,0 +1,35 @@
+# Use case configuration for Toradex Dahlia Carrier Board
+
+SectionDevice."Headphone" {
+	Comment "Headphone"
+
+	BootSequence [
```
perexg, line 6: Overall: BootSequence must be in top configuration files.

perexg (changes requested):

ucm2/Toradex/apalis-imx6/HiFi.conf
dolcini: @perexg : can you explain better what would you like? These UCM files are related to Toradex board, the actual sound card on these boards does not have much in common with NXP nor i.MX8 boards..

Or maybe you want the boards that use the `fsl-asoc-card` to be grouped together? How?

ucm2/Toradex/verdin/dahlia-HiFi.conf
hiagofranco: Fixed, thanks!

ucm2/Toradex/apalis/eval-HiFi.conf
hiagofranco: I added the `Value` node, please let me know... thanks!

ucm2/Toradex/apalis-imx6/HiFi.conf
perexg: NXP is platform in this case. I think that the best place is something like (subdirectory names):

```
ucm2/iMX6/Toradex/apalis
ucm2/iMX8/Toradex/apalis-eval
ucm2/iMX8/Toradex/apalis
```

etc...

dolcini: Ok, got it.

@perexg : One comment, this does not applies for all this file.

The boards are modular, made of a SoM, that plugs into a carrier board. The carrier boards are generic and different SoM can be plugged into those with a fully modular system.

When the sound card is part of the SoM, what you ask works fine.

Sometime, specifically for the verdin sound card, it is on the carrier board. And the exact same ucm file works whatever the SoM has an NXP soc or a TI one (just a couple of real examples we have). In that case we would have to keep the ucm2 out of the NXP directory.

We would end-up with a mix, some files in ucm2/NXP/${soc}/Toradex/${board}, some files in ucm2/Toradex/${board}.

Is this ok?

perexg: For non-platform specific configs, I would put files to `ucm2/IO-Boards/Toradex` tree. It would make clear, that this config may be shared with multiple SoC combinations. The name is inspired by Raspberry. Thanks.

hiagofranco: Hi, I pushed the requested changes. Please let me know if something needs to be fixed. Thanks.

PR #463, opened 2024-11-13
UCM2: Intel: sof-hda-dsp: Fix handling of empty sys_vendor string

singalsu:
The mistake in UCM syntax caused in alsaucm start error:

ALSA lib ucm_cond.c:367:(if_eval) unknown If.Condition.Type

Fixes: 13022a97711d ("sof-hda-dsp: Fix the case where sysfs dmi
       sys_vendor attribute is not set")


kv2019i (approved):
This seems good. "type empty" exists in asound conf, but I can't see anything for UCM in alsa-lib main, @perexg ?

perexg:
Oops. It was a thinko. I fixed also case where the product_name is empty in 6397c663d7086b87ca5cbba323ea3dcd0ecd3200 . Hopefully, without this mistake.

PR #470, opened 2024-11-24
amd-soundwire: add support for AMD generic legacy machine driver

vijendarmukunda:
Add support for AMD generic legacy(No DSP) machine driver for ACP6.3 platform using RT722 codec.

### DIFF
--- a/ucm2/amd-soundwire/rt722-sdca.conf
+++ b/ucm2/amd-soundwire/rt722-sdca.conf
@@ -1,0 +1,81 @@
+# Use case Configuration for amd-soundwire card
+
+If.RT722 {
+	Condition {
+		Type String
+		Haystack "${CardComponents}"
+		Needle "rt722" 
+	}
+	True {
+		SectionDevice."Headphones" {
+			Comment	"Headphones"
+
+			EnableSequence [
+				cset "name='Headphone Switch' on"
+			]
+
+			DisableSequence [
+				cset "name='Headphone Switch' off"
+			]
+
+			Value {
+				PlaybackPriority 200
+				PlaybackPCM "hw:${CardId}"
+				JackControl "Headphone Jack"
+			}
+		}
+
+		SectionDevice."Headset" {
+			Comment "Headset Microphone"
+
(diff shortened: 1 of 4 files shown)


ucm2/amd-soundwire/rt722-sdca.conf
```diff
@@ -0,0 +1,81 @@
+# Use case Configuration for amd-soundwire card
+
+If.RT722 {
```
perexg, line 3: The config is almost identical to ucm2/sof-soundwire . I would add/merge code. If AMD driver follow the similar components scheme (`${CardComponents}` string), it can be a good benefit to share configs.

vijendarmukunda: @perexg : Agreed. But due to time constraint, We need to release the UCM files on urgent basis to unblock the customer release.  Let this PR  go as initial version. We will look in to your suggested changes and implement the changes as next version.

perexg: Could you show the components string ? `amixer -c 0 info` will print it (replace zero with the correct ALSA card number).

vijendarmukunda:
@perexg : posted v2 version. Please review the PR

perexg:
I really like to use one configuration for all soundwire bridges, because the connected chip and chip configurations are similar. I found only one difference between your config and the config in sof-soundwire directory - `Headset Mic Switch`. Please, retest with changes in PR #478 .

Note that review comments would be, but I prefer #478 way:
 - If.RT722 condition is useless (chip selection is already in the main config file)
 - SectionDevice.InternalMic - wrong device name - https://github.com/alsa-project/alsa-lib/blob/352cbc5eb94a271a9c3c0ff5bf1742232a69e0d0/include/use-case.h#L132-L144
 - priority for internal mic and headset mic is identical - 200
 - whitespace issues

vijendarmukunda:
@perexg : If i understood correctly i need to drop PR#470 changes and I need to test with PR#478 changes correct?

PR #487, opened 2024-12-11
rt722: add headset mic led control

shumingfan:
rt722: add headset mic led control


shumingfan:
According to https://github.com/alsa-project/alsa-ucm-conf/pull/512#issuecomment-2722390040, closing this PR.

PR #490, opened 2024-12-20
USB-Audio: Add support for TASCAM Model 12

Unveiled53:
Adds support for the TASCAM Model 12 USB multi-track mixer. Couple of things to note:
- This profile assumes the "USB AUDIO" setting on the mixer itself is set to "MULTI INPUT".
- Under the "MULTI INPUT" mode, this mixer has 12 logical input channels (Last two are a stereo mix) and 10 logical output channels.
- Most of the accessible I/O is combo XLR/TRS jacks, so I opted to have the mono capture devices (With the exception of physical inputs 8 and 10 which are TRS only) named as "Mic" with the rest as "Line".

Otherwise, the profile is tested and working with the latest USB-Audio profiles.

[alsa-info.txt](https://github.com/user-attachments/files/18207924/alsa-info.txt)

### DIFF
--- a/ucm2/USB-Audio/TASCAM/Model12-HiFi.conf
+++ b/ucm2/USB-Audio/TASCAM/Model12-HiFi.conf
@@ -1,0 +1,446 @@
+Include.pcm_split.File "/common/pcm/split.conf"
+
+Macro [
+	{
+		SplitPCM {
+			Name "model12_mono_in"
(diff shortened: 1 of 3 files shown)


perexg:
Could you add a proper `Signed-off-by:` tag to your commit?

Unveiled53:
Should be done.

PR #502, opened 2025-01-25
Add UCM for MT8183 Chromebooks

WeirdTreeThing:


perexg:
Please, use more readable directory names (not stripped) in ucm2/MediaTek tree. Thanks.

PR #504, opened 2025-01-31
UCM2: Blobs: SOF: Cleanup blob names from .blob to .bin

singalsu:
The configuration blobs for SOF processing components are mixed with old filename suffix .blob and new .bin. This patch renames for cleanup the remaining .blob files to .bin to follow the chosen naming in SOF project.


kv2019i (approved):

perexg:
Applied. Thanks.

PR #506, opened 2025-02-07
rt722: change output volume of speaker to 0dB

shumingfan:
This patch also changes hardware gain to software gain when using volume bar.

### DIFF
--- a/ucm2/codecs/rt722/init.conf
+++ b/ucm2/codecs/rt722/init.conf
@@ -5,7 +5,7 @@ BootSequence [
 	cset "name='rt722 ADC 25 Mux' 'DMIC2'"
 	cset "name='rt722 ADC 22 Mux' 'MIC2'"
 	cset "name='rt722 FU1E Capture Volume' 63"
-	cset "name='rt722 FU06 Playback Volume' 60"
+	cset "name='rt722 FU06 Playback Volume' 87"
 	cset "name='rt722 FU05 Playback Volume' 87"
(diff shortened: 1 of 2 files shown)


perexg:
> This patch also changes hardware gain to software gain when using volume bar.

Why this change is required ? Hardware volume is always preferred.

perexg:
Thanks for explanation. Could you add this information to the comment in commit ?

perexg:
The mentioned range is from -65.25dB to 0dB as defined in TLV. It should be enough to have decent volume range, so I wonder what you mean with 100% scale. The lowest value (0) is not so quiet? Then does -65.25dB match the codec output?

shumingfan:
> The mentioned range is from -65.25dB to 0dB as defined in TLV. It should be enough to have decent volume range, so I wonder what you mean with 100% scale. The lowest value (0) is not so quiet? Then does -65.25dB match the codec output level? If not, the driver should be corrected at first.

There is no problem with the codec output level.  
The customer might make a question about some volume percentage getting the same value of "rt722 FU06 playback volume" control.

shumingfan:
> Pipewire debug:
> 
> ```
> systemctl --user stop wireplumber
> WIREPLUMBER_DEBUG="spa.*:5" wireplumber 2>&1 | grep set_volume
> ```

@perexg Thanks for explanation. I understood.
Will share this information with the customer.
Therefore, I would like to close the PR. If needed, I will re-open the PR.

PR #520, opened 2025-03-13
Split the GoXLR configuration into Full Device and Mini

FrostyCoolSlug:
This commit splits the GoXLR and GoXLR Mini into separate configurations.

The GoXLR Mini will always (at this point) have 21 channels defined, where as the full sized GoXLR will have either 23 or 25 depending on Firmware (23 is FAR more common at this point, as the firmware with 25 channels is still in beta).

Separating out the two devices should at least partially solve the recent issue with Pipewire.


FrostyCoolSlug:
Closing as this is superseded by #521

PR #525, opened 2025-03-17
sof-soundwire: rt1318: add playback control switch

shumingfan:
rt1318: add playback control switch

[v2] This patch defines a remapped one 'rt1318 DAC Playback Switch' whether the cases are in one amp or two amps.
[v3] This patch defines 'rt1318-1 DAC' as PlaybackMixerElem when equipped with only one amp
and defines a remapped one 'rt1318 DAC Playback Switch' when using two amps.

### DIFF
--- a/ucm2/codecs/rt1318/init.conf
+++ b/ucm2/codecs/rt1318/init.conf
@@ -1,5 +1,26 @@
 # RT1318 specific switch control settings
 
+Define.SpeakerMixerElem "rt1318-1 DAC"
+
+If.twoAmpsStereoToOne {
+	Condition {
+		Type RegexMatch
+		Regex "2"
+		String "${var:SpeakerAmps}"
+	}
+	True {
+		Define.SpeakerMixerElem "rt1318 DAC"
+		LibraryConfig.remap.Config {
+			ctl.default.map {
+				"name='rt1318 DAC Playback Switch'" {
+					"name='rt1318-1 DAC Switch'".vindex.0 [ 0 1 ]
+					"name='rt1318-2 DAC Switch'".vindex.1 [ 0 1 ]
+				}
+			}
+		}
+	}
+}
+
 If.oneAmp {
 	Condition {
 		Type ControlExists
--- a/ucm2/sof-soundwire/rt1318.conf
+++ b/ucm2/sof-soundwire/rt1318.conf
@@ -41,5 +41,7 @@ SectionDevice."Speaker" {
 	Value {
 	      PlaybackPriority 100
 	      PlaybackPCM "hw:${CardId},2"
+	      PlaybackMixer "default:${CardId}"
+	      PlaybackMixerElem "${var:SpeakerMixerElem}"
 	}
 }


perexg:
It's not complete. For stereo, only second amplifier switch will control LED. You should define a remapped (combined) stereo control using two mono switches - like: https://github.com/alsa-project/alsa-ucm-conf/blob/master/ucm2/codecs/cs42l43-dmic/init.conf and use it for `PlaybackMixerElem`.

shumingfan:
> It's not complete. For stereo, only second amplifier switch will control LED. You should define a remapped (combined) stereo control using two mono switches - like: https://github.com/alsa-project/alsa-ucm-conf/blob/master/ucm2/codecs/cs42l43-dmic/init.conf and use it for `PlaybackMixerElem`.

@perexg  The 'rt1318-1 DAC Switch' and 'rt1318-2 DAC Switch' already are stereo controls, not mono switches.
numid=52,iface=MIXER,name='rt1318-1 DAC Switch'
  ; type=BOOLEAN,access=rw------,values=2
  : values=on,on

Does it still need to do that?

perexg:
> Does it still need to do that?

Yes, you have two stereo controls which should be mapped to one stereo control for the application. Unfortunately, I added support for this mapping to the control remap plugin today - https://github.com/alsa-project/alsa-lib/commit/6855cb838d0a69183e1db6429bc9c14265b68240 . So the correct config should be like:

```
# init section
LibraryConfig.remap.Config {
	ctl.default.map {
		"name='rt1318 DACs Playback Switch" {
			"name='rt1318-1 DAC Switch'".vindex.0 [ 0 1 ]
			"name='rt1318-2 DAC Switch'".vindex.1 [ 0 1 ]
		}
	}
}

# then use
CaptureMixer "default:${CardId}"
PlaybackMixerElem "rt1318 DACs"
```

For one amp, you can use `PlaybackMixerElem "rt1318-1 DAC"` directly.

shumingfan:
> ```
> # init section
> LibraryConfig.remap.Config {
> 	ctl.default.map {
> 		"name='rt1318 DACs Playback Switch" {
> 			"name='rt1318-1 DAC Switch'".vindex.0 [ 0 1 ]
> 			"name='rt1318-2 DAC Switch'".vindex.1 [ 0 1 ]
> 		}
> 	}
> }
> 
> # then use
> PlaybackMixer "default:${CardId}"
> PlaybackMixerElem "rt1318 DACs"
> ```
> 
> For one amp, you can use `PlaybackMixerElem "rt1318-1 DAC"` directly.

@perexg I have updated the PR. PlaybackMixerElem uses "rt1318 DAC" whether one amp or two amps.

ucm2/codecs/rt1318/init.conf
```diff
@@ -1,5 +1,38 @@
...
+}
+
+If.twoAmpsStereoToOne {
+	Condition {
+		Type RegexMatch
+		Regex "2"
+		String "${var:SpeakerAmps}"
+	}
+	True {
+		LibraryConfig.remap.Config {
```
perexg, line 25: Add `Define.SpeakerMixerElem "rt1318 DAC"` here.

ucm2/codecs/rt1318/init.conf
```diff
@@ -1,5 +1,38 @@
 # RT1318 specific switch control settings
 
+If.oneAmpRemap {
```
perexg, line 3: It would be better to define a new default variable like:

```
Define.SpeakerMixerElem "rt1318-1 DAC"
```

and remove this whole If block.

ucm2/sof-soundwire/rt1318.conf
```diff
@@ -41,5 +41,7 @@ SectionDevice."Speaker" {
 	Value {
 	      PlaybackPriority 100
 	      PlaybackPCM "hw:${CardId},2"
+	      PlaybackMixer "default:${CardId}"
+	      PlaybackMixerElem "rt1318 DAC"
```
perexg, line 45: Use `PlaybackMixerElem "${var:SpeakerMixerElem}"` here.

shumingfan: > Use `PlaybackMixerElem "${var:SpeakerMixerElem}"` here.

@perexg Thanks for the review. 
I updated this PR already.

PR #526, opened 2025-03-18
ucm2: HDA: HiFi-analog/mic: Refactor the analog mic discovery and pre…

ujfalusi:
…sentation

The current mic device creation works on certain machines and fails on others. There are several places of conflicts and setups which can only just fail, but this is mostly not an issue if the user never uses the mic, only the speaker/headset - which, to be honest is what most of us do ;)

As an example:
The mic selection in most codecs are via enum and it is assumed to be named 'Input Source', which is not always the case as some device uses 'Capture Source' for the control's name.
There is also different sets of mics that one can select from: Exhibit A
numid=6,iface=MIXER,name='Input Source'
  ; type=ENUMERATED,access=rw------,values=1,items=2
  ; Item #0 'Headset Mic'
  ; Item #1 'Headphone Mic'
  : values=1

Exhibit B
numid=6,iface=MIXER,name='Capture Source'
  ; type=ENUMERATED,access=rw------,values=1,items=2
  ; Item #0 'Internal Mic'
  ; Item #1 'Headset Mic'
  : values=0

Exhibit C
numid=6,iface=MIXER,name='Capture Source'
  ; type=ENUMERATED,access=rw------,values=1,items=3
  ; Item #0 'Internal Mic'
  ; Item #1 'Headset Mic'
  ; Item #2 'Headphone Mic'
  : values=0

Exhibit D (this pushes the limits... The patch will ignore item 1) numid=6,iface=MIXER,name='Capture Source'
  ; type=ENUMERATED,access=rw------,values=1,items=3
  ; Item #0 'Internal Mic'
  ; Item #1 'Internal Mic 1'
  ; Item #2 'Mic'
  : values=2

Other issue is that we have this 'Headphone Mic', which turned out to be a 'Stereo Microphone in Headphone Jack', so if it is selected then the Headphone cannot work, they conflict, they use the same rings for different direction and purpose.

This patch aims to make the mic discovery a bit more deterministic and pragmatic.

But even if the UCM creates the use case profiles correctly, it is still up to UIs (KDE/GNOME/etc) to misunderstand how UCM presents the profiles, what they mean and most of all what 'Mic1', `Mic2', etc is. KDE presents the profiles as they are and user can selct between them to pick the right combination of output and input.
GNOME goes further with simplification (and fails with it) and presents 'random' Configuration profiles for Output and Input, plus a device selection and they do work in an interesting way. GNOME also have popup for specifying the type of the plugged accessory, which does not worl at all with UCM profiles.

But, this patch is meant for a small step to have clear rules based mic presentation for HDA.

The expectation is that what have worked will work as it used to and what did not worked should be detected and presented correctly.

### DIFF
--- a/ucm2/HDA/HiFi-mic.conf
+++ b/ucm2/HDA/HiFi-mic.conf
@@ -1,0 +1,432 @@
+# Generic handling of analog microphones on HDA devices
+
+# Microphone related variables on entry
+# If DMIC is present:
+#   (DeviceDmic == "Mic1")
+#   DeviceMic == "Mic2"
+#
+# If DMIC is not present:
+#   DeviceMic == "Mic"
+
+# Microphone related variables as used internally
+# HDA codecs can offer input source selection (mux) via "Input Source" or
+# "Capture Source" with varying options:
+#  'Internal Mic'
+#  'Headset Mic'
+#  'Headphone Mic'
+#  'Mic'
+#
+# or there can be 'Mic' and optional 'Front Mic' available
+#
+# This leads to up to maximum of four possible input devices.
+
+Define.SourceControl ""
+Define.MicJackControl ""
+
+# DeviceMic is pre-configured
+Define.DeviceMicName ""
+Define.DeviceMicComment ""
+Define.DeviceMicPriority ""
+Define.DeviceMicJack ""
+Define.DeviceMic2 ""
+Define.DeviceMic2Name ""
+Define.DeviceMic2Comment ""
+Define.DeviceMic2Priority ""
+Define.DeviceMic2Jack ""
+Define.DeviceMic3 ""
+Define.DeviceMic3Name ""
+Define.DeviceMic3Comment ""
+Define.DeviceMic3Priority ""
+Define.DeviceMic3Jack ""
+Define.DeviceMic4 ""
+Define.DeviceMic4Name ""
+Define.DeviceMic4Comment ""
+Define.DeviceMic4Priority ""
+Define.DeviceMic4Jack ""
+
+# evaluate the microphone jack name
+If.hsmicjack {
+	Condition {
+		Type ControlExists
+		Control "iface=CARD,name='Headset Mic Jack'"
+	}
+	True.Define.MicJackControl "Headset Mic Jack"
+	False.If.hpmicjack {
+		Condition {
+			Type ControlExists
+			Control "iface=CARD,name='Headphone Mic Jack'"
+		}
+		True.Define.MicJackControl "Headphone Mic Jack"
+		False.If.hpjack {
+			Condition {
+				Type ControlExists
+				Control "iface=CARD,name='Headphone Jack'"
+			}
+			True.Define.MicJackControl "Headphone Jack"
+			False.If.micjack {
+				Condition {
+					Type ControlExists
+					Control "iface=CARD,name='Mic Jack'"
+				}
+				True.Define.MicJackControl "Mic Jack"
+			}
+		}
+	}
+}
+
+# determine the name of the source selection enum, if present
+If.inputsourcectl {
+	Condition {
+		Type ControlExists
+		Control "name='Input Source'"
+	}
+	True.Define.SourceControl "Input Source"
+	False.If.capturesourcectl {
+		Condition {
+			Type ControlExists
+			Control "name='Capture Source'"
+		}
+		True.Define.SourceControl "Capture Source"
+	}
+}
+
+# Set initial names and priorities for the SectionDevices based on the presence
+# of DMIC (DMIC is named as 'Mic1')
+If.micdevs {
+	# DeviceMic == "Mic2" indicates DMIC presence
+	Condition {
+		Type String
+		String1 "${var:DeviceMic}"
+		String2 "Mic2"
+	}
+	True.Define {
+		# DeviceMic "Mic2" - already set (Mic1 is DMIC)
+		DeviceMicPriority 200
+		DeviceMic2 "Mic3"
+		DeviceMic2Priority 300
+		DeviceMic3 "Mic4"
+		DeviceMic3Priority 400
+		DeviceMic4 "Mic5"
+		DeviceMic4Priority 500
+	}
+	False.Define {
+		DeviceMic "Mic1"
+		DeviceMicPriority 100
+		DeviceMic2 "Mic2"
+		DeviceMic2Priority 200
+		DeviceMic3 "Mic3"
+		DeviceMic3Priority 300
+		DeviceMic4 "Mic4"
+		DeviceMic4Priority 400
+	}
+}
+
+# Scan the Microphone configuration in the system and set the definitions for
+# the SectionDevice creations for them.
+If.micsetup {
+	Condition {
+		Type String
+		Empty "${var:SourceControl}"
+	}
+	False {
+		# the input selection is via ENUM, the possible types are
+		# Internal, Headset, Headphone and just 'Mic'
+
+		# Internal Mic
+		If.internalmic {
+			Condition {
+				Type ControlExists
+				Control "name='${var:SourceControl}'"
+				ControlEnum "Internal Mic"
+			}
+			True.Define {
+				DeviceMicName "Internal Mic"
+				DeviceMicComment "Internal Stereo Microphone"
+			}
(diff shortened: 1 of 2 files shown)


singalsu (approved):
I've tested this to work on three different HDA laptops running SOF and UCM in Fedora 41 from three different brands, one HDA IPC4 MTL device, one HDA+DMIC IPC4 TGL-H device, one HDA+DMIC IPC3 CNL device.

ucm2/HDA/HiFi-mic.conf
```diff
@@ -0,0 +1,409 @@
...
+			Type String
+			Empty "${var:__JackName}"
+		}
+		False.Value.JackControl "${var:__JackName}"
+	}
+
+	If.conflict1 {
+		Condition {
+			Type String
+			Empty "${var:__ConflictCondition1}"
```
perexg, line 304: Use `Empty "${var:-__ConflictCondition1}"` here and remove `ConflictCondition3 "" ConflictDev3 ""` lines from callers. The minus means empty string if the variable does not exists.

perexg (changes requested):
Nice cleanup. Pls, follow my suggestion.

perexg:
This is really nice extension which shows all trouble with the device naming for the dynamic configuration tree evaluation. 

I plan to do new extension for UCM - device labels. The idea is to use something like "Mic@internal_mic" in configuration with priority table like 'DevicePriority [ "Mic@internal_mic" "Mic@front_mic" ]' etc and alsa-lib will keep the device order for applications - "Mic1", "Mic2". It will greatly simplifies those cases.

ujfalusi:
> This is really nice extension which shows all trouble with the device naming for the dynamic configuration tree evaluation.
> 
> I plan to do new extension for UCM - device labels. The idea is to use something like "Mic@internal_mic" in configuration with priority table like 'DevicePriority [ "Mic@internal_mic" "Mic@front_mic" ]' etc and alsa-lib will keep the device order for applications - "Mic1", "Mic2". It will greatly simplify those cases.

I think there a Mics without position, the 'Mic' for example..

But for the section device naming. Currently this is how it works:
The UCM checks for the presence of 'Headphone Mic' and if it is found it receives Mic2 (if there is DMIC) / Mic1 (no DMIC) and a special 'Headset' section device is also created. This looks like this on Dell G15 (w/ DMIC):
```
numid=7,iface=MIXER,name='Input Source',index=1
  ; type=ENUMERATED,access=rw------,values=1,items=2
  ; Item #0 'Headset Mic'
  ; Item #1 'Headphone Mic'
  : values=1

Device.Mic1 {
	Comment "Digital Microphone"
}
Device.Mic2 {
	Comment "Headphones Stereo Microphone"
	ConflictingDevices [
		Headset
	]
}
Device.Headset {
	Comment "Headset Mono Microphone"
	ConflictingDevices [
		Mic2
	]
}
```
On a Framework laptop w/o DMIC:
```
numid=6,iface=MIXER,name='Capture Source'
  ; type=ENUMERATED,access=rw------,values=1,items=2
  ; Item #0 'Internal Mic'
  ; Item #1 'Headset Mic'
  : values=0

Device.Mic1 {
	Comment "Headphones Stereo Microphone"
	ConflictingDevices [
		Mic2
	]
}
Device.Mic2 {
	Comment "Internal Stereo Microphone"
	ConflictingDevices [
		Mic1
	]
}
```
On Framework we have 'Capture Source' and no 'Input Source' so the headset is missed, but even if UCM was checking the correct enum control, there is no 'Headphone Mic' as selection (only 'Headset Mic')

With this patch the order and naming will be different (and correct o for Framework):
G15
```
Device.Mic1 {
	Comment "Digital Microphone"
}
Device."Headset Mic" {
	Comment "Headset Mono Microphone"
	ConflictingDevices [
		Mic3
	]
}
Device.Mic3 {
	Comment "Stereo Microphone in Headphone Jack"
	ConflictingDevices [
		"Headset Mic",
		Headphones
	]
}
```
Framework (correct now):
```
Device.Mic1 {
	Comment "Internal Stereo Microphone"
	ConflictingDevices [
		"Headset Mic"
	]
}
Device."Headset Mic" {
	Comment "Headset Mono Microphone"
	ConflictingDevices [
		Mic1
	]
}
```
Does this change in names matters? Keeping Mic1/2/3/4 order could be possible if I evaluate the 'Headphone Mic' first if that is better.

perexg:
> I think there a Mics without position, the 'Mic' for example..

If there is only one device for the given group, the number can be omitted.

> But for the section device naming. Currently this is how it works: The UCM checks for the presence of 'Headphone Mic' 
and if it is found it receives Mic2 (if there is DMIC) / Mic1 (no DMIC) and a special 'Headset' section device is also created. 

It looks correct, but...

> Device."Headset Mic" {

That's wrong. It should be just 'Headset'. See naming conventions - https://github.com/alsa-project/alsa-lib/blob/master/include/use-case.h#L132 .

> Does this change in names matters? Keeping Mic1/2/3/4 order could be possible if I evaluate the 'Headphone Mic' first if that is better.

The numbering must be continuous. See comments in the provided link.

ujfalusi:
> > Device."Headset Mic" {
> 
> That's wrong. It should be just 'Headset'. See naming conventions - https://github.com/alsa-project/alsa-lib/blob/master/include/use-case.h#L132 .

Hrm, so it should be MicX, no? For me the Headset is an accessory with headphone and mic, so is something which can be output and input.

> > Does this change in names matters? Keeping Mic1/2/3/4 order could be possible if I evaluate the 'Headphone Mic' first if that is better.
> 
> The numbering must be continuous. See comments in the provided link.

Ok. Do we need to special case the Mic which is on Headset and use `SectionDevice."Headset"` instead of the next Mic number suffix?

perexg:
> > > Device."Headset Mic" {
> > 
> > 
> > That's wrong. It should be just 'Headset'. See naming conventions - https://github.com/alsa-project/alsa-lib/blob/master/include/use-case.h#L132 .
> 
> Hrm, so it should be MicX, no? For me the Headset is an accessory with headphone and mic, so is something which can be output and input.

Headset and handsets are special cases. They do not belong to a generic microphone input type. Device names are used to determine the shared jack type for applications (e.g. jack type selection dialog in gnome). They are not looking into comments (description). You can define those as one duplex device or separate ones for playback and capture if they are independent (names like 'Headset1' and 'Headset2').

Also, the headphones mic may be marked as 'Headphone' to correctly identify 

See also: https://gitlab.gnome.org/GNOME/libgnome-volume-control/-/blob/master/gvc-mixer-control.c?ref_type=heads#L2370-2372

> Ok. Do we need to special case the Mic which is on Headset and use `SectionDevice."Headset"` instead of the next Mic number suffix?

Yes.

ucm2/HDA/HiFi-mic.conf
```diff
@@ -0,0 +1,451 @@
...
+Define.DeviceMic4Name ""
+Define.DeviceMic4Comment ""
+Define.DeviceMic4Priority ""
+Define.DeviceMic4Jack ""
+
+# evaluate the microphone jack name
+If.micjackctl {
+	Condition {
+		Type ControlExists
+		Control "iface=CARD,name='Headset Mic Phantom Jack'"
```
perexg, line 51: I would not use the "software defined" Phantom Jacks at all. They were defined for pulseaudio, because their port selection algorithm was designed in a simple way. The conflicting device links are resolving this issue. This block can just use other controls to do similar job.

I think that this decision tree should fit:

1. Headset Mic Jack
2. Headphone Mic Jack
3. Headphone Jack
4. Mic Jack

perexg (changes requested):

perexg:
> IOW, if we had Headset mic selected and next time plugging in an accessory and selecting Headphone, the Headset Mic is still selected, which is mostly fine as users wont swap accessories much, but for purely technical stand, it is annoying.
> The manual selection via the Settings->Sound works flawlessly

It looks like a gnome volume control issue.

ujfalusi:
Changes since v2:
- change the decision order for Mic Jack as per advice from @perexg (to: HS Mic, HP Mic, HP then Mic)

perexg:
Thank you for your work. I applied it with a little cleanup in aaa7455aaf36537cc7b198fabbc8cdef31ea7371 .

PR #531, opened 2025-03-24
ucm2: Qualcomm: add HP Omnibook X14  support

jglathe:
assume same as T14s (2 speakers)


ucm2/Qualcomm/x1e80100/x1e80100.conf
```diff
@@ -17,3 +17,12 @@ If.LENOVOSlim7x {
...
 	True.Include.7x.File "/Qualcomm/x1e80100/LENOVO-Slim-7x.conf"
 }
+
+If.HPOMNIBOOKX14 {
+	Condition {
+		Type RegexMatch
+		String "${sys:devices/virtual/dmi/id/board_vendor}-${sys:devices/virtual/dmi/id/product_family}"
+		Regex "HP.*Omnibook*X*"
+	}
+	True.Include.x14.File "/Qualcomm/x1e80100/HP-Omnibook-x14.conf"
```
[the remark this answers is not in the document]
obbardc: sorry, i am wrong

perexg:
Handle this with an extension for regex in `Regex "LENOVO.*ThinkPad T14s Gen 6.*"` . It may be one line change. There is no diff between `HP-Omnibook-x14.conf` and `LENOVO-T14s.conf`.

It would be probably nice to use macros for Qualcomm codecs/ files and reduce the number of included files anyway. Many things can be just handed using macro arguments or variables.

jglathe:
There will be more of these, probably. On my list: Acer Swift 14, Lenovo Thinkbook 16. I guess they will have the same Thinkpad T14s setup, only different DMI data. I am not sure if they will stay the same over time as more info might be available.

jglathe:
Ok I tried around a bit but failed. I was probably holding it wrong :tm:, anyway. Reduced the changes to device detection, but with extra section for the device. Also, I have a positive report for the HP Elitebook G1q , would be another section. If this is feasible, I would lie to extend x1e80100.conf device by device with one section each (in another PR).

ucm2/Qualcomm/x1e80100/x1e80100.conf
```diff
@@ -1,12 +1,12 @@
 Syntax 4
 
-If.LENOVOX14s {
+If.LENOVOT14s {
 	Condition {
 		Type RegexMatch
 		String "${sys:devices/virtual/dmi/id/board_vendor}-${sys:devices/virtual/dmi/id/product_family}"
 		Regex "LENOVO.*ThinkPad T14s Gen 6.*"
```
perexg, line 7: Change to `Regex "(LENOVO.*ThinkPad T14s Gen 6.*)|(HP.*Omnibook X.*)"` and remove new If block.

ucm2/Qualcomm/x1e80100/x1e80100.conf
```diff
@@ -1,12 +1,12 @@
 Syntax 4
 
-If.LENOVOX14s {
+If.LENOVOT14s {
 	Condition {
 		Type RegexMatch
 		String "${sys:devices/virtual/dmi/id/board_vendor}-${sys:devices/virtual/dmi/id/product_family}"
```
perexg, line 6: It may be useful to define a new variable like:

```
Define.DMI_info "${sys:devices/virtual/dmi/id/board_vendor}-${sys:devices/virtual/dmi/id/product_family}"
```

And use it in condition like `String "${var:DMI_info}"`.
```

perexg (changes requested):

ucm2/Qualcomm/x1e80100/x1e80100.conf
jglathe: Well I tried just that the last time, and it didn't work. Will try again.

jglathe: And this time it worked :grin: Thank you. I had to redo the whole box first, because... reasons

ucm2/Qualcomm/x1e80100/x1e80100.conf
jglathe: Okay this worked out well, too. Thanks!

jglathe:
How about the other models that will pour in over time. All to the same regex, PR by PR?

perexg:
Applied. Thanks. Please rebase and squash commits to one next time.

PR #536, opened 2025-03-27
USB-Audio: Add Roland BridgeCast One | adapted

Simon99de:
Add Roland BridgeCast One
uses the same config as the bridge cast v2 as it has the same outputs and it worked in my test.

0582:030d


perexg:
Use `Regex` match in `If.roland-bridgecastv2' and extend it for both IDs. Thank you.

Simon99de:
> Use `Regex` match in `If.roland-bridgecastv2' and extend it for both IDs. Thank you.

thank you very much, has been updated/rewritten and both devices have been tested successfully. (currently I still have both)

PR #540, opened 2025-04-02
sof-soundwire: cs42l43: Switch mixer based on output volume

mstrozek:
Hardware volume control in cs42l43 can go up to +31.5dB which can cause audio degradation. For best results, limit hardware volume to 0dB.

Link: https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/4622


ucm2/sof-soundwire/cs42l43-spk.conf
```diff
@@ -1,5 +1,14 @@
 # Use case Configuration for sof-soundwire card
 
+LibraryConfig.remap.Config {
```
perexg, line 3: This should be moved to `ucm2/codecs/cs42l43/init.conf ` .

ucm2/sof-soundwire/cs42l43-spk.conf
```diff
@@ -17,9 +26,30 @@ SectionDevice."Speaker" {
...
+			Value {
+				PlaybackMixerElem "cs42l43 Speaker Digital"
+			}
+		}
+		False {
+			Value {
+				PlaybackMixer "default:${CardId}"
+				PlaybackMixerElem "cs42l43 Speaker"
+			}
+			BootSequence [
```
perexg, line 45: This MUST be moved to ucm2/codecs/cs42l43/init.conf . BootSequence is not allowed here.

perexg (changes requested):

ucm2/sof-soundwire/cs42l43-spk.conf
mstrozek: moved

ucm2/sof-soundwire/cs42l43-spk.conf
mstrozek: moved

KaiChuan-Hsieh:
After applying the change, there is no audio output when volume is under 40%. Is it acceptable?

mstrozek:
@KaiChuan-Hsieh that does not seem acceptable. Can you please check what is the value of 'cs42l43 Speaker Digital Volume' control (for example with "amixer -D hw cget name='cs42l43 Speaker Digital Volume'") on the affected system? Might be good to also check if this control changes with the system volume.

KaiChuan-Hsieh:
> @KaiChuan-Hsieh that does not seem acceptable. Can you please check what is the value of 'cs42l43 Speaker Digital Volume' control (for example with "amixer -D hw cget name='cs42l43 Speaker Digital Volume'") on the affected system? Might be good to also check if this control changes with the system volume.

The output.
ubuntu@localhost:~$ amixer -c 1 cget name='cs42l43 Speaker Digital Volume'
numid=43,iface=MIXER,name='cs42l43 Speaker Digital Volume'
  ; type=INTEGER,access=rw---R--,values=2,min=0,max=191,step=0
  : values=128,128
  | dBscale-min=-64.00dB,step=0.50dB,mute=0

KaiChuan-Hsieh:
> @KaiChuan-Hsieh that does not seem acceptable. Can you please check what is the value of 'cs42l43 Speaker Digital Volume' control (for example with "amixer -D hw cget name='cs42l43 Speaker Digital Volume'") on the affected system? Might be good to also check if this control changes with the system volume.

Before applying the commit:
`ubuntu@ubuntu:~/Downloads$ amixer -c 0 cget name='cs42l43 Speaker Digital Volume'
numid=40,iface=MIXER,name='cs42l43 Speaker Digital Volume'
  ; type=INTEGER,access=rw---R--,values=2,min=0,max=191,step=0
  : values=144,144
  | dBscale-min=-64.00dB,step=0.50dB,mute=0
ubuntu@ubuntu:~/Downloads$ apt policy alsa-ucm-conf
alsa-ucm-conf:
  Installed: 1.2.10-1ubuntu5.4
  Candidate: 1.2.10-1ubuntu5.4
  Version table:
 *** 1.2.10-1ubuntu5.4 500
        500 http://archive.ubuntu.com/ubuntu noble-updates/main amd64 Packages
        500 http://archive.ubuntu.com/ubuntu noble-updates/main i386 Packages
        100 /var/lib/dpkg/status
     1.2.10-1ubuntu5 500
        500 http://archive.ubuntu.com/ubuntu noble/main amd64 Packages
        500 http://archive.ubuntu.com/ubuntu noble/main i386 Packages`

After applying the commit:
`ubuntu@ubuntu:~$ amixer -c 0 cget name='cs42l43 Speaker Digital Volume'
numid=45,iface=MIXER,name='cs42l43 Speaker Digital Volume'
  ; type=INTEGER,access=rw---R--,values=2,min=0,max=191,step=0
  : values=128,128
  | dBscale-min=-64.00dB,step=0.50dB,mute=0`

KaiChuan-Hsieh:
@mstrozek 
Hello,

May I know why original issue can't be eliminated by calling the function in codec driver. https://github.com/torvalds/linux/blob/master/sound/soc/soc-ops.c#L427, but need the need the change in alsa-ucm-conf?

Thanks,

KaiChuan-Hsieh:
> @KaiChuan-Hsieh Looks like the hardware volume of cs42l43 behaves as expected, being set to 0dB (assuming the value of 128 does not change with system volume?). I have tried it on Dell Pro 16 Plus (pb16250) laptop and I stop hearing audio output only when it is below ~10% because it becomes too quiet. Can you tell me more about the system you are trying it on, and if the loss of audio is sudden, or does it gradually become too quiet to hear below 40%? Also can you send me output of "amixer contents" and "amixer -D hw contents" to check the software volumes levels?
> 
> The snd_soc_limit_volume approach is already done, and has been recently merged upstream: https://git.kernel.org/pub/scm/linux/kernel/git/broonie/sound.git/commit/?id=02b44a2b2bdcee03cbb92484d31e9ca1b91b2a38, as an alternative to fixing it in pipewire: https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/4622 Since it would be difficult to determine (from alsa-ucm-conf POV) if that kernel patch is applied, this alsa-ucm-conf approach might need to stay in place until wider adoption of the kernel patch.

@mstrozek thanks for looking in to it, the issue is reported by AnthonyAshok@dell.com. I help him to report the symptom here. I am asking him to add comment here.

KaiChuan-Hsieh:
@mstrozek I have chance to reproduce the problem by myself. Here is my verification result of the kernel change.

Before applying the kernel change, the volume is
**$ amixer -c 0 cget name='cs42l43 Speaker Digital Volume'
numid=40,iface=MIXER,name='cs42l43 Speaker Digital Volume'
  ; type=INTEGER,access=rw---R--,values=2,min=0,max=191,step=0
  : values=108,108
  | dBscale-min=-64.00dB,step=0.50dB,mute=0**

After applying the kernel change, the volume is
**$ $ amixer -c 0 cget name='cs42l43 Speaker Digital Volume'
numid=44,iface=MIXER,name='cs42l43 Speaker Digital Volume'
  ; type=INTEGER,access=rw---R--,values=2,min=0,max=128,step=0
  : values=45,45
  | dBscale-min=-64.00dB,step=0.50dB,mute=0**

The value has huge difference. Do you know if the kernel will remap the range after adjust the max output volume?

mstrozek:
@KaiChuan-Hsieh Thanks for your feedback. If I understand correctly, you stop hearing the output below 20% (which might still be acceptable, depending if you are in a noisy environment?), and can hear it at 40%? I thought that it was initially reported that the output is inaudible <= 40%?
I would like to request the "amixer contents" and "amixer -D hw contents", also from [AnthonyAshok@dell.com](mailto:AnthonyAshok@dell.com) if possible, there might be differences in other volume levels which might explain why you can still hear the audio at 40% system volume and he can't.

Regarding the kernel patch, when the hardware volume is set to higher than 128 (0dB) the audio was reported by Dell to become distorted, hence the need to limit it there. It is expected that the scaling would change with this as well (simply because 40% of 128 is less than 40% of 191, which was the previous maximum). The max safe value of 128 is still available to the user. The scaling might slightly change with the ucm patch applied (because hardware volume will be fixed at max 128), but it looks like you did not use it here, correct?

This discussion relates now to the kernel patch only, do you think it would be better to move it somewhere else? (direct emails, or maybe an issue on canonical JIRA?) and leave this comment thread for issues regarding the ucm patch.

PR #542, opened 2025-04-04
ucm2: sof-soundwire: Correct include file path for dsp.conf

ujfalusi:
The file path must be given as absolut and not relative, otherwise the dsp.conf file will not be found:
$ alsaucm dump text
ALSA lib utils.c:364:(uc_mgr_config_load_into) could not open configuration file /usr/share/alsa/ucm2/conf.d/sof-soundwire/../Intel/sof-hda-dsp/dsp.conf ALSA lib parser.c:78:(uc_mgr_config_load_file) error: failed to open file /usr/share/alsa/ucm2/conf.d/sof-soundwire/../Intel/sof-hda-dsp/dsp.conf: -2 ALSA lib main.c:1554:(snd_use_case_mgr_open) error: failed to import hw:0 use case configuration -2 alsaucm: error failed to open sound card hw:0: No such file or directory

Fixes: c9d1f7eafa6c ("UCM2: sof-soundwire: Enable DRC and equalizers for")

### DIFF
--- a/ucm2/sof-soundwire/sof-soundwire.conf
+++ b/ucm2/sof-soundwire/sof-soundwire.conf
@@ -185,7 +185,7 @@ If.HdmiIec61937 {
 }
 
 # Define macros to set up processing controls
-Include.dsp-variables.File "../Intel/sof-hda-dsp/dsp.conf"
+Include.dsp-variables.File "/Intel/sof-hda-dsp/dsp.conf"
 
 # Set BootSequence and FixedBootSequence for drc, eqiir, eqfir
 Macro.SpeakerDrc.SofDrcBootSetup "endpoint='PostMixerSpeaker' direction='Playback'"


ujfalusi:
@perexg, the error happened because someone installed the ucm files incorrectly on the machine:
```
ls -al /usr/share/alsa/ucm2/conf.d/sof-soundwire/sof-soundwire.conf  
-rw-r--r-- 1 root root 4702 Apr  4 13:26 /usr/share/alsa/ucm2/conf.d/sof-soundwire/sof-soundwire.conf
```
It should be a symlink:
```
ls -al /usr/share/alsa/ucm2/conf.d/sof-soundwire/sof-soundwire.conf
lrwxrwxrwx 1 sof sof 38 21. 2.  2024 /usr/share/alsa/ucm2/conf.d/sof-soundwire/sof-soundwire.conf -> ../../sof-soundwire/sof-soundwire.conf
```

I'm not sure if this fix still makes sense.

perexg:
It's shorter than the previous so I think that it's a good minor optimization. Also, I don't prefer to use relative paths among multiple driver configs.

PR #545, opened 2025-04-10
Add support of HyperX SoloCast

serhii-nakon:
Support of
- volume control
- un/mute control

PS: About un/mute control - only from PC to Mic, vise versa does not work - like when you un/mute Mic on PC it works as expected (even LED show actual status) but when you click button on Mic side it does not reflect on PC side.

### DIFF
--- a/ucm2/USB-Audio/HyperX/SoloCast-HiFi.conf
+++ b/ucm2/USB-Audio/HyperX/SoloCast-HiFi.conf
@@ -1,0 +1,18 @@
+SectionDevice."Mic" {
+	Comment "HyperX SoloCast Mic Input"
+
+	EnableSequence [
+		cset "name='Microphone Capture Switch' on"
+	]
+
+	DisableSequence [
+		cset "name='Microphone Capture Switch' off"
+	]
+
+	Value {
+		CapturePCM "hw:${CardId},0"
+		CaptureMixerElem "Microphone,0"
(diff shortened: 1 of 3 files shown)


serhii-nakon:
Without this it show two inputs like Analog Input and SPDIF Input and when I used it it does not control internal gains and does not muted anything (but it was possible to control from alsamixer).
With this one it works just from UI - like gnome and kde and etc.

serhii-nakon:
Also I noticed this error/warning in kernel logs (with and without this fixes) - looks like volume control does not work as expected, Looks like need add some quirk to snd-usb-audio driver for this Microphone?

```
[ 1162.300577] usb 5-1: USB disconnect, device number 2
[ 1164.975042] usb 5-1: new full-speed USB device number 4 using xhci_hcd
[ 1165.142588] usb 5-1: New USB device found, idVendor=03f0, idProduct=0b8b, bcdDevice=41.00
[ 1165.142594] usb 5-1: New USB device strings: Mfr=1, Product=2, SerialNumber=0
[ 1165.142597] usb 5-1: Product: HyperX SoloCast
[ 1165.142599] usb 5-1: Manufacturer: HP, Inc
[ 1165.297850] usb 5-1: Warning! Unlikely big volume range (=5632), cval->res is probably wrong.
[ 1165.297856] usb 5-1: [5] FU [Microphone Capture Volume] ch = 1, val = -5120/512/1
```

PR #546, opened 2025-04-14
Add Surface Laptop 7 configuration

quic-kdybcio:


ucm2/Qualcomm/x1e80100/x1e80100.conf
```diff
@@ -1,12 +1,12 @@
...
 
-Define.DMI_info "${sys:devices/virtual/dmi/id/board_vendor}-${sys:devices/virtual/dmi/id/product_family}"
+Define.DMI_info "${sys:devices/virtual/dmi/id/board_vendor}-${sys:devices/virtual/dmi/id/product_family}-${sys:devices/virtual/dmi/id/board_name}"
 
 If.LENOVOT14s {
 	Condition {
 		Type RegexMatch
 		String "${var:DMI_info}"
-		Regex "LENOVO.*Think((Pad T14s Gen 6.*)|(Book 16 G7 QOY))|(HP.*Omnibook X.*)|(ASUSTeK COMPUTER.*ASUS Zenbook A14)"
+		Regex "LENOVO.*Think((Pad T14s Gen 6.*)|(Book 16 G7 QOY)).*|(HP.*Omnibook X.*)|(ASUSTeK COMPUTER.*ASUS Zenbook A14.*)|(Microsoft Corporation.*Surface.*Microsoft Surface Laptop, 7th Edition)"
```
perexg, line 9: I think that `.*` is not required at the end of matching sequence. Could you test it? The alsa-lib's UCM regex code just check, if the match is found:

```
        err = regexec(&re, s, ARRAY_SIZE(match), match, 0);
        return err == 0;
```

Example: https://regex101.com/r/PHXerV/1

quic-kdybcio: I don't have the machines affected to test it, but I wrote a small C program using regex.h and it seems like what you're saying is right, amended that

tj90241:
If it's worth anything, I can confirm this fixes audio on my SL7 laptop. +1 from a testing standpoint.

PR #549, opened 2025-04-17
ucm2: MediaTek: mt8395-evk: Add support for SOF 

julien-massot:
Similar to #513, add support for MT8395 with SOF enabled.

We use different playback and capture PCM. Also, split the mixers that are not yet available upstream.


Kxuan:
Hi @julien-massot , the feature you added is really useful, and I was wondering if you're still planning to move it forward?

PR #571, opened 2025-06-04
Intel: avs: Add new UCMs after recent kernel work

amadeuszslawinski-intel:
After recent work to rename cards [1], userspace expects different card names and device numbers.

[1] https://lore.kernel.org/linux-sound/20250407124154.1713039-1-amadeuszx.slawinski@linux.intel.com/

There is one problem with HDMI not working. As far as I can tell, this is due to macro not handling pcm=0 correctly, but I have no idea how to fix that.
```
numid=1,iface=CARD,name='HDMI/DP Jack'
numid=7,iface=CARD,name='HDMI/DP,pcm=1 Jack'
numid=13,iface=CARD,name='HDMI/DP,pcm=2 Jack'
```

```
 DefineMacro.HDMI.If.hdmi {
        Condition {
                Type ControlExists
                Control "iface=CARD,name='HDMI/DP,pcm=${var:__Device} Jack'"
        }
```

I've opted for adding files instead of moving and editing them, because there may be users who currently use old card names.

Main differences are:
* HDAudio macros can now be used for HD-Audio card
* DMIC now automatically constraints number of channels, so no need for two separate files
* rewrote some files to make use of conditional syntax


crojewsk-intel:
Friendly ping. While not an expert for alsa-ucm-conf I did review all of this:

Reviewed-by: Cezary Rojewski \<cezary.rojewski@intel.com\>

perexg:
I've tried to solve the zero HDMI device issue in https://github.com/perexg/alsa-ucm-conf/commit/0b550766cacbd0924671ccd74396709d84953aae . Could you give a test, please?

It would be also nice to move the old configs to separate directory, so they can be removed later. It does also apply for old SST drivers which are pretty obsoleted now. I think that it's time to make this cleanup.

The last thing - for UCM, the digital microphone is usually not in a standalone UCM profile - we create merged UCM profile for SOF HDA DSP and ACP HDA (AMD) - see https://github.com/alsa-project/alsa-ucm-conf/blob/master/ucm2/HDA/HDA.conf#L6 and https://github.com/alsa-project/alsa-ucm-conf/blob/master/ucm2/Intel/sof-hda-dsp/sof-hda-dsp.conf .

amadeuszslawinski-intel:
Thanks for suggestion. I've tried but it doesn't work, before I had:
```
# spa-acp-tool -vvvvvvvvvv -c 2 list-verbose 2>&1| grep jack
UCM jack HDMI/DP,pcm=2 has_control=1
UCM jack HDMI/DP,pcm=1 has_control=1
Found 2 jacks.
```
With updated file I have:
```
# spa-acp-tool -vvvvvvvvvv -c 2 list-verbose 2>&1| grep jack
Probe of jack 'Line Out Jack' succeeded (not found)
Probe of jack 'Line Out Phantom Jack' succeeded (not found)
(...)
Probe of jack 'Front Headphone Jack',1 succeeded (not found)
Probe of jack 'Front Headphone Surround Jack' succeeded (not found)
Found 0 jacks.
```

Regarding old UCM files I can move them. For old SST driver if I recall correctly there were only like 3 or 4 files.

Regarding DMIC does it make sense to merge it with other card, when it is separate card and device?
We expose each codec as separate card in our driver, meaning usually there will be separate: speakers, headphones, dmic card.
Also in theory one can only enable DMIC support in kernel, without enabling all other devices.

amadeuszslawinski-intel:
Still no go, but I've looked closer this time, we can't override the Device itself, as it is also used to match card device in:
`PlaybackPCM "hw:${CardId},${var:__Device}"`

So introducing another variable makes it work, something like:
```
DefineMacro.HDMI {
        If.1 {
                Condition {
                        Type RegexMatch
                        Regex "(^$|0)"
                        String "${var:-__Device}"
                }
                True.Define.__PCMDevice ""
                False.Define.__PCMDevice ",pcm=${var:__Device}"
        }
        If.2 {
                Condition {
                        Type ControlExists
                        Control "iface=CARD,name='HDMI/DP${var:__PCMDevice} Jack'"
                }
                True.SectionDevice."HDMI${var:__Number}" {
                        Comment "HDMI / DisplayPort ${var:__Number} Output"

                        EnableSequence [
                                cset "name='IEC958 Playback Switch',index=${eval:($__Number-1)} on"
                        ]

                        DisableSequence [
                                cset "name='IEC958 Playback Switch',index=${eval:($__Number-1)} off"
                        ]

                        Value {
                                PlaybackPriority "${var:__Priority}"
                                PlaybackPCM "hw:${CardId},${var:__Device}"
                                JackControl "HDMI/DP${var:__PCMDevice} Jack"
                        }
                }
        }
}
```
I've also tried with previous version, but it doesn't work with it, so the above seems good to go for me.


Regarding DMIC, I don't mind either way, I guess as it is built into the devices same as speakers, I should group it with them?

perexg:
> So introducing another variable makes it work, something like:

Yes. Thanks. Fixed in https://github.com/alsa-project/alsa-ucm-conf/commit/1b69ade9b6d7ee37a87c08b12d7955d0b68fa69d (applied to master branch now)

bbzw6210:
# Intel AVS: PR #571 fixes HD-Audio/HDMI, but 4-channel DMIC fails without explicit `CaptureChannels`

I tested PR #571 on an HP laptop with Intel Sunrise Point-LP HD Audio and the new AVS card names.

The PR successfully fixes the missing UCM configuration for `AVS HD-Audio` and `AVS HDMI` on this system. However, I found a reproducible issue with the `AVS DMIC` profile: the internal digital microphone only supports 4-channel capture, while the PR's DMIC UCM does not specify `CaptureChannels`.

With PipeWire/ACP 1.0.5, this causes ACP to assume stereo, attempt to open the PCM with 2 channels, and fail with `EINVAL`.

Adding only:

```text
CaptureChannels 4
```

to the DMIC UCM makes the HiFi profile work and restores the internal digital microphone.

## System

```text
Kernel:
7.0.0-28-generic

PCI audio:
Intel Corporation Sunrise Point-LP HD Audio [8086:9d71] (rev 21)

Subsystem:
Hewlett-Packard Company [103c:853d]

Kernel driver in use:
snd_soc_avs

alsa-ucm-conf:
1.2.10-1ubuntu5.14

alsa-utils:
1.2.9-1ubuntu5

PipeWire:
1.0.5-1ubuntu3.3

WirePlumber:
0.4.17-1ubuntu4.1
```

ALSA exposes the following cards:

```text
0 [PROBE   ]: avs_probe_mb - AVS PROBE
1 [DMIC    ]: avs_dmic     - AVS DMIC
2 [HDAudio ]: avs_hdaudio  - AVS HD-Audio
3 [HDMI    ]: avs_hdaudio  - AVS HDMI
```

The platform drivers are:

```text
card1 -> avs_dmic
card2 -> avs_hdaudio
card3 -> avs_hdaudio
```

## 1. HD-Audio and HDMI with distribution UCM

With the installed UCM configuration, the new cards cannot be opened:

```bash
alsaucm -c hw:2 dump text
```

returns:

```text
ALSA lib main.c:1554:(snd_use_case_mgr_open)
error: failed to import hw:2 use case configuration -2

alsaucm: error failed to open sound card hw:2:
No such file or directory
```

The same happens for:

```bash
alsaucm -c hw:3 dump text
```

The ALSA PCMs themselves are present:

```text
card 2: HDAudio [AVS HD-Audio], device 0: HDAudio Analog

card 3: HDMI [AVS HDMI], device 0: HDMI1
card 3: HDMI [AVS HDMI], device 1: HDMI2
card 3: HDMI [AVS HDMI], device 2: HDMI3
```

## 2. Controlled test with PR #571

I tested PR #571 separately using `ALSA_CONFIG_UCM2`, so the test did not depend on previously modified files under `/usr/share/alsa`.

```bash
ALSA_CONFIG_UCM2=/tmp/alsa-ucm-pr571/ucm2 \
alsaucm -c hw:2 dump text
```

This immediately produced valid devices:

```text
Device.Headphones
Device.Mic1
Device.Speaker
```

Likewise:

```bash
ALSA_CONFIG_UCM2=/tmp/alsa-ucm-pr571/ucm2 \
alsaucm -c hw:3 dump text
```

produced:

```text
Device.HDMI0
Device.HDMI1
Device.HDMI2
```

After running PipeWire/WirePlumber with this UCM tree, the system exposed:

```text
Built-in Audio Speaker + Headphones
Built-in Audio HDMI / DisplayPort 0 Output
Built-in Audio HDMI / DisplayPort 1 Output
Built-in Audio HDMI / DisplayPort 2 Output
```

Internal speaker playback works.

So PR #571 correctly fixes the HD-Audio/HDMI part on this hardware.

## 3. DMIC regression with PR #571

The internal DMIC is still detected by ALSA:

```bash
arecord -l
```

```text
card 1: DMIC [AVS DMIC],
device 0: Digital Microphone
```

The PCM constraints are:

```bash
arecord -D hw:DMIC,0 \
  --dump-hw-params \
  -f S16_LE -r 48000 -d 1 /dev/null
```

Result:

```text
FORMAT: S16_LE
CHANNELS: 4
RATE: 48000
```

The important point is:

```text
CHANNELS: 4
```

This PCM does not accept 2-channel capture.

With the unmodified PR #571 UCM tree, ACP reports:

```text
UCM available for card hw:1
Set UCM verb to HiFi

Got CapturePCM for device Mic:
_ucm0001.hw:DMIC,0

UCM file does not specify 'CaptureChannels'
for device Mic, assuming stereo.
```

It then tries to configure the PCM with 2 channels:

```text
snd_pcm_hw_params_set_channels(2) failed:
Invalid argument
```

and repeats the same failure when retrying the device.

As a result, PipeWire does not expose the HiFi digital microphone source.

The generic/pro-audio probing later discovers the real hardware constraint:

```text
Device hw:1,0 doesn't support 64 channels, changed to 4.
```

and:

```text
rate:48000 channels:4
```

## 4. Minimal change that fixes the DMIC

I added only:

```text
CaptureChannels 4
```

to the microphone `Value` section of the PR #571 DMIC HiFi configuration:

```text
SectionDevice."Mic" {
    Comment "Digital Microphone"

    Value {
        CaptureChannels 4
        CapturePCM "hw:${CardId},0"
        CaptureMixerElem "DMIC"
    }
}
```

After restarting PipeWire and WirePlumber:

```bash
systemctl --user restart \
  pipewire.service \
  pipewire-pulse.service \
  wireplumber.service
```

the internal microphone appeared immediately:

```text
Sources:

* Built-in Audio Digital Microphone
  Built-in Audio Stereo Microphone
```

ACP now reports:

```text
Got CaptureChannels for device Mic: 4
Profile HiFi supported.
```

and the HiFi mapping uses:

```text
channels:4
```

The previous:

```text
assuming stereo
```

and:

```text
snd_pcm_hw_params_set_channels(2) failed
```

failure are gone from the UCM HiFi initialization path.

## Result

With PR #571 plus the explicit channel count, the system has working:

```text
AVS DMIC
└── Digital Microphone      working, 4 channels / 48 kHz

AVS HD-Audio
├── Speaker                 working
├── Headphones              available
└── Stereo Microphone       available

AVS HDMI
├── HDMI / DisplayPort 0    available
├── HDMI / DisplayPort 1    available
└── HDMI / DisplayPort 2    available
```

Speaker and Digital Microphone work simultaneously and the configuration also survives reboot.

## Question / upstream consideration

I am **not suggesting that `CaptureChannels 4` should necessarily be hard-coded for every AVS DMIC**.

However, on this hardware the current PR #571 profile does not work with PipeWire/ACP 1.0.5 because ACP assumes stereo when `CaptureChannels` is absent, while the hardware PCM accepts exactly four channels.

Would the intended fix be:

1. to specify the capture channel count in UCM for fixed-channel AVS DMIC devices, or
2. for ACP to derive the usable channel count from the PCM constraints instead of assuming stereo?

I can provide full `spa-acp-tool -vvvvvvvv` output, `alsa-info.sh`, `aplay -l`, `arecord -l`, `amixer`, or a diff of the working UCM configuration if useful.

Problem 1
Kernel 7.0
   │
   └── new AVS card names
              │
              └── old UCM has no matching profiles
                         │
                         └── PR #571 fixes Speaker + HDMI


Problem 2
PR #571 AVS DMIC
   │
   └── CaptureChannels missing
              │
              └── ACP assumes 2 channels
                         │
                         └── hardware requires 4
                                    │
                                    └── EINVAL / Mic missing
                                               │
                                               ▼
                                    CaptureChannels 4
                                               │
                                               ▼
                                         Mic works ✓





Disclosure: This debugging work was performed on the affected hardware with AI-assisted analysis. ChatGPT was used as a diagnostic and drafting aid to analyse command output, suggest controlled tests, and structure this report. The reported observations, command outputs, A/B tests, and final workaround were reproduced and verified on the physical machine by the reporter.

PR #579, opened 2025-06-12
UCM2: Intel: sof-hda-dsp: HiFi: Fix handling of mono DMICs

ujfalusi:
When a single DMIC is present in the system we need to set the CaptureChannels to 1 since the PCM device only supports mono, PA/PW will reject the profile since it cannot open the DMIC PCM device.


perexg:
@ujfalusi : This change is causing issues: https://bugzilla.redhat.com/show_bug.cgi?id=2393552

PR #628, opened 2025-10-23
ucm2: sof-soundwire: support rt713vb codec

shumingfan:
The rt713 is the multi-function codec, so check it by MultiCodec variable. 
This patch also supports the rt713VB, which integrates with DMIC function into one interface.

### DIFF
--- a/ucm2/codecs/rt712/init.conf
+++ b/ucm2/codecs/rt712/init.conf
@@ -1,30 +1,23 @@
 # RT712 specific volume control settings
 
-If.rt712_init {
+BootSequence [
+	cset "name='rt712 FU05 Playback Volume' 87"
+	cset "name='rt712 ADC 23 Mux' 'MIC2'"
+	cset "name='rt712 FU0F Capture Volume' 57"
+	cset "name='rt712 FU0F Capture Switch' 1"
+]
+
+If.rt712vb_init {
 	Condition {
 		Type RegexMatch
 		Regex "(rt712(-sdca)?)"
 		String "${var:MultiMicShadow}"
 	}
-	True {
+	True.BootSequence [
 		# RT712-VB integrated with DMIC
-		BootSequence [
-			cset "name='rt712 FU05 Playback Volume' 87"
-			cset "name='rt712 ADC 23 Mux' 'MIC2'"
-			cset "name='rt712 FU0F Capture Volume' 57"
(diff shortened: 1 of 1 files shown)


shumingfan:
> I am a bit confused, how do you distinguish rt713(va) and the new codec variant rt713vb, because the driver probably gives 'rt713' string in components string for both. It seems that rt712 has similar issue - [e479190](https://github.com/alsa-project/alsa-ucm-conf/commit/e4791900954c3951d15038822a55a3031aac49d2) .

@perexg Thanks for reviewing.
If the machine uses the DMIC of RT712/RT713 VA, the driver will give the 'rt712-dmic' and 'rt713-dmic' in the mic: string.
The MicCodec1 variable will keep the 'rt712-dmic' or 'rt713-dmic' string. Therefore, the DMIC of RT712/RT713VA will be configured.
However, the VB version will give the 'rt712' or 'rt713' string in the mic: string. We could check with MultiMicShadow variable.

ucm2/sof-soundwire/sof-soundwire.conf
```diff
@@ -93,9 +93,9 @@ If.multi_mic {
 
 If.multi_headset {
 	Condition {
-		Type RegexMatch
-		Regex "${var:HeadsetCodec1}"
-		String "${var:MultiCodec1}"
+		Type String
+		Haystack "${var:HeadsetCodec1}"
+		Needle "${var:MultiCodec1}"
```
perexg: This seems like an optimization, but `Haystack` and `Needle` should be swapped to follow the original regex match. The whole `HeadsetCodec1` string should be in `MultiCodec1'.

ucm2/codecs/rt712/init.conf
```diff
@@ -1,30 +1,22 @@
...
-	}
-	False {
-		BootSequence [
-			cset "name='rt712 FU05 Playback Volume' 87"
-			cset "name='rt712 ADC 23 Mux' 'MIC2'"
-			cset "name='rt712 FU0F Capture Volume' 57"
-			cset "name='rt712 FU0F Capture Switch' 1"
-		]
-	}
+		cset "name='rt712 FU1E Capture Volume' 47"
```
perexg: missing `cset "name='rt712 FU1E Capture Switch' 1"` here

PR #633, opened 2025-10-31
ucm2: sof: hdmi: Add 5.1 and 7.1 variants of HDMI devices when using IPC4

ujfalusi:
Systems using IPC4 can support up to 8 channels of audio (and passthrough) via HDMI.

In UCM the default PlaybackChannels is set to 2, which prevents users from selecting multichannel configurations.

When probing the card, Pipewire will drop configurations that are not supported either by the PCM device or based on ELD information.
This means that if the equipment supports only stereo then the 5.1 and 7.1 variants should not be visible, if the equipment is 5.1 capable, then only the 7.1 variant is removed.

The kernel will refine the PCM parameters based on the ELD information as wall when https://lore.kernel.org/linux-sound/20251029073600.13624-1-peter.ujfalusi@linux.intel.com/ is applied.

@perexg, @wtay, @ford-prefect, is this something which can help PW and user space to handle the HDMI a bit better with SOF?

### DIFF
--- a/ucm2/Intel/sof-hda-dsp/sof-hda-dsp.conf
+++ b/ucm2/Intel/sof-hda-dsp/sof-hda-dsp.conf
@@ -41,7 +41,24 @@ If.devdmic {
 
 SectionUseCase."HiFi" {
 	File "/Intel/sof-hda-dsp/HiFi.conf"
-	Comment "Play HiFi quality Music"
+	If.multichannel {
+		Condition {
+			Type String
+			Empty "${var:Iec61937Pcms1}"
+		}
+		True.Comment "Play HiFi quality Music"
+		False {
+			Variant."HiFi" {
+				Comment "Play HiFi quality Music"
+			}
+			Variant."HiFi 5+1" {
+				Comment "Surround 5.1"
+			}
+			Variant."HiFi 7+1" {
+				Comment "Surround 7.1"
+			}
+		}
+	}
 }
 
 Include.hda-init.File "/HDA/init.conf"
--- a/ucm2/codecs/hda/hdmi.conf
+++ b/ucm2/codecs/hda/hdmi.conf
@@ -39,6 +39,13 @@ DefineMacro.HDMI {
 				PlaybackPCM "hw:${CardId},${var:__Device}"
 				JackControl "HDMI/DP${var:__Suffix} Jack"
 			}
+
+			Variant."HiFi 5+1".Value {
+				PlaybackChannels 6
+			}
+			Variant."HiFi 7+1".Value {
+				PlaybackChannels 8
+			}
 		}
 	}
 }
--- a/ucm2/sof-soundwire/sof-soundwire.conf
+++ b/ucm2/sof-soundwire/sof-soundwire.conf
@@ -1,14 +1,5 @@
 Syntax 7
 
-SectionUseCase."HiFi" {
-	File "/sof-soundwire/HiFi.conf"
-	Comment "Play HiFi quality Music"
-}
-
-Include.led.File "/common/ctl/led.conf"
-Include.card-init.File "/lib/card-init.conf"
-Include.ctl-remap.File "/lib/ctl-remap.conf"
-
 Define {
 	SpeakerCodec1 ""
 	SpeakerChannels1 "2"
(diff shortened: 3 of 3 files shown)


ujfalusi:
> @perexg, another thing I have observed with this variant change is that the mics starts to hoard the ConflictingDevices on the 5+1 an 7+1 variants, `Device.Mic1`, `Device.Mic2` specifically: [hda-sof-variants-ucm.txt](https://github.com/user-attachments/files/23337183/hda-sof-variants-ucm.txt)
> 
> It does not seam to affect functionality and I'm yet to figure out how to prevent such buildup of conflicts..

If I put the ConflictingDevice settings in `ucm2/HDA/HiFi-mic.conf` to be wrapped under:
```
Variant."HiFi" {
}
```
Then only the devices under the "HiFi" verb have ConflictingDevices, "HiFi 5+1" and "HiFi 7+1" devices are all without this marker, I guess this is not good either, but again, does not seam to affect UI or `spa-acp-tool` presentation of profiles, devices.

perexg:
Yes, I already noted that the forced variants are not the best idea when we know the device parameters through ELD (handling also the device parameter changes through standard ALSA control API event mechanism).

Note that in `Syntax 8` I added possibility to create `Device Variants` - https://www.alsa-project.org/alsa-doc/alsa-lib/group__ucm__conf.html with other extensions to reduce the If complexity in future for the drivers like HDA supporting many hardware variants.

Erick555:
This has the benefit that it will make it work right away. Even distros released this spring could get the fix.  Pipewire work seems to be stuck at design decisions and it requires UI support from DE which will delay it even further. Meanwhile users of past several intel generations can't get surround sound output. I think it would be nice to get this done here, then calmly wait for ELD approach to finalize.

perexg:
We may follow the SplitPCM application variable (feature handshake) - https://github.com/alsa-project/alsa-ucm-conf/blob/4d9d7adbfc4ea538c6a44c70a03b0df502d5aa2d/ucm2/common/pcm/split.conf#L55

For example: If `"${var:@HDMIdirect}"` is not set, define variants in UCM for this moment. It took more than two years to implement channel split in pipewire based on UCM configuration. The variable name may be improved (suggestions?).

EDIT: Changed variable name from `HDMIvariants` to `HDMIdirect` - it should be opposite meaning than I though.

perexg:
> HDMI via HDA can support 2/4/6/8 channels for playback, S16_LE, S32_LE, rate 32000...192000, so if we want to convey a subset of this via UCM, we will have quite an explosion of SectionDevices, no? Like 44.1_2.0, 48_2.0, 96_2.0, 44.1_5.0, 48_5.0, etc.

Yes, it's problematic. Maybe we can also add ELD parsing support to UCM to filter out the "unsupported" UCM devices, but there are other issues that UCM does not handle like plugging monitor when UCM is probed... So the best way is to handle this on the server side at the moment.

> If we could have a flag for a SectionDevice to instruct the application (PW) to look at the PCM device/ELD/etc for information on what is supported?

To my knowledge, the current PW/PA automatically detects ELD elements through the control API associated to the given PCM device. I believe that presence of this element is enough to detect the HDMI/DP PCM device. So nothing else is required.

ClusterM:
Thanks a lot for this work — much appreciated!

I tested the same approach (HiFi 5+1 / 7+1 variants + `PlaybackChannels` 6/8 in `codecs/hda/hdmi.conf`) on:

- HP OMEN Transcend Laptop 14-fb0065TX (SKU 9T0M1PA#AB2)
- Meteor Lake-P HD Audio (`8086:7e28`) + sof-soundwire
- Intel Meteor Lake HDMI codec `0x8086281d`
- `CardComponents` includes `iec61937-pcm:…`, so the multichannel variants show up as intended
- Sink: USB-C DisplayPort → AVR (ELD reports LPCM up to 8ch)

With this UCM change, PipeWire/WirePlumber correctly exposes **Surround 5.1 (HiFi 5+1)** and opens HDMI as a 6ch sink. That part works well and matches what this PR aims to do.

Two related notes from the same setup (not blockers for this PR, just FYI):

1. **Kernel:** on MTL with KAE silent-stream (`SILENT_STREAM_KAE`), switching 2.0 → 5.1 could hang the display audio codec until reboot. Stereo was fine. I sent a fix to linux-sound: https://lore.kernel.org/linux-sound/20260722220309.24432-1-cluster@cluster.wtf/

2. **Channel positions:** even with `PlaybackChannels 6`, PipeWire sometimes labels channels in HDMI/CEA order (or as `aux*`) rather than the usual application order `FL,FR,RL,RR,FC,LFE`. I currently work around that with a small WirePlumber rule (`audio.position` + `api.alsa.use-chmap = false`). If there is a preferred UCM/ACP way to advertise the channel map, I would be happy to help test a follow-up.

Thank you again to everyone involved — this UCM change is exactly what was missing to make the 5.1 profile selectable on sof-soundwire HDMI.

ujfalusi:
Hi @ClusterM,
There should be already a fix applied for your point 1: https://lore.kernel.org/linux-sound/20260715180610.1371243-1-kai.vehmanen@linux.intel.com/T/#u

I have also seen the chmap shuffle and at the end the fix was done in PW side with https://gitlab.freedesktop.org/pipewire/pipewire/-/commit/e91a79dab712df11e0ab022fcb1349e8ecdc47ee
Before that I also needed to set `api.alsa.use-chmap = false` because UCM use forces the use-chmap to true, but the chmap was not actually set.

ClusterM:
Hi!

Yes, it seems like the same fix.

PR #637, opened 2025-11-06
ucm2: Add setting LED Mode in SetLED macro

mstrozek:
Add 'Mode' argument to SetLED macro in order to allow setting a different LED mode. This will allow mute switches with inverted values to control the LEDs with 'follow-route' option instead of default 'follow-mute'.

Update all calls to SetLED with appropriate Mode argument.

### DIFF
--- a/ucm2/codecs/cs42l45-dmic/init.conf
+++ b/ucm2/codecs/cs42l45-dmic/init.conf
@@ -5,5 +5,5 @@ BootSequence [
 ]
 
 Macro [
-	{ SetLED { LED="mic" Action="attach" CtlId="cs42l45 FU 113 Mute Switch" } }
+	{ SetLED { LED="mic" Action="attach" CtlId="cs42l45 FU 113 Mute Switch" Mode="follow-route"} }
 ]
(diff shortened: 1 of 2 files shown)


perexg:
Make this argument optional, like:

```
diff --git a/ucm2/common/ctl/led.conf b/ucm2/common/ctl/led.conf
index f10ad87..9bb7dc5 100644
--- a/ucm2/common/ctl/led.conf
+++ b/ucm2/common/ctl/led.conf
@@ -46,6 +46,13 @@ DefineMacro.SetLED {
                }
        }
        If.1 {
+               Condition {
+                       Type String
+                       Empty "${var:-__Mode}"
+               }
+               True.Define.__Mode "follow-mute"
+       }
+       If.2 {
                Condition { Type AlwaysTrue }
                True.FixedBootSequence [
                        sysw "-/class/sound/ctl-led/${var:__LED}/card${CardNumber}/${var:__Action}:${var:__CtlId}"
```

so it ends up as one file change. Thank you.

mstrozek:
@perexg Thank you for the suggestion! It's cleaner now when it's optional.

Unfortunately the switches have inverted behaviour because the SDCA driver must comply with MIPI SDCA specification, which explicitly says the Feature Units have to have Mute Controls where 1 = disabled (muted) signal :(

PR #644, opened 2025-11-19
ucm2: sof-soundwire: Update cs42l45 configurations

mstrozek:
Update JackControls to follow machine driver changes [1]

[1] https://github.com/thesofproject/linux/pull/5590

Keeping this PR as a draft until the SOF PR is finalized and merged

### DIFF
--- a/ucm2/codecs/cs42l45-dmic/init.conf
+++ b/ucm2/codecs/cs42l45-dmic/init.conf
@@ -1,9 +1,22 @@
 # cs42l45 specific control settings
 
+LibraryConfig.remap.Config {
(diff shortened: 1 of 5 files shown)


perexg:
> the jack controls are more user-friendly now,

Unfortunately adding just 'Channel' is not much useful, too. I've not studied SDCA documentation, but is not there a linking between endpoints (jack) and volume/switch registers? In this case, the naming should follow the endpoint like we do in the HDA driver.

EDIT: But the jack naming is corrected, so I will apply this change when the FU 41 volume control is corrected.

PR #659, opened 2025-12-10
ucm2: USB-Audio: Add Antelope Audio Zen Go support

ryodeushii:
Add support for Antelope Audio Zen Go SC

[alsa-info.txt](https://github.com/user-attachments/files/24074814/alsa-info.txt)


ryodeushii:
Hello, @perexg 
Just checking in if we can merge this PR :)

PR #665, opened 2025-12-19
ucm2: sof-soundwire: Add support for CS42L43B variant

mstrozek:
From UCM point of view CS42L43B codec adds 2 new decimators and removes a control to change Decimator 1 mode from base CS42L43. Thus reuse CS42L43 files, and guard the CS42L43B additions behind a condition that the new Decimator controls exist.

Link: https://lore.kernel.org/all/20260306152829.3130530-1-mstrozek@opensource.cirrus.com/


DeepHarsora-Dell:
Can we get this merged ??

PR #666, opened 2025-12-24
USB-Audio: Add Focusrite Vocaster interfaces

geoffreybennett:
Add split configs for the Focusrite Vocaster One and Two interfaces.


Marzal:
Would love to have Vocaster interfaces upstreamed

PR #669, opened 2025-12-29
Add support for the Behringer ULM200D wireless microphones.

kevincox:
This is a USB receiver for two wireless microphones. Each microphone appears on a different channel. By default these would be detected as left and right.

Just a disclaimer, I don't know what I am doing here. I used the UMC202HD as a reference and changed the values where it seemed reasonable. This device also has no output so I deleted everything that seemed output related. Seems to work.


kevincox (commented):
Thank you very much for the review and assistance.

PR #671, opened 2026-01-02
Remove outdated DisableSequence elements

mstrozek:
cs42l43's 'Headset Microphone' and 'Microphone' devices were both used
to control mic mute LED in the past, but this was changed [1] to only
allow the built-in microphones to control the LED. Thus clean up the
DisableSequences to remove the switching off that is no longer needed.
Also, cs42l45 had those Sequences copied over from cs42l43, so remove 
them here as well.

[1]: https://github.com/alsa-project/alsa-ucm-conf/blob/master/ucm2/common/ctl/led.conf

### DIFF
--- a/ucm2/sof-soundwire/cs42l45-dmic.conf
+++ b/ucm2/sof-soundwire/cs42l45-dmic.conf
@@ -7,10 +7,6 @@ SectionDevice."Mic" {
 		"Headset"
 	]
 
-	DisableSequence [
-		cset "name='cs42l45 FU 113 Channel Switch' 0"
-	]
-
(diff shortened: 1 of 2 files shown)


DeepHarsora-Dell:
The commit https://github.com/alsa-project/alsa-ucm-conf/pull/671/commits/9ddaab76d7ffcb0c786af67ebb0f68ab2c72e0ee is verified on a Dell Platform with CS42L43 codec with Intel PTL and it fixes, mic mute LED for headset is always ON issue. With this commit, it stays always OFF for headset MIC MUTE LED.

Issue #555, opened 2025-04-25
SSL2+ MK II "ucm is not supported for this device"


ATRAhassis:
"
ALSA lib parser.c:244:(error_node) UCM is not supported for this USB device (Solid State Logic SSL 2+ Mk II at usb-0000:0b:00.1-6, high speed @ USB31e9:0009)
ALSA lib main.c:1554:(snd_use_case_mgr_open) error: failed to import hw:0 use case configuration -6
alsaucm: error failed to open sound card hw:0: No such device or address

"

i believe i just need to make a file for it but i am too new to understand exactly how. willing to help if i can get any direction. was told if a UCM can be made for this device that its helpful for all distros so i figured id make this issue ticket

perexg:
Follow work like f10c04e8da06415c239fd62b9a41a89931994063 .

ATRAhassis:
> Follow work like f10c04e8da06415c239fd62b9a41a89931994063 .

So I just make a copy of the regular for the mk2 I'm assuming?

ATRAhassis:
> Nvm it turns out that all I had to do was to also restart pipewire. It seems to be working like that.
> 
> <img width="505" height="341" alt="Image" src="https://github.com/user-attachments/assets/b6c11a21-da68-4561-af18-d22a7fe1c231" />
> 
> Interesting enough my browser seems to also see a stereo input that has no sound. Not sure whats up with this.



Yeah I completely gave up on trying as I kept encountering so many issues with the + version I own.

PR #681, opened 2026-01-06
USB-Audio: Add support for the SSL 2 MK II and SSL2+ MK II

MTRNord:
This adds support for https://solidstatelogic.com/products/ssl-2-mkii and the https://solidstatelogic.com/products/ssl-2-plus-mkii and solves #555 

The MK II shows up the same way as the SSL2, but as a different USB Device.

I only own the SSL2 MK2, but not the SSL2+ MK2. So I can only verify it for the SSL2 MK2. However, I am fairly certain that the + version likely is also behaving identically. There are no visible spec changes between the v1 and MK2 of these devices, from all I was able to tell based on the internet. So, likely only the USB ID changed (at least for the things alsa-ucm is concerened about)

### DIFF
--- a/ucm2/USB-Audio/SolidStateLabs/SSL2PlusMK2.conf
+++ b/ucm2/USB-Audio/SolidStateLabs/SSL2PlusMK2.conf
@@ -1,0 +1,11 @@
+Comment "Solid State Labs SSL 2+"
+
+SectionUseCase."HiFi" {
+	Comment "HiFi"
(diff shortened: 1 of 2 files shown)


perexg:
I just used regex to match both USB IDs. Could you test https://github.com/alsa-project/alsa-ucm-conf/commit/0413a88d6b0455f9f4993dd0cada804c305c9bf0  ? Thank you.

Issue #547, opened 2025-04-15
Steinberg UR22C: Version 1.2.14 introduced "[ALSA UCM error]" on older firmware.


NoXPhasma:
Since version 1.2.14 the `node.description` of the Steinberg UR22C shows  `[ALSA UCM error]`:

```
node.description = "Steinberg UR22C [ALSA UCM error] Direct Steinberg UR22C"
```

```
> journalctl --user -u wireplumber --boot 0

systemd[1286]: Started Multimedia Service Session Manager.
wp-internal-comp-loader: Loading profile 'main'
spa.alsa: Error in ALSA UCM profile for _ucm0001.hw:UR22C,0 (HiFi: Line 3: source): CaptureChannels=4 > avail 2
spa.alsa: Errors in ALSA UCM profile for card Steinberg UR22C
wireplumber[1359]: wp-device: SPA handle 'api.libcamera.enum.manager' could not be loaded; is it installed?
wireplumber[1359]: s-monitors-libcamera: PipeWire's libcamera SPA plugin is missing or broken. Some camera types may not be supported.
wireplumber[1359]: default: Failed to get percentage from UPower: org.freedesktop.DBus.Error.NameHasNoOwner
wireplumber[1359]: spa.bluez5: BlueZ system service is not available

```

Reverting to version 1.2.13 fixes this issue.

NoXPhasma:
Changing `HWChannels` from `4` to `2`  in [UR22C-HiFi.conf](https://github.com/alsa-project/alsa-ucm-conf/blob/master/ucm2/USB-Audio/Steinberg/UR22C-HiFi.conf) fixes it.

pv:
There's maybe multiple variants of the hardware then. What number of channels the audio HW exposes on USB is separate question from how many physical ports the card has.

NoXPhasma:
I think there is only one variant of the UR22C, but there are other UR devices: https://www.steinberg.net/audio-interfaces/ur-c-series/

There are also only two input channels exposed for me. Those are combined as a stereo input device.

pv:
You should to attach more information about your hardware: In https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/4609 there's Steinberg UR22C with 4 capture channels

- `alsa-info.sh`
[...]

PR #684, opened 2026-01-11
USB-Audio: Add support for UR24C firmware version channel differences, fix UR22C

jwlademann:
Steinberg UR24C has a similar change to the number of capture channels for version 3 firmware that UR22C has. During implementation I found an issue with the condition used for UR22C which I fixed also.

See: https://github.com/alsa-project/alsa-ucm-conf/pull/554 and https://github.com/alsa-project/alsa-ucm-conf/issues/547


jwlademann:
[alsa-info.txt](https://github.com/user-attachments/files/24552963/alsa-info.txt)

PR #693, opened 2026-01-21
ucm2: Add config for Tascam US-2x2HR

jfrob:
USB ID 0644:8070 TEAC Corp. US-2x2HR

Line1/Mic1/Mic2 comments mostly match the device labels:
2 mono inputs, labeled IN1 and IN2
1 stereo output, shared with physical Line OUT L/R jacks and a Phone jack

[alsa-info.txt](https://github.com/user-attachments/files/24826345/alsa-info.txt)

### DIFF
--- a/ucm2/USB-Audio/TASCAM/US2x2HR-HiFi.conf
+++ b/ucm2/USB-Audio/TASCAM/US2x2HR-HiFi.conf
@@ -1,0 +1,54 @@
+Include.pcm_split.File "/common/pcm/split.conf"
+
+Macro [
+	{
+		SplitPCM {
+			Name "us2x2hr_mono_in"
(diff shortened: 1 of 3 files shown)


jfrob:
Does this still need something from me or is the check just broken?

perexg:
> Does this still need something from me or is the check just broken?

Fix commit author as noted in bot comment. Thank you.

jfrob:
> > Does this still need something from me or is the check just broken?
> 
> Fix commit author as noted in bot comment. Thank you.

~~I thought I already did it two weeks ago, exactly as the label says. If that is not enough, then I don't have a clue what is needed.~~
Never mind, saw the bot actually amended the label.

PR #712, opened 2026-03-02
ucm2: rt1320: Add Use case Configuration for rt1320 with DMIC

AndreGilerson:
My surface pro 11 for business uses the rt1320 with a microphone. I have patches for the kernel to enable the configuration in: https://github.com/AndreGilerson/linux-surface


perexg:
Applied #743. Closing. Create new PR if something is missing.

PR #719, opened 2026-03-10
Remove .gitignore from ucm2/conf.virt.d

nguyen140320:
Reason: 
This file is available when I fetching the source code with yocto project. This is empty now, so I think this is unnecessary. 
I belive we should remove this in upstream


perexg:
This path is enabled by default in https://github.com/alsa-project/alsa-ucm-conf/blob/master/ucm2/ucm.conf and users can store "virtual" configurations there. It's intentional directory.

PR #738, opened 2026-04-06
ucm2: sof-soundwire: cs42l43: support UAJ-less configuration

mstrozek:
In some cases cs42l43/cs42l43b devices can be set up without UAJ. In this case, guard the ConflictingDevices behind a check for a cs42l43's headset's presence so the UCM can still load even if the headset is missing.


perexg:
I put a bit simpler version to repo. It would be nice, if you can check it.

mstrozek:
> I put a bit simpler version to repo. It would be nice, if you can check it.

@perexg yeah, your simpler version works perfectly fine too, thanks!

PR #743, opened 2026-04-10
Support Lenovo laptops with rt722 and rt1320 codecs

oortcomet:
Add configs to support Lenovo laptops with rt722 and rt1320 codecs.


perexg:
Could you use `git rebase` and `git --force push` to the former branch next time? The former PR was #722.

PR #746, opened 2026-04-15
Update qcm6490-idp ucm2 conf

mohsRafi:
Fix internal mic mute handling by selecting Mic1 to mute the internal mic when the headphone microphone is enabled.

Rename the ucm2 conf for Qualcomm qcm6490-idp. Removed snd-card tags from conf, as name did not
correctly match the actual sound card name.


perexg:
2 is the default for channel fields, so it's not required to put this to configurations explicitly. Please, remove those lines. Thank you.

mohsRafi:
> 2 is the default for channel fields, so it's not required to put this to configurations explicitly. Please, remove those lines. Thank you.

Thanks for the review.
Updated commit changes as suggested.

Issue #567, opened 2025-05-27
Regex question


shumingfan:
We have an Intel platform machine that equips the rt722 multi-codecs and rt1320 amplifier.
This machine driver will use the amp of rt722 and rt1320 both.
Therefore, the component string is shown below.

`Components    : 'HDA:8086281d,80860101,00100000  cfg-amp:2 hs:rt722 spk:rt722 spk:rt1320'`

I tried to modify the regular expression to capture substrings to SpeakerCodec1 and SpeakerCodec2 variables.
However, I can't get SpeakerCodec2="rt1320", even though I changed the regular expression to 
```
SpeakerCodec {
	Regex " spk:(rt722)| spk:(rt1320)"
	String "${CardComponents}"
}
```
The variable SpeakerCodec2 is still an empty string.
What do I miss?

perexg:
It would be probably better to use only one `spk:` identifer in the driver like `spk:rt722+rt1320` in this case and use whole `rt722+rt1320` string as codec identifier (and use this later for the speaker configuration filename), because the rt722 and rt1320 config files must be merged for this combo, too.

The answer to regex question - you must define two matching groups like https://regex101.com/r/PVxPyF/1 (just remove `|`).

shumingfan:
Thanks for the suggestions. Will test on my side.

PR #747, opened 2026-04-16
ucm2: sof-soundwire: support dual-codec cs42l43-spk + cs35l56 speakers

pma1:
Some machines have both cs42l43 and cs35l56 speaker amps on SoundWire. Only the first spk: tag in CardComponents was handled, leaving the second codec without init, control remap, or mute LED.

Detect cs35l56 directly in CardComponents instead of SpeakerCodec1, and when cs42l43-spk is also present, initialise both codecs and extend the mute LED remap to cover all amp channels.

Toggle Speaker Switch in both device confs and set up cross-codec routing and mute tracking as needed.

### DIFF
--- a/ucm2/sof-soundwire/cs42l43-spk.conf
+++ b/ucm2/sof-soundwire/cs42l43-spk.conf
@@ -15,6 +15,7 @@ SectionDevice."Speaker" {
 	}
 
 	EnableSequence [
+		cset "name='Speaker Switch' on"
 		cset "name='cs42l43 Speaker L Input 1' 'DP6RX1'"
 		cset "name='cs42l43 Speaker R Input 1' 'DP6RX2'"
 	]
@@ -22,6 +23,7 @@ SectionDevice."Speaker" {
 	DisableSequence [
 		cset "name='cs42l43 Speaker L Input 1' 'None'"
 		cset "name='cs42l43 Speaker R Input 1' 'None'"
+		cset "name='Speaker Switch' off"
 	]
 
 
@@ -36,6 +38,18 @@ SectionDevice."Speaker" {
 		}
 		False.Value {
 			PlaybackMixerElem "cs42l43 Speaker Digital"
+		}
+	}
+
+	If.cs35l56_mute {
+		Condition {
+			Type RegexMatch
+			Regex "spk:cs35l56( |$)"
+			String "${CardComponents}"
+		}
+		True.Value {
+			PlaybackMixer "default:${CardId}"
+			PlaybackMixerElem "cs35l56 Speaker"
 		}
 	}
 
--- a/ucm2/sof-soundwire/sof-soundwire.conf
+++ b/ucm2/sof-soundwire/sof-soundwire.conf
@@ -108,10 +108,39 @@ If.multi_headset {
 If.spk_init {
 	Condition {
 		Type RegexMatch
-		Regex "(rt1318(-1)?|cs35l56(-bridge)?)"
-		String "${var:SpeakerCodec1}"
+		Regex "spk:cs35l56( |$)"
+		String "${CardComponents}"
 	}
-	True.Include.spk_init.File "/codecs/${var:SpeakerCodec1}/init.conf"
+	True {
+		If.dual_spk {
+			Condition {
+				Type RegexMatch
+				Regex "spk:cs42l43-spk"
+				String "${CardComponents}"
+			}
+			True {
+				Define.SpeakerAmps "${eval:($SpeakerAmps1-1)}"
+				Include.cs35l56_init.File "/codecs/cs35l56/init.conf"
+				LibraryConfig.remap.Config.ctl.default.map {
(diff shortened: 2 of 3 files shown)


perexg:
How does the components string look for this configuration ? 

EDIT: Also, I don't think that we should have multiple spk: labels in this string.  It should be something like `spk:cs42l43-spk+cs35l56` .

ucm2/sof-soundwire/cs42l43-spk.conf
```diff
@@ -39,6 +41,18 @@ SectionDevice."Speaker" {
...
 
+	If.cs35l56_mute {
+		Condition {
+			Type RegexMatch
+			Regex "spk:cs35l56( |$)"
+			String "${CardComponents}"
+		}
+		True.Value {
+			PlaybackMixer "default:${CardId}"
+			PlaybackMixerElem "cs35l56 Speaker"
```
mstrozek, line 52: Is that going to be combined with the cs42l43 MixerElem or replaced by it? Did you check if the switches are toggled by a mute button?

ucm2/sof-soundwire/sof-soundwire.conf
```diff
@@ -108,10 +108,39 @@ If.multi_headset {
 If.spk_init {
 	Condition {
 		Type RegexMatch
-		Regex "(rt1318(-1)?|cs35l56(-bridge)?)"
-		String "${var:SpeakerCodec1}"
+		Regex "spk:cs35l56( |$)"
+		String "${CardComponents}"
+	}
+	True {
+		If.dual_spk {
```
mstrozek, line 115: Probably the 'correct' approach is to change the sdw_utils and combine separate 'spk:' entries into one spk:cs42l43-spk+cs35l56 as outlined here: https://github.com/alsa-project/alsa-ucm-conf/issues/567#issuecomment-2911667717 and add a separate .conf for that configuration? I can send the patch for sdw_utils soon if that is the way to go

ucm2/sof-soundwire/sof-soundwire.conf
```diff
@@ -108,10 +108,39 @@ If.multi_headset {
...
+			Condition {
+				Type RegexMatch
+				Regex "spk:cs42l43-spk"
+				String "${CardComponents}"
+			}
+			True {
+				Define.SpeakerAmps "${eval:($SpeakerAmps1-1)}"
```
mstrozek, line 122: think it would be nicer to modify the macro within cs35l56 init.conf (or the cs42l43-spk+cs35l56.conf?) to check for 'AMP$Amp Speaker Switch' control presence rather than rely on modifying the number of amps.

ucm2/sof-soundwire/sof-soundwire.conf
mstrozek: based on https://github.com/alsa-project/alsa-ucm-conf/pull/747#issuecomment-4258678013, going to send an alternative for the combined 'spk:cs42l43-spk+cs35l56'

mstrozek: @perexg actually combining 'spk:cs42l43-spk+cs35l56' in the kernel can break older UCMs as the files will be missing, is this acceptable?

mstrozek (commented):
oops, forgot to submit review yesterday. The components are separately `spk:cs42l43-spk spk:cs35l56`. Would be nice to join them like `rt722+rt1320` but doing this in the kernel could break older UCMs as they would not have a cs42l43-spk+cs35l56.conf file present, thoughts @perexg ?

perexg:
> could break older UCMs as they would not have a cs42l43-spk+cs35l56.conf file present

@mstrozek We can eventually join SpeakerCodec1 and SpeakerCodec2 to keep the backwards compatibility:

```
If.multi_spk {
    Condition { Type String Empty "${var:-SpeakerCodec2}" }
    False.Define.SpeakerCodec1 "${var:SpeakerCodec1}+${var:SpeakerCodec2}"
}
```

It would be nice to test this solution.

EDIT: Yes, but the problem is when the driver is updated then the old UCM config won't work. I'm not too strict about this change - UCM configs can be upgraded in system and apparently this combo has never worked properly, or I miss something ?

mstrozek:
> > could break older UCMs as they would not have a cs42l43-spk+cs35l56.conf file present
> 
> EDIT: Yes, but the problem is when the driver is updated then the old UCM config won't work. I'm not too strict about this change - UCM configs can be upgraded in system and apparently this combo never worked properly, or I miss something ?

@perexg That's right, looks like it never worked fully properly. So if you are not strict with that, I will submit the alternative for joined spk tags, and will test your solution for the older kernel + newer UCM case

mstrozek:
@pma1 @perexg Alternative for cs42l43-spk+cs35l56 is here: https://github.com/alsa-project/alsa-ucm-conf/pull/748

perexg:
Applied #748 . Closing this one. Thanks

PR #748, opened 2026-04-17
ucm2: sof-soundwire: add separate init and combined cs42l43-spk+cs35l56 config 

mstrozek:
Following from a problem reported and discussed here: https://github.com/alsa-project/alsa-ucm-conf/pull/747

In some systems cs42l43-spk and cs35l56 can be used simultaneously to
drive seprate speakers (e.g. tweeters and woofers), add a configuration
to support that.
cs42l43-spk+cs35l56 case require its own init.conf file, but can reuse the cs42l43.conf file, so cs42l43-spk+cs35l56.conf is a symlink.

Also including an improvement for UAJ-less handling of separate cs42l43-spk since both these commits change spk_init file loading so they apply cleanly in one go.


pma1:
Tested with the kernel patch applied. The kernel produces:

    spk:cs35l56+cs42l43-spk

Two issues:

1. DefineRegex.SpeakerCodec: `\+[a-z0-9]+` doesn't capture `-spk` on
   the second component, so SpeakerCodec1 ends up as `cs35l56+cs42l43`.

2. Component order: the kernel puts cs35l56 first (lower SoundWire link),
   but the UCM expects `cs42l43-spk+cs35l56` in directory names and
   If.spk_init regex.

mstrozek:
Thanks @pma1, updated both the kernel patch and the ucm here, can you re-test them on your machine?

pma1:
With the V2 kernel patch, and applied these patches, Speaker playback, mute LED, mic LED, and remap all work correctly.

Reviewed-by: Aaron Ma <aaron.ma@canonical.com>
Tested-by: Aaron Ma <aaron.ma@canonical.com>

ucm2/codecs/cs35l56+cs42l43-spk
```diff
@@ -0,0 +1 @@
+cs42l43-spk+cs35l56
```
perexg, line 1: This file should be removed, doesn't ?

ucm2/sof-soundwire/sof-soundwire.conf
```diff
@@ -55,6 +55,23 @@ DefineRegex {
...
 
+# check to keep combatibility with older kernels that have two separate spk: tags
+If.multi_spk {
+	Condition {
+		Type RegexMatch
+		Regex "spk:cs42l43-spk( |$)"
+		String "${var:SpeakerCodec1}"
+	}
+	True {
+		Condition {
```
perexg: Missing If block and I would prefer to combine SpeakerCodec1 and SpeakerCodec2 regex output as I proposed in https://github.com/alsa-project/alsa-ucm-conf/pull/747#issuecomment-4266235215 . It should combine two 'spk:' labels to one.

ucm2/sof-soundwire/sof-soundwire.conf
```diff
@@ -108,7 +125,7 @@ If.multi_headset {
 If.spk_init {
 	Condition {
 		Type RegexMatch
-		Regex "(rt1318(-1)?|cs35l56(-bridge)?)"
+		Regex "(rt1318(-1)?|cs42l43-spk(\\+cs35l56)?|cs35l56(-bridge)?)"
```
perexg: You're creating `cs42l43-spk+cs35l56` and `cs35l56+cs42l43-spk' .conf links. If the older kernel honor this order, we should create only one link.

perexg (changes requested):

ucm2/codecs/cs35l56+cs42l43-spk
mstrozek: I think both need to be kept, there is no guarantee of the order the components will show up in. I'm getting spk:cs42l43-spk+cs35l56, but if different sdw links are used it can be cs35l56+cs42l43-spk like in https://github.com/alsa-project/alsa-ucm-conf/pull/748#issuecomment-4278786856

ucm2/sof-soundwire/sof-soundwire.conf
mstrozek: Added the If block, thanks. But looks like SpeakerCodec2 cannot be used, IIRC on my system when SpeakerCodec1 == cs42l43-spk, then SpeakerCodec2 == -spk

ucm2/sof-soundwire/sof-soundwire.conf
mstrozek: There is no guarantee of the order, both need to be there I'm afraid

mstrozek:
> With the V2 kernel patch, and applied these patches, Speaker playback, mute LED, mic LED, and remap all work correctly.
> 
> Reviewed-by: Aaron Ma [aaron.ma@canonical.com](mailto:aaron.ma@canonical.com) Tested-by: Aaron Ma [aaron.ma@canonical.com](mailto:aaron.ma@canonical.com)

Hi @pma1 , there were slight adjustments to the patch, do you mind reviewing/testing it again please?

ucm2/codecs/cs35l56+cs42l43-spk
perexg: Note that there is already similar .conf file. This one is without suffix.

ucm2/sof-soundwire/sof-soundwire.conf
perexg: Yes, I need update DefineRegex to use first group from all matches, it gives only groups for the first match right now:

```
Pattern:  spk:([a-z0-9]+((-sdca)|(-spk)|(-bridge))?(\+[a-z0-9]+((-spk)?))?)
Input:   a b c spk:cs42l43-spk spk:cs35l56

Matches:
  Match 1: " spk:cs42l43-spk"
    Group 1: "cs42l43-spk"
    Group 2: "-spk"

  Match 2: " spk:cs35l56"
    Group 1: "cs35l56"

Total matches: 2
```

ucm2/sof-soundwire/sof-soundwire.conf
perexg: Then this regex won't work correctly when cs35l56 is first.

EDIT: Forget this. The order is already corrected with previous Ifs.

ucm2/sof-soundwire/sof-soundwire.conf
perexg: Could you just combine two regex If block to one like 
`"(spk:cs42l43-spk .*spk:cs35l56( |$))|(spk:cs35l56 .*spk:cs42l43-spk( |$))"` ?

ucm2/codecs/cs35l56+cs42l43-spk
mstrozek: Without suffix because this is a link to a directory, do you prefer to link only the .conf files?

perexg: I see. But is this link required? The spk_init regex in sof-soundwire.conf does not match this combination.

mstrozek: yes, it was my mistake, it should be matching it now: https://github.com/alsa-project/alsa-ucm-conf/pull/748/changes#diff-19b4828c4ff1a7e90986d95fbc15fb685f07af5efd13fc0e67974fc8fe490c45R128

ucm2/sof-soundwire/sof-soundwire.conf
mstrozek: Yes, the combined regex works well, added now, thank you

PR #756, opened 2026-04-30
Sound over Display support for Microsoft Windows Dev Kit 2023 (Blackrock)

jglathe:
Introduce a device profile for the Microsoft Windows Dev Kit 2023. It has a miniDP out and 2 Type-C connectors capable of DP Altmode, and no other (easily) accessible othe audion input/outputs. To use it you also need the corresponding topology, will be a [PR in audioreach-topology](https://github.com/jglathe/audioreach-topology/tree/jg/blackrock).
I also tried to integrate it with the existing Lenovo Thinkpad X13s profile to get automatic detection (works here).

### DIFF
--- a/ucm2/Qualcomm/sc8280xp/sc8280xp.conf
+++ b/ucm2/Qualcomm/sc8280xp/sc8280xp.conf
@@ -7,5 +7,13 @@ If.LENOVOX13s {
 		Regex "LENOVO.*ThinkPad X13s.*"
 	}
 	True.Include.x13s.File "/Qualcomm/sc8280xp/LENOVO-X13s.conf"
-	False.Error "SC8280XP - ${sys:devices/virtual/dmi/id/board_vendor}-${sys:devices/virtual/dmi/id/product_family} model not supported"
+}
+
+If.microsoftblackrock {
+	Condition {
+		Type RegexMatch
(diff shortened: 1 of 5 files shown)


ucm2/Qualcomm/sc8280xp/sc8280xp.conf
```diff
@@ -7,5 +7,13 @@ If.LENOVOX13s {
 		Regex "LENOVO.*ThinkPad X13s.*"
 	}
 	True.Include.x13s.File "/Qualcomm/sc8280xp/LENOVO-X13s.conf"
-	False.Error "SC8280XP - ${sys:devices/virtual/dmi/id/board_vendor}-${sys:devices/virtual/dmi/id/product_family} model not supported"
```
perexg, line 10: You removed this error with this change. It would make sense to assign the included file to a variable and use this variable like USB-Audio.conf does.

perexg (changes requested):

ucm2/Qualcomm/sc8280xp/sc8280xp.conf
jglathe: Hmm X13s and Blackrock use different parts if the DMI for identification, therefore no single DMI variable yet. But it can be made that way, with three-part DMI like in x1e80100. Since sc8280xp now knows more than one device I oriented the change on x1e80100.conf, so no error when not known.

PR #764, opened 2026-05-24
HDA/HiFi-analog.conf: skip Line2 ConflictingDevice when loctl is empty

### LINKED ISSUE #763: HDA/HiFi-analog.conf: Line2 ConflictingDevice["Line1"] causes HiFi verb parse failure when loctl is empty
## Description

`HDA/HiFi-analog.conf` creates a `Line2` SectionDevice when `Front Playback Switch` exists. That device unconditionally declares `ConflictingDevice ["Line${var:LineDevice}"]`, but `Line1` (the device `LineDevice` expands to) is only created when `loctl != ""` — i.e. only when a `Line Playback Switch` or `Line Out Playback Switch` ALSA control is present.

On hardware that has `Front Playback Volume/Switch` but no `Line Playback Switch`, `loctl` stays empty, `Line1` is never created, but `Line2` still references it as a ConflictingDevice. The UCM HiFi verb then fails to parse entirely with:

```
ucm: unable to find device 'Line1'
ucm: device management error in verb 'HiFi'
```

WirePlumber receives the parse error and falls back to the `stereo-fallback` profile, which hides the DMIC (`hw:1,6`) from PipeWire entirely. The user sees only white noise from the floating headset jack.

## Affected hardware

- **Laptop**: Acer Predator PH16-73 (Arrow Lake-HX)
- **CPU**: Intel Core Ultra 9 275HX
- **Codec**: Realtek ALC245, PCI SSID 1025:1826
- **Driver**: `sof-audio-pci-intel-mtl` (SOF HDA DSP)
- **alsa-ucm version**: 1.2.15.3

ALSA controls present on this hardware:
- `Front Playback Volume` ✓
- `Front Playback Switch` ✓
- `Line Playback Switch` ✗ (absent — so `loctl = ""`)

## Root cause

In `HDA/HiFi-analog.conf`, the `If.linefront` block creates `Line2` when `Front Playback Switch` exists. The `Line2` device references `Line${var:LineDevice}` (i.e. `Line1`) as a ConflictingDevice unconditionally, but `Line1` only exists when the outer `If.linesw` block fires (requires `loctl != ""`).

## Impact

Any laptop with an HDA codec that exposes `Front Playback Switch` but no `Line Playback Switch` will hit this bug, causing DMIC to be silently hidden from PipeWire/WirePlumber. The stereo-fallback profile that WirePlumber falls back to does not include the DMIC PCM device.

## Fix

A patch is available in #764.

aalaei:
When `Front Playback Switch` exists but `Line Playback Switch` does not (`loctl=""`), the `Line2` SectionDevice is created but `Line1` is not. The unconditional `ConflictingDevice ["Line${var:LineDevice}"]` in `Line2` then causes the entire HiFi verb to fail:

```
ucm: unable to find device 'Line1'
ucm: device management error in verb 'HiFi'
```

WirePlumber falls back to `stereo-fallback` profile as a result, hiding the DMIC from PipeWire entirely.

Fix by making the `ConflictingDevice` conditional on `loctl` being non-empty — i.e. only declare the conflict when `Line1` was actually created.

Tested on: Acer Predator PH16-73 (Arrow Lake-HX, ALC245, PCI SSID 1025:1826)

Closes #763

### DIFF
--- a/ucm2/HDA/HiFi-analog.conf
+++ b/ucm2/HDA/HiFi-analog.conf
@@ -283,9 +283,15 @@ If.linefront {
 	True.SectionDevice."Line2" {
 		Comment "Second Line Output"
 
-		ConflictingDevice [
-			"Line${var:LineDevice}"
-		]
+		If.line1exists {
+			Condition {
+				Type String
+				Empty "${var:loctl}"
+			}
+			False.ConflictingDevice [
+				"Line${var:LineDevice}"
+			]
+		}
 
 		EnableSequence [
 			cset "name='Front Playback Switch' on"


alaeisw:
Hey @perexg ,
when is it expected to be released(i.e. a new tag).

PR #771, opened 2026-06-01
ucm2: USB-Audio: add Creative Sound Blaster G8 support

romain-pereira:
Add UCM2 configuration for the Creative Sound Blaster G8 (USB 041e:329b).

Devices (HiFi verb):
- Headphones: playback via Speaker DAC (hw:${CardId},0, -64 to 0 dB)

- External Mic, Line In, S/PDIF In, What U Hear: capture via hardware mux (PCM Capture Source), mutually exclusive via ConflictingDevice

BootSequence disables hardware monitoring paths (Line In/S/PDIF In Playback Switch are on by default in firmware) and sets External Mic as the default capture source.

Direct verb provided via common/directm.conf macro.

[alsa-info.txt](https://github.com/user-attachments/files/28466909/alsa-info.txt)


ucm2/USB-Audio/Creative/SoundBlaster-G8-HiFi.conf
```diff
@@ -0,0 +1,95 @@
...
+		CapturePriority 300
+		CapturePCM "hw:${CardId},0"
+		CaptureMixerElem "S/PDIF In"
+		CaptureVolume "S/PDIF In Capture Volume"
+		CaptureSwitch "S/PDIF In Capture Switch"
+	}
+}
+
+SectionDevice."Direct 1" {
+	Comment "Loopback — capture what is being played"
```
max-foss, line 77: Why don't we call it "Loopback", in line with the Evo4 among others.

ucm2/USB-Audio/Creative/SoundBlaster-G8-HiFi.conf
```diff
@@ -0,0 +1,95 @@
...
+		PlaybackPriority 200
+		PlaybackPCM "hw:${CardId},0"
+		PlaybackMixerElem "Speaker"
+		PlaybackVolume "Speaker Playback Volume"
+		PlaybackSwitch "Speaker Playback Switch"
+	}
+}
+
+SectionDevice."Mic" {
+	Comment "External Microphone"
```
max-foss, line 14: It does not have an internal mic, how about just "Microphone", it will be listed along with USB anyway.

ucm2/USB-Audio/Creative/SoundBlaster-G8-HiFi.conf
```diff
@@ -0,0 +1,95 @@
...
+	ConflictingDevice [ "Line In" "SPDIF In" "Direct 1" ]
```
max-foss, line 15: I would not go for "Direct" for the Loopback, unless its somehow has to do with the feature to skip the DSP.

ucm2/USB-Audio/Creative/SoundBlaster-G8-HiFi.conf
```diff
@@ -0,0 +1,95 @@
...
+
+	EnableSequence [
+		cset "name='PCM Capture Source' 'External Mic'"
+		cset "name='External Mic Capture Switch' on"
+	]
+	DisableSequence [
+		cset "name='External Mic Capture Switch' off"
```
max-foss, line 22: Do we really need this? It appears to work without the `DisableSequence ` and introduces additional risk.

max-foss:
Hi @romain-pereira,

I am not a reviewer here, just a user with the same device as you and wanted to share what I came up with.

By the way, what the Windows driver calls the **Direct** mode can be accessed the following way:
`sudo bash -c 'for h in /sys/class/hidraw/hidraw*; do grep -q "0000041E:0000329B" "$h/device/uevent" && [ "$(wc -c < "$h/device/report_descriptor")" = 29 ] && { { printf "\x5a\x39\x03\x00\x05\x00"; printf "\0%.0s" {1..58}; } > "/dev/${h##*/}"; exit; }; done; echo "Error: SoundBlaster G8 HID interface not found" >&2; exit 1'`

It can give slightly better sound quality and lower latency by bypassing the DSP, but I have not yet find the right spot to upstream this to.

Thanks

Max

ucm2/USB-Audio/Creative/SoundBlaster-G8-HiFi.conf
```diff
@@ -0,0 +1,95 @@
...
+		CapturePriority 400
+		CapturePCM "hw:${CardId},0"
+		CaptureMixerElem "Line In"
+		CaptureVolume "Line In Capture Volume"
+		CaptureSwitch "Line In Capture Switch"
+	}
+}
+
+SectionDevice."SPDIF In" {
+	Comment "S/PDIF / Optical Digital Input"
```
max-foss, line 56: Why not just "S/PDIF Input" or even "S/PDIF In". Either someone knows what this is or they don't and can look it up, unnecessary bloat not in line with other devices.

max-foss (commented):
Tested an equivalent profile on real hardware (G8 fw 10.23, Fedora 44, kernel 7.0.14, PipeWire 1.6.7, alsa-ucm 1.2.16.1): playback path and all four capture selector paths work, all referenced control names match the device. External Mic verified with non-zero duplex capture up to 96 kHz, so the control policy here is sound.

Two more things besides the inline nits:

- The PR description says the BootSequence disables the hardware monitoring paths ("Line In/S/PDIF In Playback Switch are on by default in firmware"), but the patch only sets 'Speaker Playback Switch' on and the default capture source. Either the description or the BootSequence should be updated to match.
- The "Direct 1"/Loopback Comment contains a non-ASCII dash, which makes GitHub flag the file for hidden characters. Plain ASCII would also be friendlier to grep.

ucm2/USB-Audio/Creative/SoundBlaster-G8-HiFi.conf
romain-pereira: Changed the user-facing comment to "Microphone" as suggested.

ucm2/USB-Audio/Creative/SoundBlaster-G8-HiFi.conf
romain-pereira: Agreed that this capture path should be presented as a loopback, not confused with the Creative DSP-bypass Direct mode. The visible comment is now "Loopback"; "Direct 1" remains only as the generic UCM device identifier.

ucm2/USB-Audio/Creative/SoundBlaster-G8-HiFi.conf
romain-pereira: Updated the user-facing comment to "Loopback". I kept "Direct 1" as the UCM device identifier, consistent with the generic UCM device naming used elsewhere in this repository.

ucm2/USB-Audio/Creative/SoundBlaster-G8-HiFi.conf
romain-pereira: I retested the updated profile on the G8 and all paths work correctly. I kept the DisableSequence entries so that an inactive source is explicitly muted when switching the hardware mux; no issue or state inconsistency was observed with them.

ucm2/USB-Audio/Creative/SoundBlaster-G8-HiFi.conf
romain-pereira: Changed the user-facing comment to "S/PDIF Input" as suggested.

romain-pereira:
@max-foss Thanks again for testing the profile and for the detailed review. I pushed commit d81a3e9 with the suggested user-facing naming cleanups and replaced the non-ASCII dash. I also fixed the BootSequence to explicitly disable the Line In and S/PDIF hardware monitoring paths, so the implementation now matches the PR description. I tested the updated profile on the G8 and playback plus all capture paths work correctly.\n\nThe HID command for the Creative DSP-bypass Direct mode is useful information, but I believe support for that mode should be handled separately from this declarative UCM profile, likely by a dedicated driver or control tool.

PR #772, opened 2026-06-03
Jg/revert t14s full mm setup

jglathe:
Unfortunately (or thank god, your choice) [the required topology change](https://github.com/linux-msm/audioreach-topology/pull/59) was declined in Audioreach, with very valid reasons. So, although this works really great on my boxes, it shouldn't be the default ucm config for T14s and same-config laptops. Therefore revert to the previous version that matches what you get from Audioreach.


SlyEcho:
I can confirm that this would restore the outputs on my Asus Zenbook A14.

PR #775, opened 2026-06-04
sof-soundwire: cleanups for codecs with multiple functions

perexg:
The load mechanism assumed only one multi-codec in the system, but  we saw in #774 multiple instances of those codecs (rt1320 and rt713).
    
Make the load mechanism straight:
    
- introduce file variables
- remove shadow variables
- probe multi-codecs separately and modify the file variables
- test for duplicate file inclusion (in both sof-soundwire / HiFi configs)
- merge dmic configs for rt712,rt713 to one file
    
This change also improves filters for speaker/headset/mic init includes.

Link: https://github.com/alsa-project/alsa-ucm-conf/issues/774

PR #778, opened 2026-06-05
Fixes for cs42l43/cs35l56 bridge speaker amp configs

charleskeepax:
Update a couple of issues with systems where the speaker amps are connected through the codec and the headphone output is not being used on the codec.


perexg:
Thank you @charleskeepax . I will apply this soon, but may I ask you for help with #775 ?

If you can give few tests with Cirrus hardware, it would be helpful. I would like to avoid major regressions. Thank you again.

charleskeepax:
Sure I can give that a test. Should be able to get round to that hopefully today but maybe tomorrow.

Issue #785, opened 2026-06-12
HDA/HiFi-mic.conf: Internal mic incorrectly gated by external Mic Jack on ALC287 without Input Source enum


fabianmartins:
### Machine
- **Model:** Lenovo Legion Pro 7 16IAX10H (DMI product name: 83F5)
- **Codec:** Realtek ALC287 (Vendor Id: 0x10ec0287, Subsystem Id: 0x17aa3906)
- **Driver:** sof-hda-dsp
- **BIOS:** Q7CN44WW
- **alsa-ucm-conf version:** 1.2.16

### Problem

On this machine, the internal microphone (Node 0x12, Pin Default 0x90a60140: `[Fixed] Mic at Int N/A`) is reported as unavailable by PipeWire/WirePlumber because the UCM configuration assigns the wrong `JackControl`.

The codec exposes:
- `Internal Mic Phantom Jack` = on (always-on, internal mic present)
- `Mic Jack` = off (external 3.5mm jack, nothing plugged in)
- **No** `Input Source` or `Capture Source` enum control

### Root Cause

In `ucm2/HDA/HiFi-mic.conf`, the `If.micsetup` condition checks for an `Input Source` / `Capture Source` enum. When neither exists, it enters the `True` branch which unconditionally sets:

```
DeviceMicJack "Mic Jack"
```

This associates the internal mic device with the external jack's availability. Since `Mic Jack` = off (no external mic plugged in), PipeWire's ACP marks the mic route as `available: no`, and GNOME reports no microphone available.

The internal mic hardware works perfectly — direct ALSA recording (`arecord -D hw:1,0`) succeeds.

### Expected Behavior

The internal mic should have no `JackControl` (or use `Internal Mic Phantom Jack`) so its availability is not tied to the external jack state.

### Proposed Fix

In the `True` branch of `If.micsetup` (the "no source selector enum" fallback), add a condition: if `Internal Mic Phantom Jack` exists, don't assign `DeviceMicJack "Mic Jack"`. Something like:

```
True {
    Define {
        DeviceMicName "Mic"
        DeviceMicComment "Stereo Microphone"
        DeviceMicJack "Mic Jack"
        DeviceMicPriority 200
    }

    If.internalmic_phantom {
        Condition {
            Type ControlExists
            Control "iface=CARD,name='Internal Mic Phantom Jack'"
        }
        True.Define.DeviceMicJack ""
    }
}
```

[...]

PR #792, opened 2026-06-16
HDA: Fix internal mic availability for HDA codecs with phantom jack

rwekyes:
## Problem

On HDA codecs that have both a `Mic Jack` (external, unplugged) and an 
`Internal Mic Phantom Jack` (fixed, always-on) control, the internal 
microphone gets assigned `Mic Jack` as its JackControl. PipeWire monitors 
this control for availability and marks the internal mic as "not available" 
since the external jack is unplugged, making it invisible to browsers and 
applications.

Confirmed on: Lenovo IdeaPad Slim 5 16AKP10 with Conexant SN6140 codec 
(PCI SSID 17aa:38b6).

## Fix

Add a block immediately before the `HDACaptureDevice` macro call that 
overrides `DeviceMicJack` to use `Internal Mic Phantom Jack` when that 
control exists. Since phantom jacks always report as connected, PipeWire 
correctly marks the internal mic as available.

The condition is safe for other hardware — if no `Internal Mic Phantom Jack` 
control exists, the block has no effect.

## Known limitation

This fix requires UCM to be loaded for the card. On non-ACP HDA cards, 
`HDA-Intel.conf` does not currently activate UCM, requiring a 
machine-specific workaround. A companion fix to `HDA-Intel.conf` is needed 
but has not been tested broadly enough to include here — input from 
maintainers on the right approach would be welcome.

## Related

A companion kernel patch adding a pin config quirk for the SN6140 on this 
hardware has been submitted to alsa-devel@alsa-project.org.

### DIFF
--- a/ucm2/HDA/HiFi-mic.conf
+++ b/ucm2/HDA/HiFi-mic.conf
@@ -388,6 +388,13 @@ DefineMacro.HDACaptureDevice.SectionDevice."${var:__DevName}" {
 	}
 }
 
+If.phantomjackoverride {
+	Condition {
+		Type ControlExists
+		Control "iface=CARD,name='Internal Mic Phantom Jack'"
+	}
+	True.Define.DeviceMicJack "Internal Mic Phantom Jack"
+}
 If.mic1 {
 	Condition {
 		Type String


perexg:
Seems similar to #785 . Bogus AI analysis. Could you show alsa-info output?

rwekyes:
I was kinda on the fence about making a PR at all, and I had Claude format it for me because I'm pretty new to all this(CIS student). Here's the alsa-info file:
[alsa-info.txt](https://github.com/user-attachments/files/29013376/alsa-info.txt)
To break it down a little more, I recently switched this laptop from Windows to Ubuntu and the mic stopped working. It showed as "unavailable" but with arecord I was able to verify it still worked physically. This laptop has 2 mic controls, 1 internal, and 1 external, but there is no enum control for it to select from. It looks like #785 had the same issue on a different model, but the PR was told that it was a kernel-side fix. I submitted a patch to the kernel as well, to add a quirk line for the pin config ([Here](https://lore.kernel.org/linux-sound/20260616173257.37373-1-rwekyes@gmail.com/)), but even that didn't work to get my mic on all the way. Pipewire still showed the mic as "unavailable". I had to do the fix in this PR, and I had to add a machine-specific config file (not included because it seemed not appropriate), but in the end I got it to work. At this point I'm definitely above my head, but since I got results (my laptop mic works now!) I figured I'd try and contribute. If this PR isn't the right approach or you've got some other advice, I'd love to get set on the right path, like I said I'm a student and I'd really like to help get this mic problem solved for the other folks out there that are having the same issue. Thanks!

perexg:
It seems that we have third case for the input source selection. The driver may do source switch (mic selection) automatically for some codecs without individual on/off switches.

I would prefer to detect those inputs using 'Boost' volume controls rather than bogus (phantom) jack introduced for the legacy pulseaudio/pipewire code stack. Also `DeviceMicName` in ucm config must be changed to `Internal Mic` to use the Boost volume control correctly (it means that the second mic device should be `Mic` for external input).

I will try prepare correct PR when my time permits.

perexg:
Please, check with latest configs. I hope that I fixed this situation in c17dfb63aeff72efb37996049c5dda525c33cd31 , but there are other cleanups preceding this one.

Also note that the AMD ACP driver should not be activated for your device:

```
2 [acppdmmach     ]: acp-pdm-mach - acp-pdm-mach
                      LENOVO-83HY-IdeaPadSlim516AKP10-LNVNB1612160
```

UCM uses this unused microphone input, too. The kernel driver should be modified and the device should be added to the filter table. EDIT: The ACPI table (BIOS) advertise this capability wrongly.

rwekyes:
After a big deep rabbit hole on this with Claude, I've got some info that might help you out here, but I am way over my head at this point. I had it write up the following report, and I attached my updated alsa-info. Note that I'm running a modified kernel just so my mic works with my hack that I've got, but the quirk table entry I submitted to the kernel did get accepted. When I used the latest config files after your update and disabled ACP, the mic worked physically, but it would not work with Firefox, Claude suggested it had something to do with UCM. You can close this PR if it's no longer relevant, or let me know if I should send this info somewhere else. Thanks so much for your time, and if you need any more info from me about this bug, please don't hesitate to get ahold of me.

Here's Claude's writeup:

Machine: Lenovo IdeaPad Slim 5 16AKP10
DMI: LENOVO / 83HY, board LNVNB161216
Codec (real capture path): Conexant SN6140 on HDA link, PCI SSID 17aa:38b6
alsa-ucm-conf: 1.2.15.3-1ubuntu1
Kernel: 7.1.0-rc7 (local build; carries the pending SN6140 pincfg quirk, otherwise stock)
The acp-pdm-mach card (card 2, label LENOVO-83HY-IdeaPadSlim516AKP10-LNVNB161216) enumerates but records no audio. The functional capture device is the analog Conexant SN6140 on the HDA link.
In check_acp_config() (sound/soc/amd/acp/acp-legacy-common.c), is_pdm_dev is set from the _WOV ACPI method:
cret = acpi_evaluate_integer(handle, "_WOV", NULL, &dmic_status);
if (!ACPI_FAILURE(ret))
    chip->is_pdm_dev = dmic_status;
On this machine _WOV returns a hardcoded constant. Decompiled from the SSDT under \_SB_.PCI0.GPPA.ACP_:
Name (WOVS, 0x01)
Method (_WOV, 0, NotSerialized)
{
    Return (WOVS)
}
WOVS is 0x01 with no other writes in the table, so _WOV unconditionally returns 1, forcing is_pdm_dev = true.
In acp_machine_select() (same file), is_pdm_dev == true combined with FLAG_AMD_LEGACY_ONLY_DMIC registers the acp-pdm-mach platform device:
cif (chip->flag == FLAG_AMD_LEGACY_ONLY_DMIC && chip->is_pdm_dev) {
    chip->mach_dev = platform_device_register_data(chip->dev, "acp-pdm-mach", ...);
}
The ACP device's _DSD also advertises "acp-audio-config-flag", 0x10.
Reproduction / workaround: unloading snd_acp_legacy_mach removes card 2. Direct capture from the Conexant device (arecord/pw-record) succeeds independently of the ACP card's presence.
alsa-info attached (captured with snd_acp_legacy_mach loaded, card 2 present).
[alsa-info-acp.txt](https://github.com/user-attachments/files/29865871/alsa-info-acp.txt)

perexg:
I'm closing this. If the internal microphone is not connected to AMD ACP bridge (seems like this case), the snd_acp_legacy_mach driver should not be activated in kernel. The correct fixes:

1) ideally BIOS should not advertise that ACP digital microphone is connected on this board (WOV)
2) you can override ACPI detection in driver - like https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=fc12cf16df9af867e1455b011cdd9310474a612e

PR #798, opened 2026-06-28
USB-Audio/Focusrite: fix 18i20 gen1/gen2 digital capture input mapping

amplogik:
The 1st/2nd-gen Scarlett 18i20 expose 18 USB capture channels, laid out
as Analog 1-8 (ch 0-7), S/PDIF 1-2 (ch 8-9), ADAT 1-8 (ch 10-17). The
3rd/4th gen insert two extra channels at 8-9, shifting S/PDIF to 10-11
and ADAT to 12-19. The HiFi.conf only ever encoded the 20-channel layout
for the digital inputs.

A recent fix worked around the resulting out-of-range error on gen1/gen2
by shrinking the ADAT input to 6 channels (12-17), but that is incorrect:
gen1/gen2 still have all 8 ADAT inputs (they sit at 10-17), and their
S/PDIF inputs are at 8-9, not 10-11. The previous version therefore
dropped ADAT 7-8 and mislabelled ADAT 1-2 as S/PDIF.

Make the digital-input channel indices generation-aware via new
SpdifInCh*/AdatInCh* variables (default = 20-ch layout, overridden in the
gen1/gen2 blocks). Both layouts have 8 ADAT channels, so the conditional
channel-count split collapses back into a single Direct 2 device.

Verified on 1st-gen hardware (USB 1235:800c): the capture routing
(Input Source NN Capture Route) reports Analog 1-8 / S/PDIF 1-2 / ADAT
1-8 on channels 0-17, and the profile now parses without UCM errors.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


Randesign:
I have  verified on 2nd Generation 18i20 hardware (8201) that this PR removes the UCM errors I was getting, and all channels are available.

@amplogik I don't know if you noticed the message above from the bot. It would be great to get this merged.  Thanks for your work on this.

amplogik:
@Randesign Thanks for testing on 2nd-gen hardware, and for the nudge — I'd missed the bot. Commit is now signed off (62f9847) and the DCO check passes. Should be good to merge.

PR #802, opened 2026-06-30
Add useful displaynames for the ThinkPad T14s gen2

WhyNotHugo:
The Lenovo ThinkPad T14s gen2 has three outputs, with generic names that come from the underlying audio device:

- HDMI / DisplayPort 1 Output
- HDMI / DisplayPort 2 Output
- HDMI / DisplayPort 3 Output

Add card-specific overrides for this particular Laptop model, renaming the first two to:

- USB-C DisplayPort Output
- HDMI Output

The last output retains its original name, since I haven't found any way to actually use it. Perhaps the dock (which I do not own) uses it in some way?


WhyNotHugo:
Signed. I don't see any references to the DCO in the repository itself.

perexg:
Unfortunately the ALSA HDMI devices are mapped dynamically (they are not persistent), so you cannot connect / describe hardware ports in this way. PA/PipeWire should change port name using the connected device identification obtained from EDID/ELD.

PR #812, opened 2026-07-17
ucm2: sof-soundwire: Add tas2783 support

jimxbj-dot:
The update was tested on B5406 (0x35E1) and CashmereXI.


xanthein:
@perexg Can you take a look at this PR when you're available?

perexg:
Applied with removal of tas2783 from the MultiCodecRegex variable (it defines only Speaker device).

PR #823, opened 2026-08-02
Hp 8ea1 ucm profile

m-rokai:


perexg:
Ideally the things should work with the standard auto-detection sof-soundwire profile. Could you show the components string (`amixer -c 0 info`) ?

PR #831, opened 2026-08-13
Add Blue Snowball UCM Mono Profile

XirXes:
This forces the microphone to be detected as a mono input device correctly. This fixes a hardware crash caused by OBS attempting to access two PCM channels in low latency mode at launch. Without this the microphone works in stereo, when OBS isn't running.


XirXes:
I have only validated USB ID 0d8c:0005 as that's the mic I have, but this did fix my longstanding issue with the mic under Linux. I previously had made a wireplumber conf file to force the audio to mono, but that solution felt like a bandaid.

XirXes:
Got it all squashed into one commit, hopefully this looks ready to go at this point.

PR #838, opened 2026-08-24
HDA: HiFi-spk: define 4-channel speaker layout when Bass Speaker is present

melcheikh:
### Summary

Systems with secondary smart amplifiers (such as Lenovo Legion Pro 7i Gen 10 `16IAX10H`, Legion 5i, and various Yoga models with AW88399 / CS35L41 amplifiers) expose a `Bass Speaker Playback Switch` control in ALSA.

This PR adds an `If.bass` condition to `ucm2/HDA/HiFi-spk.conf` so that when `Bass Speaker Playback Switch` is present, ALSA UCM defines `PlaybackChannels 4` and channel positions (`FL`, `FR`, `RL`, `RR`).

### Benefits
- Automatically routes audio across all 4 speakers (tweeters + woofers) in modern sound servers (PipeWire / PulseAudio).
- Eliminates the need for hardcoded machine-by-machine overrides for 4-speaker HDA setups.
- Tested on Lenovo Legion Pro 7i Gen 10 (DMI Product `83F5`).

Signed-off-by: Martin el Cheikh <martinelcheikh@gmail.com>


melcheikh:
Closing this myself — the premise no longer holds.

I proposed the 4-channel layout based on a Lenovo Legion Pro 7i Gen 10
(16IAX10H), where the rear AW88399 woofers were not being driven. That machine
is now handled in the driver: `alc287_fixup_legion_16iax10h_aw88399()` in
`sound/hda/codecs/realtek/alc269.c` (commit 494978ae8243) reroutes the bass
speaker pin to a DAC that has volume controls, and explicitly caps playback to
two channels:

```c
spec->gen.multiout.max_channels = 2;
spec->gen.pcm_rec[0]->stream[SNDRV_PCM_STREAM_PLAYBACK].channels_max = 2;
```

So for the hardware that motivated this PR, declaring `PlaybackChannels 4` in
UCM now works against the driver rather than with it.

Beyond that, the change was written against a single machine but lands in
`HiFi-spk.conf`, which every HDA system uses. I have not tested what it does on
other machines exposing a `Bass Speaker Playback Switch`, so it should not go in
on the evidence I have.

If the 4-channel layout does turn out to be correct for other machines exposing
that control, it should come with test reports from those machines rather than
from mine.

PR #841, opened 2026-08-26
ucm2: Qualcomm: sc8280xp: add DisplayPort devices on X13s

jglathe:
Model HDMI0/HDMI1 after the T14s HiFi layout: same MultiMedia1 frontend as Headphones, explicit RX_CODEC vs DISPLAY_PORT mixer sequences, and ConflictingDevice between Headphones and each DP sink.

Speakers stay on MultiMedia2 and are not part of that conflict, so the WSA path remains in the ACP profile across DP plug/unplug. JackControl is "DP0 Jack" / "DP1 Jack". No JackHWMute on playback; headset vs internal mic is priority-only (200 vs 100).

Compatible with the existing SC8280XP-LENOVO-X13S topology.

Assisted-by: Grok 4.5 (xAI)

### DIFF
--- a/ucm2/Qualcomm/sc8280xp/HiFi.conf
+++ b/ucm2/Qualcomm/sc8280xp/HiFi.conf
@@ -3,6 +3,8 @@
 
 SectionVerb {
 	EnableSequence [
+		cset "name='DISPLAY_PORT_RX_0 Audio Mixer MultiMedia1' 0"
+		cset "name='DISPLAY_PORT_RX_1 Audio Mixer MultiMedia1' 0"
 		cset "name='RX_CODEC_DMA_RX_0 Audio Mixer MultiMedia1' 1"
 		cset "name='WSA_CODEC_DMA_RX_0 Audio Mixer MultiMedia2' 1"
 		cset "name='MultiMedia4 Mixer VA_CODEC_DMA_TX_0' 1"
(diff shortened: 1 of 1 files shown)


jglathe:
Additional notes: Tested on my X13s with [slightly amended device tree](https://github.com/jglathe/linux_ms_dev_kit/commit/077f9f3ed4852e6d95090042c6bc4cbc1f20f991). The upstream audioreach topology is unchanged, so this shouldn't break anything.

Srinivas-Kandagatla:
Displayports can have dedicated FE (MultimediaX ..) , currently the patch is muxing Multimedia1 with other Back ends. This is not a good design. You should consider adding dedicated FE's which will allow you to run multiple playbacks on multiple Sinks. BTW, this will require audioreach-topology changes too.

jglathe:
Hi @Srinivas-Kandagatla , I am happy to oblige to do just that. Had my run-in with the [ucm vs. shared dp issue](https://github.com/alsa-project/alsa-ucm-conf/pull/843#issue-5293629888) multiple times, remember the independent variant of T14s? Should we reactivate it? Will update the audioreach topology, too. Converting to draft for the time being.

Srinivas-Kandagatla:
Yes, I think we have all the bits in kernel now. with recent codec side changes aswell.

jglathe:
Amended the config to independent multimedia streams for the DP outs. This one needs a new [audioreach topology](https://github.com/linux-msm/audioreach-topology/pull/77).

PR #846, opened 2026-08-31
HDA-Intel: Add ALC233 profile for Lenovo SSID 17aa:33c2

RenjithPananchikkal:
Realtek ALC233 with Lenovo subsystem ID 17aa:33c2 has no UCM2 profile. Without it, PipeWire/WirePlumber uses the generic ACP path which marks the Speaker as unavailable when a 3.5mm headset is plugged in, causing GNOME to hide the Speaker output entirely.

Add a CardComponents check for 17aa33c2 in HDA-Intel.conf and a dedicated HiFi.conf with Speaker always visible (no JackControl) and Headphones using jack detection via Front Headphone Jack.

Tested-by: Mark Pearson <mpearson-lenovo@squebb.ca>


RenjithPananchikkal:
[alsa-info-alc233-17aa33c2.txt](https://github.com/user-attachments/files/31662600/alsa-info-alc233-17aa33c2.txt)
alsa-info.sh output for ALC233 SSID 17aa:33c2

PR #848, opened 2026-09-01
ucm2: Qualcomm: Add support for HP OmniBook 5

jspingu:
Add DMI match for the Purwa-based HP OmniBook 5 Laptop 14-he0xxx. Setup is identical to the existing Omnibook X14 (WCD938x codec, 2x WSA speakers), which is matched in the Lenovo T14s block.


Srinivas-Kandagatla:
is this https://github.com/alsa-project/alsa-ucm-conf/pull/860/changes/a1b79e55f1cb9b6f09f05554149636d606d27918  covering Omnibook 5?

PR #852, opened 2026-09-06
ucm2: codecs: add rt721+rt1320 companion init configuration

### LINKED ISSUE #851: sof-soundwire: missing codecs/rt721+rt1320/init.conf makes UCM fail to load on cards reporting "spk:rt721+rt1320"
### Summary

On a machine whose card components report `spk:rt721+rt1320`, UCM fails to open the card entirely because `sof-soundwire.conf` includes `/codecs/${var:SpeakerCodecFile}/init.conf` with `SpeakerCodecFile` still set to the *combined* string `rt721+rt1320`, and no such directory exists.

`ucm2/sof-soundwire/rt721+rt1320.conf` was added by #744, but no matching `ucm2/codecs/rt721+rt1320/` was added alongside it. `codecs/rt721/` and `codecs/rt1320/` both exist; only the combined directory is missing. This is still the case on master.

The result is not a degraded profile but a hard failure: PipeWire/WirePlumber falls back to a generic ACP profile that opens the wrong PCM, so the machine has no working speakers and no microphones.

### Error

```
$ alsaucm -c hw:1 list _verbs
ALSA lib utils.c:373:(uc_mgr_config_load_into) [error.ucm] could not open configuration file /usr/share/alsa/ucm2/codecs/rt721+rt1320/init.conf
ALSA lib parser.c:80:(uc_mgr_config_load_file) [error.ucm] failed to open file /usr/share/alsa/ucm2/codecs/rt721+rt1320/init.conf: -2
ALSA lib main.c:1805:(snd_use_case_mgr_open) [error.ucm] failed to import hw:1 use case configuration -2
alsaucm: error failed to open sound card hw:1: No such file or directory
```

### Environment

- alsa-ucm-conf 1.2.16.1
- alsa-lib 1.2.16.1
- kernel 7.1.9 (Arch), Intel Panther Lake, `sof-audio-pci-intel-ptl`
- Card longname: `LENOVO-83SN-YogaPro715IPH11-LNVNB161216`
- Components: `HDA:80862822,80860101,00100000 cfg-amp:2 mic:dmic cfg-mics:4 iec61937-pcm:7,6,5 hs:rt721 spk:rt721+rt1320`

The kernel has no RT721 machine-driver entry for this platform, so `sof_sdw` uses the generic fallback ("Use SoundWire default machine driver with function topologies"). That path emits `spk:rt721+rt1320` as a single combined token.

### Root cause

In `ucm2/conf.d/sof-soundwire/sof-soundwire.conf`, `SpeakerCodecFile` is initialised to `SpeakerCodec1` (`rt721+rt1320`), then `MultiCodecInit` is supposed to narrow it to a single codec id. For this components string the macro is:

- `Haystack` = `=rt721+rt1320=rt721=dmic`
- For `CodecId=rt721`: the outer `Needle "=rt721"` matches, but the inner speaker test uses `Regex "=(rt721(-sdca)?)="` against `String "=rt721+rt1320="`, which does **not** match (the `+rt1320` sits between `rt721` and the closing `=`). So `SpeakerCodecFile` is not narrowed. It *does* match the headset branch, so `HeadsetCodecFile` correctly becomes `rt721`.
- For `CodecId=rt1320`: the outer `Needle "=rt1320"` does not match the haystack at all, because `rt1320` is preceded by `+`, not `=`.

`SpeakerCodecFile` therefore stays `rt721+rt1320`. The subsequent include filter regex contains the alternative `rt721`, which matches as a substring search against `rt721+rt1320`, so the guard passes and:

```
True.Include.a.File "/codecs/${var:SpeakerCodecFile}/init.conf"
```

resolves to the nonexistent `/codecs/rt721+rt1320/init.conf` and aborts the whole import.

For comparison, the Cirrus combination is handled by shipping the combined directory (`codecs/cs42l43-spk+cs35l56/`) rather than by narrowing, so the same code path succeeds there.

### Fix

Adding an (effectively empty) `ucm2/codecs/rt721+rt1320/init.conf` is sufficient. I verified this against an unmodified tree using `ALSA_CONFIG_UCM2`, and the verb then loads correctly:

```
$ alsaucm -c hw:1 set _verb HiFi list _devices
  0: HDMI3 ...
  3: Headphones
  4: Headset      Headset Microphone
  5: Mic          Digital Microphone
  6: Speaker      Speaker
$ alsaucm -c hw:1 set _verb HiFi get PlaybackPCM/Speaker
  PlaybackPCM/Speaker=_ucm0001.hw:sofsoundwire,2
```

Nothing from `codecs/rt1320/init.conf` is actually required for this combination:

- `codecs/rt721/init.conf` is already pulled in via the headset branch (`HeadsetCodecFile` = `rt721`), so the RT721 `BootSequence` still runs.
- `sof-soundwire/rt721+rt1320.conf` drives `rt1320-1 OT23 L/R Switch` directly in its Enable/DisableSequence rather than through `${var:SpeakerMixerElem}`, so the mixer-element definitions in `codecs/rt1320/init.conf` are unused here.

Including `codecs/rt1320/init.conf` from the combined file would in fact be actively wrong on this machine: the components report `cfg-amp:2`, which selects the `twoAmps` branch and remaps controls for a second amp (`rt1320-2 ...`) that does not exist — there is one RT1320 driving both channels, and `rt721+rt1320.conf` already handles that correctly via its `ControlExists` test on `rt1320-2 OT23 L Switch`.

If you would prefer the narrowing approach over shipping a combined directory, the alternative is to make `MultiCodecInit` split `A+B` speaker tokens (and set `SpeakerCodecFile` to the amp), but that would change behaviour for the existing Cirrus combination too, so the combined directory looks like the lower-risk fix.

### Possibly also affected

`ucm2/sof-soundwire/rt712+rt1320.conf` exists with no corresponding `ucm2/codecs/rt712+rt1320/`. If any machine reports `spk:rt712+rt1320`, it should fail in exactly the same way. I do not have that hardware to confirm.

I'm happy to open a PR with the added file if that's useful.

sri-sandeep108:
Add `ucm2/codecs/rt721+rt1320/init.conf` to support machines using the Realtek RT721 + RT1320 companion amplifier combination on `sof-soundwire`.

### Problem
When the kernel SoundWire driver detects RT721 and RT1320, it emits the combined component token `spk:rt721+rt1320`.
In `sof-soundwire.conf`, `SpeakerCodecFile` remains `rt721+rt1320`. When attempting to include `/codecs/${var:SpeakerCodecFile}/init.conf`, UCM fails because `/usr/share/alsa/ucm2/codecs/rt721+rt1320/init.conf` does not exist, causing the entire soundcard use case configuration import to fail (`-2`).

### Solution
Provide `ucm2/codecs/rt721+rt1320/init.conf`. RT721 initialization is handled through `codecs/rt721/init.conf` (via the headset branch), and RT1320 mixer controls are managed directly by `sof-soundwire/rt721+rt1320.conf`.

### Verification
Tested on Lenovo Yoga Pro 7 15IPH11 (`83SN`, Intel Core Ultra Panther Lake with RT721 + RT1320). UCM imports successfully, exposing HiFi playback on speakers, headphones, and multi-channel DMIC capture.

Closes: https://github.com/alsa-project/alsa-ucm-conf/issues/851
Fixes: 196b79b04126 ("sof-soundwire: add companion amp config with rt721 and rt1320")
Signed-off-by: Sri Sandeep Sakthivel <sandyorton7@gmail.com>


perexg:
A better fix is in the commit above. Could you retest, please ?

PR #855, opened 2026-09-14
sof-soundwire: fix speaker init include for combined codecs

pgavin:
Since commit cd342774cd02 ("sof-soundwire: cleanups for codecs with multiple functions") the speaker init filter regex is not anchored, so a combined speaker codec string like rt721+rt1320 or rt712+rt1320 also matches the single codec names from MultiCodecRegex. The config then tries to include codecs/rt721+rt1320/init.conf, which does not exist, and the whole UCM configuration fails to load:

  could not open configuration file .../ucm2/codecs/rt721+rt1320/init.conf
  failed to import hw:1 use case configuration -2

PipeWire falls back to a stereo profile on PCM 0 and the internal speakers (PCM 2) are never used.

Guard the regex with "=" like the other multi-codec checks, so only exact codec names with a codecs/<name>/init.conf are included. The combined codecs keep using sof-soundwire/<spk>.conf from HiFi.conf, and the rt721 codec init is still included via the headset codec, as before the cleanup.

Tested on Razer Blade 16 (RZ09-0581) with components:
  cfg-amp:2 mic:dmic cfg-mics:2 iec61937-pcm:7,6,5 hs:rt721 spk:rt721+rt1320


perexg:
>  , and the rt721 codec init is still included via the headset codec, as before the cleanup.

Please, retest with latest ucm tree. See above commit. This fix is not ideal. Thank you.

